PR #1, opened 2015-02-17
Restore default route

Sliim:
Apparently, the default route is not automatically restored on all linux
distribution.
Uncomment appropriate code to set default route when restoring routes.

I'm on debian wheezy OS.


adrienverge:
Nickel !

PR #6, opened 2015-09-14
Some improvements

lkundrak:
A couple of fixes and addition of --plugin option that potentially makes it possible to hook the tool to NetworkManager.


adrienverge:
@lkundrak, this is great. Just a suggestion in cf29d37 before merging it.

lkundrak:
Updated.

By the way, the work on the NetworkManager plugin has started here: https://github.com/lkundrak/NetworkManager-fortisslvpn

In order to do saner SSL verification we'll need to merge the "ssl" branch of my fork (not tested at all). I don't presently have an access to an actual SSLVPN gateway; testing with a mockup: http://paste.fedoraproject.org/267898/05412144

Will likely push more pull requests once the bits fit together and I have a chance to test this.

adrienverge:
Thanks, @lkundrak.

The bad news is that I don't have access to an SSLVPN anymore, neither.

PR #9, opened 2015-09-23
Some more fixes

lkundrak:
This time successfully tested with FortiOS 5 and some other unknown version which is considerably more buggy and picky.

Tested against a mockup too; here's the tool used:
https://git.gnome.org/browse/network-manager-fortisslvpn/tree/contrib

It would be cool if @jpoet and @k0k could verify if this works well for them and possibly use sslproxy.pl to capture the traffic that doesn't work (don't publish the dumps as they are; they could contain secrets!).

If it turns to work well it would be cool to have a tagged release the Linux distros could ship.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,6 +8,7 @@ AM_INIT_AUTOMAKE([foreign subdir-objects])
 
 # Checks for programs.
 AC_PROG_CC
+AC_GNU_SOURCE
 
 AM_SILENT_RULES([yes])
 
@@ -33,7 +34,7 @@ AC_TYPE_UINT8_T
 # Checks for library functions.
 AC_FUNC_MALLOC
 AC_FUNC_REALLOC
-AC_CHECK_FUNCS([bzero gethostbyname inet_ntoa memmove memset select socket strerror strstr strtol])
+AC_CHECK_FUNCS([bzero gethostbyname inet_ntoa memmove memset memmem select socket strerror strstr strtol])
 
 #AC_CONFIG_FILES([Makefile])
 AC_OUTPUT(Makefile)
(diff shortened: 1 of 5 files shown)


lkundrak:
> The "official" client I had did not use receive-all nor 1354... It probably depends on the version.

> The best would be to detect the version and dynamically adapt consequently, but it would require lots of work. Let's keep your options as they sound newer.

Yeah, they should be compatible; the peers seem to negotiate the common MRU, so if the old server uses 1024 then it wins. As for receive-all, I believe it only applies if we botch the HDLC encoding; which shouldn't happen. It's safe in any case; can't break things really.

lkundrak:
The push restored the permission denied check && extended it what I've seen with FortiOS 5.

kurzon:
Hey,

When I try to compile, I receive the following error.

$ make
  CC       src/openfortivpn-tunnel.o
src/tunnel.c: In function ‘pppd_run’:
src/tunnel.c:94:21: error: ‘struct vpn_config’ has no member named ‘plugin’
   if (tunnel->config->plugin) {
                     ^
src/tunnel.c:96:30: error: ‘struct vpn_config’ has no member named ‘plugin’
    args[i++] = tunnel->config->plugin;
                              ^
Makefile:552: recipe for target 'src/openfortivpn-tunnel.o' failed
make: *** [src/openfortivpn-tunnel.o] Error 1

adrienverge:
Hi @kurzon, you should try compiling latest master version.

kurzon:
Hi, yes ı'm using the version 1.1.0.

adrienverge:
Then why are you posting on this pull request page?

Have you tried `make clean` before compiling?

k0k:
Hi @lkundrak, I compile v1.1.3 and it's work for me. 
Thank you.

PR #11, opened 2015-09-24
Fix GPL + OpenSSL license clash

adrienverge:
In order to be allowed to package openfortivpn in a distribution, the
license must be slightly modified. This patch adds an exception in it.

See this explanation from Lubomir Rintel:

    There's a well known clash between the licenses (OpenSSL adds an
    "additional restriction" which is specifically not allowed by GPL).
    Therefore the license doesn't allow anyone to redistribute the
    resulting binaries (since a e.g. a Linux distro shipping both is
    technically a derived work of the package).


lkundrak:
Looks good to me.

PR #14, opened 2015-09-29
Understand split VPN routes

lkundrak:
Only supports FortiOS 5 with /remote/fortisslvpn_xml resource. For FortiOS we'd need to parse the /remote/fortisslvpn response it seems.

Adds some hooks for NetworkManager to manage the routes too.


adrienverge:
Looks good to me.

PR #16, opened 2015-09-30
ssl: add support for client keys and certificates

lkundrak:


lkundrak:
Note that this is not thoroughly tested as the original reporter didn't respond and I'm a lazy fuck. We may still want to merge it and eventually fix if we get negative feedback later.

adrienverge:
Agree.

PR #19, opened 2015-10-04
config: Rename --plugin to --pppd-plugin

adrienverge:
For consistency with other pppd-related options, and to preserve the
ability to add a plugin system to openfortivpn in the future, rename
the pppd plugin option to --pppd-plugin.


lkundrak:
Uh, oh. The 1.0.0 and 1.0.6 releases of NetworkManager-fortisslvpn already use this. They probably shouldn't have depended on a pre-release option, but damage done already. If you insist on renaming this, can we keep a compatibility option (perhaps undocumented) for some time?

adrienverge:
Sorry, I should have said that when you submitted this feature... But I really think `--plugin` isn't specific enough.

I think the compatibility trick is a good idea, I'll do this; in the end next releases of NetworkManager-fortisslvpn should use the new option.

PR #27, opened 2015-11-26
Add initial support for two-factor authentication.

dsgwork:
This commit adds initial support for two-factor authentication, where
the FortiGate sends a challenge through email/SMS after initial
user/password login and expects the user to provide the challenge
as a second authentication method.

This is tested working on FortiOS 5.2, with SMS tokens. It's very likely
that not all possible token authentication setups are supported (not sure
what the "code2=" parameter does, for example).

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -20,6 +20,7 @@
 #include "log.h"
 #include "ssl.h"
 #include "ipv4.h"
+#include "main.h"
 
 #define BUFSZ 0x8000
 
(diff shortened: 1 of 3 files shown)


dsgwork:
NOTE: This is probably not suitable for merging as is. Please review this
code carefully, I am not an expert C programmer. Make suggestions for
cleanups and I will amend the pull request. :)

Maybe we want to move read_password() out of main.c to something like io.c?

src/http.c
```diff
@@ -223,6 +224,63 @@ static int http_request(struct tunnel *tunnel, const char *method,
 }
 
 /*
+ * Read value for key from a string like 'key1=value1,key2=value2'.
+ *
+ * @return  1   in case of success
+ *          < 0 in case of error
+ */
+static int get_value_from_response(char *buf, char *keyname, char *retbuf, int retbuflen) {
```
adrienverge: This line is too long (80-char max).

src/main.h
```diff
@@ -0,0 +1,24 @@
+/*
+ *  Copyright (C) 2015 Adrien Vergé
```
adrienverge: `Adrien Vergé` -> `Davíð Steinn Geirsson`

src/main.h
```diff
@@ -0,0 +1,24 @@
...
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _MAIN_H
+#define _MAIN_H
+
+void read_password(const char *prompt, char *pass, size_t len);
+
+#endif
+
```
adrienverge: no extra newline

src/http.c
```diff
@@ -223,6 +224,63 @@ static int http_request(struct tunnel *tunnel, const char *method,
...
+	while (isalnum(*pos)) {
+		if (outpos >= retbuflen) {
+			return(-2);
+		}
+		retbuf[outpos] = *pos;
+		pos++;
+		outpos++;
+	}
+	retbuf[outpos] = '\0';
+	return(1);
```
adrienverge: `return 1;`

src/http.c
```diff
@@ -223,6 +224,63 @@ static int http_request(struct tunnel *tunnel, const char *method,
 }
 
 /*
+ * Read value for key from a string like 'key1=value1,key2=value2'.
+ *
+ * @return  1   in case of success
+ *          < 0 in case of error
```
adrienverge: Please describe the possible return values, for instance:

```
 * @return  1   when found
 *          -1  when key not found
 *          -2  when value is invalid
```

adrienverge:
Thanks @dsgwork, this looks good!

I made a few remarks, mostly cosmetics.

> Maybe we want to move read_password() out of main.c to something like io.c?

You are right, it would be better to move `read_password()` to a new file like `password.c` or `userinput.c` (`io.c` is actually for the I/O with the gateway).

adrienverge:
I just pushed some changes (automatic linting); so you'll need to rebase on `master` (no big deal -- just a conflict in `src/http.c`).

dsgwork:
Thanks for the code review. I'm closing this pull request, please see https://github.com/adrienverge/openfortivpn/pull/29 for the updated one with these changes.

adrienverge:
Thanks @dsgwork! I'll have a look as soon as I can.

Next time, please don't close the PR. Just `push --force` on the branch to update it ;)

PR #29, opened 2015-12-01
Implement two-factor authentication support

dsgwork:
This is a rework of the work from PR27, I've amended the issues raised during the code review. Also rebased on new master.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,8 @@ bin_PROGRAMS = openfortivpn
 openfortivpn_SOURCES = src/config.c src/config.h src/hdlc.c src/hdlc.h \
 		       src/http.c src/http.h src/io.c src/io.h src/ipv4.c \
 		       src/ipv4.h src/log.c src/log.h src/tunnel.c \
-		       src/tunnel.h src/main.c src/ssl.h src/xml.c src/xml.h
+		       src/tunnel.h src/main.c src/ssl.h src/xml.c \
+		       src/xml.h src/userinput.c src/userinput.h
 openfortivpn_CFLAGS = -Wall -Werror --pedantic -std=gnu99
 
 confdir=$(sysconfdir)/openfortivpn
--- a/src/http.c
+++ b/src/http.c
@@ -16,12 +16,14 @@
  */
 
 #include <string.h>
+#include <ctype.h>
 
 #include "http.h"
 #include "xml.h"
 #include "log.h"
 #include "ssl.h"
 #include "ipv4.h"
+#include "userinput.h"
 
 #define BUFSZ 0x8000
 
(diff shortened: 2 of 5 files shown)


Makefile.am
```diff
@@ -4,7 +4,8 @@ bin_PROGRAMS = openfortivpn
 openfortivpn_SOURCES = src/config.c src/config.h src/hdlc.c src/hdlc.h \
 		       src/http.c src/http.h src/io.c src/io.h src/ipv4.c \
 		       src/ipv4.h src/log.c src/log.h src/tunnel.c \
-		       src/tunnel.h src/main.c src/ssl.h src/xml.c src/xml.h
+		       src/tunnel.h src/main.c src/ssl.h src/xml.c \
+                       src/xml.h src/userinput.c src/userinput.h
```
adrienverge: Weird indentation here. Try to keep the same indent/spaces mix as lines above.

src/userinput.h
```diff
@@ -0,0 +1,6 @@
+#ifndef _USERINPUT_H
```
adrienverge: We need a copyright + license header in every file to comply with GPLv3 and beeing able to package openfortivpn for Fedora, Debian, etc.

src/http.c
```diff
@@ -252,28 +316,37 @@ int auth_log_in(struct tunnel *tunnel)
...
+
+		if (strncmp(res, "HTTP/1.1 200 OK\r\n", 17)) {
+			ret = ERR_HTTP_BAD_RES_CODE;
+			goto end;
+		}
+
+		ret = get_auth_cookie(tunnel, res);
 	}
 
+
```
adrienverge: please remove

dsgwork:
Thanks, updated the commit to fix these issues. Let me know if there's any more and I'll fix. :)

adrienverge:
Real-time updating :-)

Looks good to me now :+1:

adrienverge:
Unfortunately I am not able to test this, nor to check whether it breaks existing behavior for previous VPN versions (without two-factor auth). Any feedback from any user is welcome here!

If we don't get feedback in one or two weeks, I'll merge anyway since the code is rather clean.

dsgwork:
I've tested this on FortiOS 5.2.x (not sure of the specific version our network department is running now) both with and without two-factor authentication, and  haven't had any problems.

I don't have access to any FortiGate with older version to check though, so yeah... probably a good idea to give it a couple of weeks and see if anyone complains. :)

dsgwork:
Accidentally did "comment+close" instead of "comment", sorry about that.

dsgwork:
Hi,

I have another small patch in the waiting, this one is to allow authentication using different realms (the official client accepts a connection URI like "https://<vpn_gateway>:<port>/<realm>"). I add a --realm option that just sets the realm= parameter in auth_log_in.

I did this on master, but can't actually check it there as the only FortiGate I have access to that uses realms also uses two-factor authentication. Should I base that patch on the 2factor-rework branch and push an extra patch into this pull request, or is it better if I wait until this is merged before sending that as a seperate PR?

PS: I think it would make sense for us to accept a connection URI in the same format as the official client, I suspect most people just get that string from their IT departments and aren't sure how to map that to host, port and realm. But that might break the networkmanager stuff, haven't looked at that.

src/http.c
```diff
@@ -227,6 +229,67 @@ static int http_request(struct tunnel *tunnel, const char *method,
...
+		pos++;
+		outpos++;
+	}
+	retbuf[outpos] = '\0';
+	return 1;
+}
+
+static int get_auth_cookie(struct tunnel *tunnel, char *buf)
+{
+	int ret;
```
adrienverge: Set `ret` to `0`: stack variables are not initialized in C.

adrienverge:
Hi @dsgwork,

Thinking about this pull request, it would be better to have a command-line option to enable the two-factor auth. Being explicit will really help users identifying the problem in case auth failure. Something like `--two-factor` would do.

About realms: great! Maybe it's better to wait before opening the new PR.

About the URI: why not, but we would have to be extra careful to avoid breaking the current API. A possible solution could be to treat the first command-line argument differently whether it looks like `host:port` or `https://...`.

src/http.c
```diff
@@ -252,26 +316,34 @@ int auth_log_in(struct tunnel *tunnel)
...
+		get_value_from_response(res, "polid=", polid, 32);
+
+		read_password("2factor authentication token: ", tokenresponse, 255);
+
+		snprintf(data, 256, "username=%s&realm=&reqid=%s&polid=%s&grp=%s"
+		         "&code=%s&code2=&redir=%%2Fremote%%2Findex&just_logged_in=1",
+		         tunnel->config->username, reqid, polid, group, tokenresponse);
+		ret = http_request(tunnel, "POST", "/remote/logincheck", data, &res);
+		if (ret != 1)
+			return ret;
```
adrienverge: This should be `goto end;` to make sure `res` is freed.

adrienverge:
@dsgwork I wrote a commit to add a `--two-factor` option, it's on my `two-factor-auth` branch, along with your commit (which I modified to fix the two issues I just spotted).

Could you check if it works for you? If you like it, fell free to update this pull request:

```sh
# Get my branch
git fetch https://github.com/adrienverge/openfortivpn.git two-factor-auth
git checkout 2factor-rework
git reset --hard FETCH_HEAD

# Update the PR
git push -f
```

dsgwork:
@adrienverge Thanks, I merged your changes and changed the "return ret" to "goto end".

I'm not sure if I like the --two-factor option, as it feels unnecessary to make the user request it explicitly when the gateway indicates to us it's required by sending the code= parameter. But I defer to your judgement here :)

I'll wait on opening a pull request for the realm auth until this is merged to master, as the changes will conflict.

adrienverge:
> I'm not sure if I like the --two-factor option, as it feels unnecessary to make the user request it explicitly when the gateway indicates to us it's required by sending the code= parameter.

Does the gateway do this? I wasn't aware of that. In this is the case, you're right, we don't need a `--two-factor` option. But I don't see `code=` anywhere in your code? The client needs to be able to distinguish authentication failures from two-factor-auth requests.

PS: Please rebase on latest master (so the GitHub diff is more readable)

dsgwork:
Sorry, I meant to write tokeninfo= rather than code=. If we see tokeninfo= in the response, it means the gateway already sent a 2FA token and expects a second call to /remote/logincheck with the sent code. The original PR only changed the code path for the case where we got such a response.

I'll rebase on master. But did you want to keep the --two-factor option?

adrienverge:
> If we see tokeninfo= in the response, it means the gateway already sent a 2FA token and expects a second call to /remote/logincheck with the sent code. 

OK, it's clear for me now. Sorry I didn't get that from the beginning and asked for a `--two-factor` option. You're right: it's not needed, you can remove it. Also, could you add a small paragraph that describes how the gateway behaves in both normal and two-factor modes? In place of:

```
// Do we need to do two-factor authentication?
```

dsgwork:
I just updated the PR, removed the --two-factor option and added a better description of the two-factor-auth process.

adrienverge:
Thanks @dsgwork.

PR #23, opened 2015-11-20
Up COOKIE_SIZE to 240 as the SVPNCOOKIE is longer in newer FortiOS ve…

dsgwork:
…rsions

The previous value of COOKIE_SIZE resulted in the cookie being cut off at
least for FortiOS 5.2.x. Most likely they use a longer hash. This patch
allows connecting to these versions.

Issue #18, opened 2015-10-03
Received bad header from gateway


mrmodolo:
Hi!

I am trying to connect (mygateway is https://my.gateway.com.br:10443):

sudo /usr/bin/openfortivpn my.gateway.com.br:10443 -u marcelo --no-routes --no-dns --trusted-cert c755c435e2ec1221dea85847c190f9b9200013780bf82cefb25b6074562df2cd 
VPN account password: 
INFO:   Connected to gateway.
INFO:   Authenticated.
INFO:   Remote gateway has allocated a VPN.
ERROR:  Received bad header from gateway:
  (hex) 48 54 54 50 2f 31 2e 31 20 34 30 33 20 46 6f 72 62 69 64 64 65 6e 0d 0a 44 61 74 65 3a 20 53 61 74 2c 20 30 33 20 4f 63 74 20 32 30 31 35 20 31 33 3a 30 31 3a 32 35 20 47 4d 54 0d 0a 53 65 74 2d 43 6f 6f 6b 69 65 3a 20 20 53 56 50 4e 43 4f 4f 4b 49 45 3d 3b 20 70 61 74 68 3d 2f 3b 20 65 78 70 69 72 65 73 3d 53 61 74 2c 20 30 33 2d 4f 63 74 2d 32 30 31 35 20 31 33 3a 30 31 3a 32 35 20 47 4d 54 3b 20 73 65 63 75 72 65 3b 20 68 74 74 70 6f 6e 6c 79 3b 0d 0a 53 65 74 2d 43 6f 6f 6b 69 65 3a 20 53 56 50 4e 4e 45 54 57 4f 52 4b 43 4f 4f 4b 49 45 3d 3b 20 70 61 74 68 3d 2f 72 65 6d 6f 74 65 2f 6e 65 74 77 6f 72 6b 3b 20 65 78 70 69 72 65 73 3d 53 61 74 2c 20 30 33 2d 4f 63 74 2d 32 30 31 35 20 31 33 3a 30 31 3a 32 35 20 47 4d 54 3b 20 73 65 63 75 72 65 3b 20 68 74

  (raw) HTTP/1.1 403 Forbidden.
Date: Sat, 03 Oct 2015 13:01:25 GMT.
Set-Cookie:  SVPNCOOKIE=; path=/; expires=Sat, 03-Oct-2015 13:01:25 GMT; secure; httponly;.
[...]

PR #31, opened 2015-12-04
Increase COOKIE_SIZE (again)

adrienverge:
COOKIE_SIZE was recently changed from 213 to 240 to support newer VPN versions.

It seems that it was still not enough for some versions. This patchs sets it to 300 bytes.

References:
* https://github.com/adrienverge/openfortivpn/issues/18#issuecomment-158406434
* https://github.com/adrienverge/openfortivpn/pull/23#issuecomment-161621177

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -49,7 +49,7 @@ struct x509_digest {
 };
 
 #define FIELD_SIZE	64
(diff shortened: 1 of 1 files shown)


adrienverge:
The argtable changes sounds good. Patches welcome.

PR #33, opened 2015-12-10
Add support for realm authentication

dsgwork:
The official fortinet client accepts a connection string of the form:

    https://<vpnhost>:<port>/<realm>

Most VPN setups do not use the realm, so it's empty. It's only used
to provide multiple VPNs with different configuration on the same gateway.

This commit adds a --realm option to specify an authentication realm
to authenticate to. It defaults to empty.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -58,6 +58,7 @@ struct vpn_config {
 	char		username[FIELD_SIZE];
 	char		password[FIELD_SIZE];
 	char		cookie[COOKIE_SIZE + 1];
+	char            realm[FIELD_SIZE];
 
 	int	set_routes;
 	int	set_dns;
@@ -76,6 +77,7 @@ struct vpn_config {
 	do { \
 		(cfg)->gateway_host[0] = '\0'; \
 		(cfg)->gateway_port = 0; \
+		(cfg)->realm[0] = '\0'; \
 		(cfg)->username[0] = '\0'; \
 		(cfg)->password[0] = '\0'; \
 		(cfg)->pppd_log = NULL; \
--- a/src/http.c
+++ b/src/http.c
@@ -16,7 +16,6 @@
  */
 
 #include <string.h>
-#include <ctype.h>
 
 #include "http.h"
 #include "xml.h"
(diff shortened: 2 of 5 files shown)


dsgwork:
This patch is pretty straightforward. The change to get_value_from_response is because the tokeninfo= parameter can contain non-alphanumeric characters (in my case, it's the phone number the OTP was sent to, including a leading '+' for country code).

src/config.h
```diff
@@ -58,6 +58,7 @@ struct vpn_config {
 	char		username[FIELD_SIZE];
 	char		password[FIELD_SIZE];
 	char		cookie[COOKIE_SIZE + 1];
+	char            *realm;
```
adrienverge: It's better to copy the realm string into the config struct, rather than keeping a pointer to a memory area that can change. You should use:

```c
	char		realm[FIELD_SIZE];
```

src/main.c
```diff
@@ -175,6 +178,11 @@ int main(int argc, char **argv)
 				break;
 			}
 			if (strcmp(long_options[option_index].name,
+			           "realm") == 0) {
+				cfg.realm = optarg;
```
adrienverge: cf. previous comment in config.h:

```c
strncpy(cfg.realm, optarg, FIELD_SIZE - 1);
```

adrienverge:
Don't forget to update the man page ;-)

dsgwork:
Thanks @adrienverge, I've updated the PR with the suggested changes.

src/http.c
```diff
@@ -250,14 +250,16 @@ static int get_value_from_response(char *buf, char *keyname, char *retbuf,
 
 	pos = pos + keynamelen;
 	outpos = 0;
-	while (isalnum(*pos)) {
```
adrienverge: If you remove this, I think you can also remove the `#include <ctype.h>`.

src/http.c
```diff
@@ -250,14 +250,16 @@ static int get_value_from_response(char *buf, char *keyname, char *retbuf,
...
+	for (outpos = 0;
+	     (pos[0] != '=' && pos[0] != '&' && pos[0] != '\n');
```
adrienverge: Shouldn't you test the `!= '\0'` case also?

dsgwork:
Updated the PR to remove the ctype.h include from http.c, and check for the '\0' case in get_value_from_response.

That for loop is getting kind of ugly, but not sure how to best format it given the line length constraints.

adrienverge:
Thanks.

> That for loop is getting kind of ugly, but not sure how to best format it given the line length constraints.

Why not using `strtok` from the standard library?

adrienverge:
I'll do a patch later to use `strtok`. Anyway thanks for this improvement @dsgwork!

PR #34, opened 2015-12-11
http: Improve get_value_from_response using strtok

adrienverge:
Extract key/value pairs from "key1=value1&key2=value2" strings more cleanly, using strtok(3) from the standard library.

Also use `const char *` instead of `char *` for not modified arguments.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -228,39 +228,44 @@ static int http_request(struct tunnel *tunnel, const char *method,
 }
 
 /*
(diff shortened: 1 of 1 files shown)


adrienverge:
@dsgwork Can you confirm this works for you?

dsgwork:
@adrienverge Nice, I was working on something similar after your comment but yours is cleaner.

I can confirm it works if you add the comma as seperator to the strtok() calls :) :+1:

PR #39, opened 2015-12-23
add set-dns and set-routes flag to config file

kdmcclel:
Just wanted to add these to config so I didn't have to type in --no-dns every time I started the tunnel.

Regards,
Kevin

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -18,6 +18,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/stat.h>
+#include <string.h>
+#include <ctype.h>
 
 #include "config.h"
 #include "log.h"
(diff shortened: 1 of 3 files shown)


src/config.c
```diff
@@ -140,6 +140,22 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			strncpy(cfg->username, val, FIELD_SIZE - 1);
 		} else if (strcmp(key, "password") == 0) {
 			strncpy(cfg->password, val, FIELD_SIZE - 1);
+		} else if (strcmp(key, "set-dns") == 0) {
+			int set_dns = strtol(val, NULL, 0);
```
adrienverge: Please `long int` here (so it compiles correctly on other arch than x86_64).

src/config.c
```diff
@@ -140,6 +140,22 @@ int load_config(struct vpn_config *cfg, const char *filename)
...
+			if (set_dns < 0 || set_dns > 1) {
+				log_warn("Bad set-dns in config file: \"%d\".\n",
+				         set_dns);
+				continue;
+			}
+			cfg->set_dns = set_dns;
+		} else if (strcmp(key, "set-routes") == 0) {
+			int set_routes = strtol(val, NULL, 0);
```
adrienverge: same here

src/config.c
```diff
@@ -140,6 +140,22 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			strncpy(cfg->username, val, FIELD_SIZE - 1);
 		} else if (strcmp(key, "password") == 0) {
 			strncpy(cfg->password, val, FIELD_SIZE - 1);
+		} else if (strcmp(key, "set-dns") == 0) {
+			int set_dns = strtol(val, NULL, 0);
+			if (set_dns < 0 || set_dns > 1) {
+				log_warn("Bad set-dns in config file: \"%d\".\n",
+				         set_dns);
```
adrienverge: I'd rather not display the `"%d"` value here. For instance if an user sets: `set-dns = true`, he/she will be returned: `Bad set-dns in config file: "0".` (because that's what `strtol` returns).

adrienverge:
Thanks @kdmcclel, looks like a good improvement!

Could you squash the three commits into one to keep a clean git history?

adrienverge:
Thanks!

Issue #41, opened 2016-01-21
Can not connect using FortiToken


patrickhanft:
Hi! First of all thanks for this implementation of a fortisslvpn module, which just feels much better than using the proprietary client.

Using **version 1.1.3**, unfortunately, if my account within the fortigate is associated with a FortiToken which is our company standard, I just get this error:

    INFO:   Connected to gateway.
    ERROR:  Could not authenticate to gateway (No cookie given).
    INFO:   Closed connection to gateway.

If I release the association with a fortitoken, it works. I'm a bit irritated now, as some pull requests imply that there should be now some support for "two-factor authentication", but I don't know, if that should mean FortiTokens are supported yet.

Unfortunately I'm not enough developer to find the issue myself, but I'm happy to help with debugging, if someone tells me what I can do.

Thanks!

dsgwork:
Hi @patrickhanft 

I wrote the patch to add two-factor support to the client, so I think I can give you some more information. As far as I know, there is no documentation for the protocol used by the FortiVPN. The two-factor support I wrote was simply done by some simple reverse-engineering of the protocol. Unfortunately, I can only test things as my company has them set up, and we don't use the tokens (our setup has the OTP generated by the FortiGate, which then sends an email which gets converted to an SMS), so the tokens have never been tested (but based on what you say, they probably do not work).

Based on my experience, I expect adding support for the tokens would be quite an easy job, but still needs a VPN account associated with a token for testing. Most likely it's only needed to figure out how the initial login_check response differs from the normal one (to check if we need to do token auth) and in which parameter to the second login_check call we should stuff the OTP code.

If anyone wants to give this a go, the way I used to reverse the protocol was to just set up a stunnel that listens locally, with a self-created key+cert, and then makes an SSL connection to the actual VPN host/port. Then set up a tcpdump on the local port, make a successful connection to it with the official fortinet client and disconnect immediately. Then you can configure wireshark with the self-generated key to decrypt the communications, open the capture file and see what parameters are being sent over the wire.

patrickhanft:
Hi @dsgwork 

Thanks for your comment!

I tried to do a setup to intercept my communication with the fortigate. In the hope that you or anyone else might be able to draw the right conclusions from this, I have two logs created with stunnel and socat.

* [First try with openfortivpn](https://gist.github.com/patrickhanft/82b49ad7911298284d02) which failed
[...]

PR #42, opened 2016-01-22
Perform 2FA auth with zero-length tokeninfo

dsgwork:
Previously, we assumed that the tokeninfo parameter must be present and
contain data if we are to perform two-factor authentication. Based on
GitHub issue #41, in some cases, the tokeninfo parameter will be present
but contain an empty string.

This commit removes the check for string length, as it was not needed.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -374,7 +374,7 @@ int auth_log_in(struct tunnel *tunnel)
 		 */
 
 		ret = get_value_from_response(res, "tokeninfo=", token, 128);
-		if (ret != 1 || strlen(token) < 1) {
+		if (ret != 1) {
 			// No SVPNCOOKIE and no tokeninfo, return error.
 			ret = ERR_HTTP_NO_COOKIE;
 			goto end;


adrienverge:
Looks great!

Just a small request: could you add a debug line when `token` is empty (after line 381)? I'm sure it would ease future debugging.

dsgwork:
Sure, do you mean before line 381 if the token is not present at all, or to actually check specifically if it's present but an empty string?

If it's the latter, I'm not sure it's going to be too useful for debugging as we now just handle that case the same as the case where we have a token with a non-empty string. Do you have a suggestion for how the log message should look?

adrienverge:
> If it's the latter, I'm not sure it's going to be too useful

You're right. I just realized that in any case (before or after line 381), such a message won't bring any extra information.

I'll merge, unless you want to add something?

dsgwork:
Nothing to add on my end, I think this is good for merging.

adrienverge:
:+1:

PR #45, opened 2016-01-28
split routes: Use interface if assigned-addr is missing in XML

thp:
Some VPNs do not have "assigned-addr" set in the configuration XML
file, but still supply split route information. In this case, set
up routes using the interface instead of a gateway address.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -212,7 +212,7 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
 	setenv(env_var, dest, 0);
 	sprintf(env_var, "VPN_ROUTE_MASK_%d", tunnel->ipv4.split_routes);
 	setenv(env_var, mask, 0);
-	if (*gateway) {
+	if (gateway != NULL) {
 		sprintf(env_var, "VPN_ROUTE_GATEWAY_%d",
 		        tunnel->ipv4.split_routes);
 		setenv(env_var, gateway, 0);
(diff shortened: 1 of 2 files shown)


adrienverge:
Looks good!

I think with this change, you'll also need to change
```c
ipv4_add_split_vpn_route(tunnel, dest, mask, "");
```
to
```c
ipv4_add_split_vpn_route(tunnel, dest, mask, NULL);
```
in `parse_config()` (line 505).

src/ipv4.c
```diff
@@ -212,7 +212,7 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
 	setenv(env_var, dest, 0);
 	sprintf(env_var, "VPN_ROUTE_MASK_%d", tunnel->ipv4.split_routes);
 	setenv(env_var, mask, 0);
-	if (*gateway) {
+	if (gateway) {
```
adrienverge: Can you explicit to `if (gateway != NULL)`?

src/ipv4.c
```diff
@@ -223,8 +223,12 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
 	route_init(route);
 	route_dest(route).s_addr = inet_addr(dest);
 	route_mask(route).s_addr = inet_addr(mask);
-	route_gtw(route).s_addr = inet_addr(gateway);
-	route->rt_flags |= RTF_GATEWAY;
+	if (gateway) {
```
adrienverge: same here

thp:
Will look into your comments soon and update the PR then.

thp:
Updated PR.

adrienverge:
Thanks!

PR #53, opened 2016-03-21
http: auth_log_in: Initialize res to avoid crash on http_request() error

SimonKagstrom:
Misconfiguration from my part triggered a code path which causes a crash. This is a fix for that.

The crash is caused by freeing of an uninitialized pointer, the fix simply initializes it.

This also shows up when running with clang scan-build for those that are interested!


adrienverge:
Thanks!

PR #59, opened 2016-08-10
At least for openwrt this needs unistd.h

schlatterbeck:
I'm in the process of porting this to OpenWRT -- The ssize_t type seems to be defined only when including unistd.h.

### DIFF
--- a/src/hdlc.h
+++ b/src/hdlc.h
@@ -20,6 +20,7 @@
 
 #include <stdint.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #define ERR_HDLC_BUFFER_TOO_SMALL	-1
 #define ERR_HDLC_NO_FRAME_FOUND		-2


adrienverge:
Hey,

Thanks for the pull request.

It seems that only non-GNU systems need `unistd.h`:

> **Type size_t**
> This type has in some implementations a signed variant with name ssize_t, that is defined in the UNIX header file "unistd.h". For GNU C the type ssize_t is defined in "stddef.h" and thus resides in the same file as size_t.

https://en.wikibooks.org/wiki/C_Programming/C_Reference/stddef.h#Type_size_t

Could you add the following conditional for this new include?
```c
#ifndef __GNUC__
```

schlatterbeck:
On Fri, Aug 12, 2016 at 12:29:56AM -0700, Adrien Verg� wrote:
> Hey,
> 
> Thanks for the pull request.
> 
> It seems that only non-GNU systems need `unistd.h`:
> 
> > **Type size_t**
> > This type has in some implementations a signed variant with name
> > ssize_t, that is defined in the UNIX header file "unistd.h". For GNU
> > C the type ssize_t is defined in "stddef.h" and thus resides in the
> > same file as size_t.
> 
> https://en.wikibooks.org/wiki/C_Programming/C_Reference/stddef.h#Type_size_t
> 
> Could you add the following conditional for this new include?
> ```c
> #ifndef __GNUC__
> ```

This is on OpenWRT, clearly a GNU C-Compiler. Maybe this has changed in
later versions of the compiler or is library dependent? This is quite
bleeding-edge gcc 5.3 I think. And not built with GNU libc but with
uClibc. But __GNUC__ is probably defined.

I'd say if it doesn't hurt other systems include it unconditionally.
I've tried to compile on Debian oldstable X86 with this patch and it
works too.

Ralf
-- 

adrienverge:
As you say, it cannot hurt. Let's merge then!

PR #61, opened 2016-08-11
Option: --pppd-no-peerdns cfg: pppd-use-peerdns

schlatterbeck:
Sometimes we don't want pppd to change /etc/resolv.conf
so we have an option to turn this off and a config-file
variable which by default is on.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -24,7 +24,8 @@
 
 #define USAGE \
 "Usage: openfortivpn [<host>:<port>] [-u <user>] [-p <pass>]\n" \
-"                    [--realm=<realm>] [--no-routes] [--no-dns]\n" \
+"                    [--realm=<realm>] [--no-routes]\n" \
+"                    [--no-dns] [--pppd-no-peerdns]\n" \
 "                    [--pppd-log=<file>] [--pppd-plugin=<file>]\n" \
 "                    [--ca-file=<file>] [--user-cert=<file>]\n" \
 "                    [--user-key=<file>] [--trusted-cert=<digest>]\n" \
@@ -65,6 +66,8 @@ USAGE \
 "                                <digest> is the X509 certificate's sha256 sum.\n" \
 "                                This option can be used multiple times to trust\n" \
 "                                several certificates.\n" \
+"  --pppd-no-peerdns             Do not ask peer ppp server for DNS addresses\n" \
+"                                and do not make pppd rewrite /etc/resolv.conf\n" \
 "  --pppd-log=<file>             Set pppd in debug mode and save its logs into\n" \
 "                                <file>.\n" \
 "  --pppd-plugin=<file>          Use specified pppd plugin instead of configuring\n"\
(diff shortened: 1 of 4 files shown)


src/tunnel.c
```diff
@@ -95,7 +95,8 @@ static int pppd_run(struct tunnel *tunnel)
 			"/usr/sbin/pppd", "38400", "noipdefault", "noaccomp",
 			"noauth", "default-asyncmap", "nopcomp", "receive-all",
 			"nodefaultroute", ":1.1.1.1", "nodetach",
-			"lcp-max-configure", "40", "usepeerdns", "mru", "1354",
+			"lcp-max-configure", "40", "mru", "1354",
+			tunnel->config->pppd_use_peerdns ?  "usepeerdns" : NULL,
```
adrienverge: Can you use the syntax that is already used for `--pppd-log` and `--pppd-plugin`? (for readability)

See https://github.com/adrienverge/openfortivpn/blob/5c023b9/src/tunnel.c#L102-L110

src/main.c
```diff
@@ -113,6 +117,7 @@ int main(int argc, char **argv)
 		{"password",      required_argument, 0, 'p'},
 		{"no-routes",     no_argument, &cfg.set_routes, 0},
 		{"no-dns",        no_argument, &cfg.set_dns, 0},
+		{"pppd-no-peerdns", no_argument, &cfg.pppd_use_peerdns, 0},
```
adrienverge: Don't hesitate to re-indent other lines ;-)

adrienverge:
@schlatterbeck great job!

Just a few remarks before I can merge.

src/tunnel.c
adrienverge: Furthermore, the current implementation would break the `--pppd-log` and `--pppd-plugin` options (see the `i` variable).

schlatterbeck: On Fri, Aug 12, 2016 at 12:33:33AM -0700, Adrien Verg� wrote:
> > @@ -95,7 +95,8 @@ static int pppd_run(struct tunnel *tunnel)
> >  			"/usr/sbin/pppd", "38400", "noipdefault", "noaccomp",
> >  			"noauth", "default-asyncmap", "nopcomp", "receive-all",
> >  			"nodefaultroute", ":1.1.1.1", "nodetach",
> > -			"lcp-max-configure", "40", "usepeerdns", "mru", "1354",
> > +			"lcp-max-configure", "40", "mru", "1354",
> > +			tunnel->config->pppd_use_peerdns ?  "usepeerdns" : NULL,
> 
> Furthermore, the current implementation would break the `--pppd-log` and `--pppd-plugin` options (see the `i` variable).

OK, I'll look into this later today.

I also want to add syslog support. Do you want me to make this a
compile-time option (with runtime config option for turning it on) or
just runtime?

Ralf
-- 

adrienverge: If it doesn't bring any compilation dependency, runtime is enough.

PR #62, opened 2016-08-11
Fix off-by-one error in trusted-cert

schlatterbeck:
Now we compare only the contents of the certificate hash (without the
trailing 0-byte) and when copying it we add a 0-byte, not the other way
round.


adrienverge:
Thanks @schlatterbeck :+1:

PR #71, opened 2016-09-27
tunnel: disable insecure openssl default protocols/ciphers

### LINKED ISSUE #69: Restrict SSL options/ciphers to secure defaults
Hi,

ssl_connect in tunnel.c[1] probably should probably be using
`SSL_CTX_set_options(tunnel->ssl_context, SSL_OP_NO_SSLv3 | SSL_OP_NO_COMPRESSION)`
("my" server already refuses SSLv3, wasn't able to check if it'd accept compression, but it's likely safer to disable anyway)

We also want to limit the cipher lists, the openssl wiki[2] gives
`SSL_set_cipher_list(tunnel->ssl_handle, "HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4")`

This one is pretty important given "my" server supports very weak ciphers like RC4-MD5 and RC4-SHA, so a man-in-the-middle could downgrade our cipher unless we're strict about it.

For reference, here is a full list of ciphers my server seems to support (obtained looping through available ciphers as described here[3]):
```
    Cipher    : ECDHE-RSA-AES256-GCM-SHA384
    Cipher    : ECDHE-RSA-AES256-SHA384
    Cipher    : ECDHE-RSA-AES256-SHA
    Cipher    : DHE-RSA-AES256-GCM-SHA384
    Cipher    : DHE-RSA-AES256-SHA256
    Cipher    : DHE-RSA-AES256-SHA
    Cipher    : DHE-RSA-CAMELLIA256-SHA
    Cipher    : AES256-GCM-SHA384
    Cipher    : AES256-SHA256
    Cipher    : AES256-SHA
    Cipher    : CAMELLIA256-SHA
    Cipher    : ECDHE-RSA-AES128-GCM-SHA256
    Cipher    : ECDHE-RSA-AES128-SHA256
    Cipher    : ECDHE-RSA-AES128-SHA
    Cipher    : DHE-RSA-AES128-GCM-SHA256
    Cipher    : DHE-RSA-AES128-SHA256
    Cipher    : DHE-RSA-AES128-SHA
    Cipher    : DHE-RSA-SEED-SHA
    Cipher    : DHE-RSA-CAMELLIA128-SHA
    Cipher    : AES128-GCM-SHA256
    Cipher    : AES128-SHA256
    Cipher    : AES128-SHA
    Cipher    : SEED-SHA
    Cipher    : CAMELLIA128-SHA
    Cipher    : ECDHE-RSA-RC4-SHA
    Cipher    : RC4-SHA
    Cipher    : RC4-MD5
    Cipher    : ECDHE-RSA-DES-CBC3-SHA
    Cipher    : EDH-RSA-DES-CBC3-SHA
    Cipher    : DES-CBC3-SHA
```

I'm not sure all servers are as up to date as mine so don't know what defaults are really safe, I'd say the example here is good enough, but haven't looked at openfortivpn's configuration to see if it'd be easy to make configurable :)

[1] https://github.com/adrienverge/openfortivpn/blob/master/src/tunnel.c#L323
[2] https://wiki.openssl.org/index.php/SSL/TLS_Client
[3] http://superuser.com/questions/109213/how-do-i-list-the-ssl-tls-cipher-suites-a-particular-website-offers

martinetd:
It is still possible to re-enable SSLv3/compression through --insecure-ssl
or config file, or to specify a different cipher-list.

The insecure-ssl option covers both the option and cipher list as it is
more accessible, and if SSLv3 is allowed then weak ciphers probably aren't
too high on priority list...

If the server does not accept either TLS protocols or cipher the user will
get a message similar to this one:

ERROR:  SSL_connect: error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure
You might want to try --insecure-ssl or specify a different --cipher-list

Closes #69.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -211,6 +211,16 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			cfg->user_cert = strdup(val);
 		} else if (strcmp(key, "user-key") == 0) {
 			cfg->user_key = strdup(val);
+		} else if (strcmp(key, "insecure-ssl") == 0) {
+			int insecure_ssl = strtob(val);
+			if (insecure_ssl < 0) {
+				log_warn("Bad insecure-ssl in config file: \"%s\".\n",
+				         val);
+				continue;
+			}
+			cfg->insecure_ssl = insecure_ssl;
+		} else if (strcmp(key, "cipher-list") == 0) {
+			cfg->cipher_list = strdup(val);
 		} else {
 			log_warn("Bad key in config file: \"%s\".\n", key);
 			goto err_free;
--- a/src/main.c
+++ b/src/main.c
@@ -66,6 +66,13 @@ USAGE \
 "                                <digest> is the X509 certificate's sha256 sum.\n" \
 "                                This option can be used multiple times to trust\n" \
 "                                several certificates.\n" \
+"  --insecure-ssl                Do not disable insecure SSL protocols/ciphers.\n" \
+"                                If your server requires a specific cipher, consider\n" \
+"                                using --cipher-list instead.\n" \
+"  --cipher-list=<ciphers>       Openssl ciphers to use. If default does not work\n" \
+"                                you can try with the cipher suggested in the output\n" \
+"                                of 'openssl s_client -connect <host:port>'\n" \
+"                                (e.g. AES256-GCM-SHA384)\n" \
 "  --pppd-no-peerdns             Do not ask peer ppp server for DNS addresses\n" \
 "                                and do not make pppd rewrite /etc/resolv.conf\n" \
 "  --pppd-log=<file>             Set pppd in debug mode and save its logs into\n" \
@@ -106,6 +113,7 @@ int main(int argc, char **argv)
 	cfg.set_routes = 1;
 	cfg.set_dns = 1;
 	cfg.verify_cert = 1;
+	cfg.insecure_ssl = 0;
 	cfg.pppd_use_peerdns = 1;
 
 	struct option long_options[] = {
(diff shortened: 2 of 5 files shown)


src/config.c
```diff
@@ -211,6 +211,16 @@ int load_config(struct vpn_config *cfg, const char *filename)
...
 			cfg->user_key = strdup(val);
+		} else if (strcmp(key, "insecure-ssl") == 0) {
+			int insecure_ssl = strtob(val);
+			if (insecure_ssl < 0) {
+				log_warn("Bad insecure-ssl in config file: \"%s\".\n",
+				         val);
+				continue;
+			}
+			cfg->insecure_ssl = insecure_ssl;
+		} else if (strcmp(key, "cipher_list") == 0) {
```
adrienverge: Could you use `cipher-list` (no underscore) to be consistent with other config options?

src/main.c
```diff
@@ -66,6 +66,12 @@ USAGE \
 "                                <digest> is the X509 certificate's sha256 sum.\n" \
 "                                This option can be used multiple times to trust\n" \
 "                                several certificates.\n" \
+"  --insecure-ssl                Do not disable insecure SSL protocols/ciphers.\n" \
+"                                If your server requires a specific cipher, consider\n" \
+"                                using --cipher-list instead.\n" \
+"  --cipher-list=<list>          openssl ciphers to use. If default does not work\n" \
+"                                you can try openssl s_client to test alternative\n" \
+"                                ciphers.\n" \
```
adrienverge: This text is good and understandable! But could you show the expect list format for newbies? (With an example for instance `"HIGH:!MD5:!RC4"`).

Also could you precise that it works only when `--insecure-ssl` is not set?

src/config.c
martinetd: Actually that's what I wanted too, mixed it up with the variable and shamefully admitting I only tried if command-line option worked!

src/main.c
martinetd: Tried to reword a bit, tell me if it sounds better (we're getting close to the 4k limit for a single string btw, whoever adds the next option will get extra fun splitting help into two strings!)

Also made --cipher-list work even if --insecure-ssl is used e.g. insecure-ssl only disables default cipher list, not explicit ones

adrienverge:
@martinetd, this is great.

Thanks!

PR #72, opened 2016-10-17
update man page, example config and help message

mrbaseman:
The example config file, the help message and the man page were missing some options which are actually supported. This pull request suggests a few changes to complete them.


mrbaseman:
some checks have failed, I'll redo

mrbaseman:
Finally managed to pass the tests. Sorry for the noise, but the results at travis give different results than when I run them locally. Indentation in enums seems to be different depending on the environment in which the tests are run.

doc/openfortivpn.1
```diff
@@ -115,14 +126,24 @@ username = foo
 .br
 password = bar
 .br
+user-cert = /etc/openfortivpn/user-cert.pem
+.br
+user-key =/etc/openfortivpn/user-key.pem
+.br
```
adrienverge: This section is just an example, I prefer to keep it clear by just having a few options (not all).

etc/openfortivpn/config
```diff
@@ -3,3 +3,12 @@ host =
...
 password =
+# set-dns =
+# set-routes =
+# pppd-use-peerdns =
+# trusted-cert =
+# ca-file =
+# user-cert =
+# user-key =
+# insecure-ssl =
+# cipher-list =
```
adrienverge: Same here, this is unneeded and I prefer keep conf files simple...

src/main.c
```diff
@@ -95,7 +93,15 @@ USAGE \
...
-"      trusted-cert = othercertificatedigest6631bf...\n"
+"      trusted-cert = othercertificatedigest6631bf...\n" \
+"      set-dns = 1\n" \
+"      set-routes = 1\n" \
+"      pppd-use-peerdns = 1\n" \
+"      ca-file = /etc/openfortivpn/ca-bundle.pem\n" \
+"      user-cert = /etc/openfortivpn/user-key.pem\n" \
+"      user-key = /etc/openfortivpn/user-cert.pem\n" \
+"      insecure-ssl = 0\n" \
+"      cipher-list = HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4\n"
```
adrienverge: same as above

mrbaseman:
actually the undocumented config features were the decisive factor for me to make this pull request. I found the options by reading the source code, but I would expect them to be documented somewhere. I'm fine with removing some options from the sample config and from the help message, but I would keep them in the man page and refer to that place in the other files. Would that be ok for you?

adrienverge:
> I'm fine with removing some options from the sample config and from the help message, but I would keep them in the man page and refer to that place in the other files. Would that be ok for you?

That's great.

Thanks for your contribution!

PR #77, opened 2016-11-08
Be consistent and specify AF_INET when doing DNS lookups too.

dxtr:
This stops issues in NAT64 setups.


src/tunnel.c
```diff
@@ -177,17 +177,29 @@ int ppp_interface_is_up(struct tunnel *tunnel)
```
lkundrak: You ruin the witespace in indentation in the whole subroutine.

src/tunnel.c
```diff
@@ -177,17 +177,29 @@ int ppp_interface_is_up(struct tunnel *tunnel)
 
 static int get_gateway_host_ip(struct tunnel *tunnel)
 {
-	struct hostent *host = gethostbyname(tunnel->config->gateway_host);
-	if (host == NULL) {
-		log_error("gethostbyname: %s\n", hstrerror(h_errno));
-		return 1;
+    struct addrinfo *result = NULL;
+    //struct addrinfo *res = NULL;
```
lkundrak: No useless comments please.

src/tunnel.c
```diff
@@ -177,17 +177,29 @@ int ppp_interface_is_up(struct tunnel *tunnel)
...
+    struct addrinfo hints;
+    int err;
+
+    memset(&hints, 0, sizeof(struct addrinfo));
```
lkundrak: Please use a static initializer instead:

struct addrinfo hints = { .ai_family = AF_INET };

lkundrak:
I see that you fix the whitespace issues in a latter commit. Good, but please squash the commits into a single one (and get rid of the following merge commit).

DimitriPapadopoulos:
On the other hand the GETHOSTBYNAME(3) manual page reads:
> The **gethostbyname\*()**, **gethostbyaddr\*()**, **herror()**, and **hstrerror()** functions are obsolete.  Applications should use **getaddrinfo(3)**, **getnameinfo(3)**, and **gai_strerror(3)** instead.

DimitriPapadopoulos:
Also _gethostbyname()_ is not part of POSIX any more. See [The Open Group Base Specifications Issue 7, IEEE Std 1003.1-2008, 2016 Edition](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netdb.h.html#tag_13_30_09):
> The obsolescent _h_errno_ external integer, and the obsolescent _gethostbyaddr_() and _gethostbyname_() functions are removed, along with the HOST_NOT_FOUND, NO_DATA, NO_RECOVERY, and TRY_AGAIN macros.

The last POSIX version supporting _gethostbyname()_ is [The Open Group Base Specifications Issue 6, IEEE Std 1003.1, 2004 Edition](http://pubs.opengroup.org/onlinepubs/009695399/functions/gethostbyaddr.html#tag_03_225_03):
> The _gethostbyname_() function shall return an entry containing addresses of address family AF_INET for the host with name _name_.

So in theory _gethostbyname()_ is IPv4 only anyway. In practice it may attempt to support IPv6 on Linux.

I'm for the change, but it would be nice to know how “This stops issues in NAT64 setups.” I could squash the change using _Squash and merge_ myself, but we really need a better comment with respect to the NAT64 issue. It would be best to open an issue that describes the NAT64 problem you have been experiencing, refering to this pull request.

DimitriPapadopoulos:
@dxtr Could you perhaps describe the issue experienced with NAT64 in more detail? Which is the IPv4 network segment and which is the IPv6 network segment?

DimitriPapadopoulos:
See #206.

PR #83, opened 2016-12-19
mac os x port

mrbaseman:
I have rebased and squashed the mac os x port. 

The following commits are were merged into this one:
Added macOS build instructions to the readme (fretn)
more macOS stuff (fretn)
macOS documentation (fretn)
readme layout (fretn)
fix routing for 2 vpns under OS X (Michele Cella)
comply with coding style (mrbaseman)
correct indentation of enums with latest astyle (mrbaseman)

### DIFF
--- a/src/ssl.h
+++ b/src/ssl.h
@@ -31,6 +31,21 @@
 #include <openssl/ssl.h>
 #include <string.h>
 
+#ifdef __APPLE__
+/*
+ * Get rid of OSX 10.7 and greater deprecation warnings
+ * see for instance https://wiki.openssl.org/index.php/Hostname_validation
+ * this pragma selectively suppresses this type of warnings in clang
+ */
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+/*
+ * ERESTART is one of the recoverable errors which might be returned.
+ * However, in OSX this constant is not defined in errno.h so we define
+ * a dummy value here.
+ */
+#define ERESTART -1
+#endif
+
 #define ERR_SSL_AGAIN		0
 #define ERR_SSL_CLOSED		-1
 #define ERR_SSL_CERT		-2
--- a/src/tunnel.h
+++ b/src/tunnel.h
@@ -30,12 +30,23 @@
 #define _TUNNEL_H
 
 #include <openssl/ssl.h>
+#ifndef __APPLE__
 #include <semaphore.h>
+#endif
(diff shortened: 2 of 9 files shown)


src/ssl.h
```diff
@@ -31,6 +31,11 @@
 #include <openssl/ssl.h>
 #include <string.h>
 
+#ifdef __APPLE__
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#define ERESTART -1
```
adrienverge: Why this? Is it used somewhere?

If this is really needed, please add a small comment to explain.

src/tunnel.h
```diff
@@ -30,13 +30,19 @@
...
 #include <semaphore.h>
+#endif
 #include <unistd.h>
 
 #include "config.h"
 #include "io.h"
 #include "ipv4.h"
 
+#ifdef __APPLE__
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
adrienverge: Could you add a comment to explain the reason for this line?

mrbaseman:
I have added the requested comments. It seems to be a quite common practice to suppress these warnings when dealing with openssl on OS X. Of course it is a workaround until the currently deprecated functions are completely removed. Anyhow, currently it works and I would wait with code changes until they are really needed.

adrienverge:
@mrbaseman thanks.

My first comment was about `#define ERESTART -1`: is it used somewhere? If this is really needed, please add a small comment to explain why.

mrbaseman:
ERESTART is one of the recoverable errors which may occur, but this value is not standardized and on OSX it is not defined in errno.h, but it is used in the code for the openfortivpn linux client in ssl.h. It is one of three cases, which just won't occur on OSX. The simple approach is to define a dummy value so that the code compiles. I was thinking if it should be 85 like on linux systems, but on OSX 85 means EBADEXEC. Actually, -1 is used for ERESTART on OSX inside the kernel (https://opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/sys/errno.h), but not for normal applications. Therefore, we have to define it separately.

adrienverge:
Thanks a lot Martin!

mrbaseman:
@adrienverge: I should emphasize once more that most of the coding was done by @fretn. I just did a bit of code cleanup and documentation in this branch. Thanks for merging it into your master branch.

Issue #79, opened 2016-12-16
Two-factor authentication


schlatterbeck:
One of my customers has recently switched to two-factor auth using a security-token.
When connecting with the official client it asks for a second password after the first.
With openfortivpn I'm getting 
Fri Dec 16 12:08:18 2016 daemon.err openfortivpn[7389]: Could not authenticate to gateway (Bad HTTP response code).
(this message with my patch to allow log-output to syslog, see open pull request #65)
Are you interested in making 2-factor auth work and help me debug & patch openfortivpn?
I'd guess my first step would be to print the http response in the step that fails above...

Ralf

dsgwork:
@schlatterbeck yes please provide the HTTP requests/responses during the authentication phase (minus any sensitive information of course). It should help to diagnose what's going wrong.

The interesting code here is in the auth_log_in function in src/http.c. Please check if it's the first or second header check (i.e., before or during two-factor auth) that's returning ERR_HTTP_BAD_RES_CODE. 

If there's nothing obvious that jumps out from that info, it would also be very useful to have the HTTP requests/responses from the official fortinet client when connecting to the same VPN. For example, there's two code fields (code= and code2=). Since I have no idea what code2= is supposed to convey and it wasn't used in the VPNs I tested with when adding the two-factor support, we send it empty. Perhaps your gateway is expecting some data there? Having the dialog from the official client to compare should help determine that.

adrienverge:
@dsgwork Thanks for taking care of this :)

(I don't have access to a Fortinet VPN anymore)

schlatterbeck:
On Fri, Dec 16, 2016 at 04:28:29AM -0800, Dav�� Steinn Geirsson wrote:
> @schlatterbeck yes please provide the HTTP requests/responses during
> the authentication phase (minus any sensitive information of course).
> It should help to diagnose what's going wrong.
> 
> The interesting code here is in the auth_log_in function in
> src/http.c. Please check if it's the first or second header check
> (i.e., before or during two-factor auth) that's returning
> ERR_HTTP_BAD_RES_CODE. 

Thanks for the pointer.

I'm now debugging this in gdb.
The error appears in the first header check.
I'm getting (I've replace possibly sensitive fields with <censored>):

HTTP/1.1 401 Authorization Required
Content-Length: 1348
Keep-Alive: timeout=5, max=1000
Connection: Keep-Alive
Cache-Control: no-cache
Content-Type: text/html

<html>
<head>
[...]

PR #84, opened 2016-12-21
Add token-based one-time-password support

schlatterbeck:
See also issue #79

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -58,6 +58,7 @@ struct vpn_config {
 	uint16_t	gateway_port;
 	char		username[FIELD_SIZE + 1];
 	char		password[FIELD_SIZE + 1];
+	char		otp[FIELD_SIZE + 1];
 	char		cookie[COOKIE_SIZE + 1];
 	char            realm[FIELD_SIZE + 1];
 
--- a/src/http.c
+++ b/src/http.c
@@ -325,6 +325,140 @@ static int get_auth_cookie(struct tunnel *tunnel, char *buf)
 	return ret;
 }
 
+static
+int try_otp_auth(struct tunnel *tunnel, const char *buffer, char **res)
+{
+	char data[256];
+	char path[40];
+	char tmp [40];
+	char prompt [80];
+	const char *t = NULL, *n = NULL, *v = NULL, *e = NULL;
+	const char *s = buffer;
+	char *d = data;
+	const char *p = NULL;
+	/* Length-check for destination buffer */
+#define SPACE_AVAILABLE(sz) (sizeof (data) - (d - data) >= (sz))
+	/* Get the form action */
+	s = strcasestr (s, "<FORM");
+	if (s == NULL)
+		return -1;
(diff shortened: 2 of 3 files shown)


adrienverge:
Hi @schlatterbeck, thanks, your code looks great.

I just have two questions:
- Do you have the possibility to try this with a regular VPN (no 2FA, no OTP), to check it does not break anything in these cases? (@dsgwork [reported](https://github.com/adrienverge/openfortivpn/issues/79#issuecomment-268520474) it works with his SMS-based 2FA VPN.)
- If after merging this change breaks anything for other users (especially ones that connect to older Fortinet VPN versions), will you fix it?

schlatterbeck:
On Thu, Dec 22, 2016 at 12:02:29PM -0800, Adrien Verg� wrote:
> Hi @schlatterbeck, thanks, your code looks great.
> 
> I just have two questions:
> - Do you have the possibility to try this with a regular VPN (no 2FA,
> no OTP), to check it does not break anything in these cases? (@dsgwork
> [reported](https://github.com/adrienverge/openfortivpn/issues/79#issuecomment-268520474)

No, but I think this won't break anything for the following reason:

- The two-factor auth works like the normal auth with a user-chosen
  password in the first step. Only *after* this first step has completed
  my code kicks in.
- We used to compare the return code from the first auth step to http
  return-code 200 OK.
- The VPN requests the *second* authentication step (with the
  token-generated one-time password) by returning 401 Authorization
  Required instead of 200 OK after the first authentication.
- I've inserted a compare to 401 Authorization Required *before* the
  compare to 200 OK in Line 493 of http.c -- this should never match for
  one-factor authentication as this will return 200 OK after the first
  authentication step. Only at this point the new code is called.
- One possibility we *could* match 401 Auth Required here is if the user
  entered a wrong password (with only one-factor auth). My hunch was
  that maybe then the VPN could also return 401 Auth Required here. I've
  tested this and I'm getting a different return code with the message
  "No cookie given". I haven't instrumented the implementation to find
  out what http return code is returned here.

This leads me to believe the one-factor auth would work as normal but I
currently have no way to confirm this with a test.

> - If after merging this change breaks anything for other users
>   (especially ones that connect to older Fortinet VPN versions), will
>   you fix it?

Sure.

Ralf
-- 

adrienverge:
Thanks Ralf

mrbaseman:
Just for completeness: I have just tested the current master branch with a mobile token and client certificate (under Linux). It still works, so Ralfs code does not break anything, at least not for me.

schlatterbeck:
On Tue, Jan 03, 2017 at 05:27:27AM -0800, Martin Hecht wrote:
> Just for completeness: I have just tested the current master branch
> with a mobile token and client certificate (under Linux). It still
> works, so Ralfs code does not break anything, at least not for me.

Thanks for the feedback!

Ralf
-- 

adrienverge:
Thanks @mrbaseman!

Issue #25, opened 2015-11-24
Routes not correctly set


paride:
Now that #18 is fixed openfortivpn basically works for me, but I found that the the routing table is not configured correctly, while it is when using the proprietary client.

The proprietary client adds some specific routes for some networks, while openfortivpn just adds a default route, breaking my Internet connection (I already have a working default route that should not be changed).

As a workaround it is possible to use --no-routes and then add the relevant routes manually.
If you need more information on this issue just let me know. Thank you!

adrienverge:
Before setting the new default route, openfortivpn backs up the current default, and adds a route to the VPN gateway (using the current default). It seems like it's not working on your system: which GNU/Linux are you running? You could set openfortivpn in debug mode to get more info (and look for `ip route add/del` lines).

legovini:
Hi @adrienverge, I'm using Debian sid. This is my routing table before running the VPN client:

```
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.1.1     0.0.0.0         UG    600    0        0 wlan0
169.254.0.0     0.0.0.0         255.255.0.0     U     1000   0        0 wlan0
192.168.1.0     0.0.0.0         255.255.255.0   U     600    0        0 wlan0
```

The proprietary client brings to:

```
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.1.1     0.0.0.0         UG    600    0        0 wlan0
10.0.0.0        172.16.0.4      255.128.0.0     UG    0      0        0 ppp0
10.100.70.15    172.16.0.4      255.255.255.255 UGH   0      0        0 ppp0
10.128.0.0      172.16.0.4      255.128.0.0     UG    0      0        0 ppp0
93.XX.40.0      172.16.0.4      255.255.255.0   UG    0      0        0 ppp0
93.XX.40.1      192.168.1.1     255.255.255.255 UGH   0      0        0 wlan0
169.254.0.0     0.0.0.0         255.255.0.0     U     1000   0        0 wlan0
192.168.1.0     0.0.0.0         255.255.255.0   U     600    0        0 wlan0
```

where 93.XX.40.1 is the VPN remote endpoint. With openfortivpn I get:

```
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.1.1     0.0.0.0         UG    600    0        0 wlan0
[...]

PR #88, opened 2017-01-05
fix split routes route definition (issue #25)

Valantin:
this fix the route to use ppp ip as gateway for vpn subnet.
need to add the definition of a static route to reach the gateway if the gateway subnet match vpn subnet

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -281,8 +281,9 @@ static int ipv4_set_split_routes(struct tunnel *tunnel)
 		route = &tunnel->ipv4.split_rt[i];
 		strncpy(route_iface(route), tunnel->ppp_iface,
 		        ROUTE_IFACE_LEN - 1);
-		if (route_gtw(route).s_addr == -1)
+		if (route_gtw(route).s_addr == 0)
 			route_gtw(route).s_addr = tunnel->ipv4.ip_addr.s_addr;
+		route->rt_flags |= RTF_GATEWAY;
 		ret = ipv4_set_route(route);
 		if (ret == ERR_IPV4_SEE_ERRNO && errno == EEXIST)
 			log_warn("Route to gateway exists already.\n");


adrienverge:
I'm sorry but I don't understand what this pull request does.

The commit message needs to explain what it does, the reason why, and if we should expect any change for other users.

Thanks!

Valantin:
Issue #25 

the ip assigned by fortigate to my client is 10.30.0.5, the subnet to add route is 172.16.0.0/20

official client (working)
ip route
10.10.152.0/22 via 10.30.0.5 dev ppp0
172.16.0.0/20 via 10.30.0.5 dev ppp0

openfortivpn (not working)
ip route
10.10.152.0/22 dev ppp0 scope link
172.16.0.0/20 dev ppp0 scope link

openfortivpn patched (working)
ip r
10.10.152.0/22 via 10.30.0.5 dev ppp0
172.16.0.0/20 via 10.30.0.5 dev ppp0 


My code add flag RTF_GATEWAY to add "via" in the route table, check if "s_addr == 0" (not -1, is unsigned int) to add "10.30.0.5" if no gateway set

no change expected for other user.

mrbaseman:
I have created a testing branch and at a first glance this patch doesn't break things in my setup

adrienverge:
I'm sorry but I still don't understand what this does: is it a bug fix? If yes, what's the bug? How does it solve it? Please provide a *clear* explanation. And the commit message needs to contain this explanation.

mrbaseman:
actually, for me the "via" part is added to the routes without this patch already. @Valantin which versions of openfortivpn did you compare? A quite recent checkout from master with and without the proposed patch?

src/ipv4.c
```diff
@@ -281,8 +281,9 @@ static int ipv4_set_split_routes(struct tunnel *tunnel)
 		route = &tunnel->ipv4.split_rt[i];
 		strncpy(route_iface(route), tunnel->ppp_iface,
 		        ROUTE_IFACE_LEN - 1);
-		if (route_gtw(route).s_addr == -1)
+		if (route_gtw(route).s_addr == 0)
```
mrbaseman: This is the decision if the route has already a gateway set or not. -1 casted to an unsigned int would result in UINT_MAX which is 255.255.255.255 in IP notation. Routes without a gateway however carry the gateway IP 0.0.0.0 which is the proposed 0 here. With -1 the following line would (normally) never be executed. So this is a bugfix.

src/ipv4.c
```diff
@@ -281,8 +281,9 @@ static int ipv4_set_split_routes(struct tunnel *tunnel)
...
 			route_gtw(route).s_addr = tunnel->ipv4.ip_addr.s_addr;
+		route->rt_flags |= RTF_GATEWAY;
```
mrbaseman: this adds the gateway flag to the routes received from the Fortigate. Without the gateway flag the destination is directly connected (over the vpn tunnel), which is not always the case. There might be more hops behind the remote gateway to reach the network to which this route points.

Valantin:
@mrbaseman I've tested the master branch with a fortigate 800c with fortios 5.0
My VPN is configured to use a /16 subnet instead the default ip pool for client connection

@adrienverge sorry I can't explain better the problem.
In my setup this partially fix issue #25 the problem is described in the issue

mrbaseman:
@Valantin: Thanks for your reply. In the meantime I understand better what these changes are good for (see my comments on the changes).
@adrienverge: Does this make sense to you? As you can see, I have also pulled these changes into my other work on the routing code

Valantin:
@mrbaseman I've tried your ppp-routes branch, It seems to fix issue #25 
On my setup, fortigate gateway is in the same /23 routed by fortigate vpn, No more need to add manualy the /32 route

adrienverge:
@mrbaseman thanks for taking the time to explain this change.

@Valantin: as said before
> The commit message needs to explain what it does, the reason why, and if we should expect any change for other users.

> the commit message needs to contain this explanation.

it is important to be able to track changes and `git blame` problems using the git history (and impendently of GitHub).

mrbaseman:
now that #95 has been merged this one can be closed. I have picked up the proposed changes in the larger rework of the routing code.

PR #89, opened 2017-01-14
Fix build error with clang

Mabin-J:
It has build error in Mac OSX Sierra with HomeBrew
It also has build error in Linux with clang (of cource).


adrienverge:
Looks good. Thanks!

PR #90, opened 2017-01-17
fix detecting new connection bug when ppp connections are existed.

Mabin-J:
When ppp connections are existed before connect with openfortivpn, openfortivpn use only 'ppp0'.
So, after connect vpn with openfortivpn that situation, routing table get tangled.

There is also same situation when connecting one more vpn with only openfortivpn.

So, I changed checking new connection logic.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -165,6 +165,8 @@ int ppp_interface_is_up(struct tunnel *tunnel)
 {
 	struct ifaddrs *ifap, *ifa;
 
+	log_debug("Got Address: %s\n", inet_ntoa(tunnel->ipv4.ip_addr));
+
 	if (getifaddrs(&ifap)) {
 		log_error("getifaddrs: %s\n", strerror(errno));
 		return 0;
@@ -173,10 +175,21 @@ int ppp_interface_is_up(struct tunnel *tunnel)
 	for (ifa = ifap; ifa != NULL; ifa = ifa->ifa_next) {
 		if (strstr(ifa->ifa_name, "ppp") != NULL
 		    && ifa->ifa_flags & IFF_UP) {
-			strncpy(tunnel->ppp_iface, ifa->ifa_name,
-			        ROUTE_IFACE_LEN - 1);
-			freeifaddrs(ifap);
-			return 1;
+			if (&(ifa->ifa_addr->sa_family) != NULL
+			    && ifa->ifa_addr->sa_family == AF_INET) {
+				struct in_addr if_ip_addr =
+				        cast_addr(ifa->ifa_addr)->sin_addr;
+
+				log_debug("Interface Name: %s\n", ifa->ifa_name);
+				log_debug("Interface Addr: %s\n", inet_ntoa(if_ip_addr));
+
+				if (tunnel->ipv4.ip_addr.s_addr == if_ip_addr.s_addr) {
+					strncpy(tunnel->ppp_iface, ifa->ifa_name,
+					        ROUTE_IFACE_LEN - 1);
+					freeifaddrs(ifap);
+					return 1;
+				}
+			}
 		}
 	}
 	freeifaddrs(ifap);


mrbaseman:
I have created a testing branch and at a first glance this doesn't break things in my setup. I'm also interested in running openfortivpn over an LTE modem connection which itself uses ppp0 already. However, I haven't tested that, yet.

Mabin-J:
@mrbaseman 
For showing wrong case, 
I fix 'ppp_interface_is_up' with my code base for get interface's address type.
(because original logic return earlier than my logic)
```c
int ppp_interface_is_up(struct tunnel *tunnel)
{
        struct ifaddrs *ifap, *ifa;

        log_debug("Got Address: %s\n", inet_ntoa(tunnel->ipv4.ip_addr));

        if (getifaddrs(&ifap)) {
                log_error("getifaddrs: %s\n", strerror(errno));
                return 0;
        }

        for (ifa = ifap; ifa != NULL; ifa = ifa->ifa_next) {
                log_debug("Interface Name: %s\n", ifa->ifa_name);
                if (&(ifa->ifa_addr->sa_family) != NULL){
                        log_debug("Interface's Address Type: %d\n", ifa->ifa_addr->sa_family);
                } else {
                        log_debug("Interface's Address Type: NULL\n");
                }

                if (strstr(ifa->ifa_name, "ppp") != NULL
                    && ifa->ifa_flags & IFF_UP) {
                        if (&(ifa->ifa_addr->sa_family) != NULL
                            && ifa->ifa_addr->sa_family == AF_INET) {
                                struct in_addr if_ip_addr =
                                        ((struct sockaddr_in *) ifa->ifa_addr)->sin_addr;

                                log_debug("Interface Addr: %s\n", inet_ntoa(if_ip_addr));

                                if (tunnel->ipv4.ip_addr.s_addr == if_ip_addr.s_addr) {
                                        strncpy(tunnel->ppp_iface, ifa->ifa_name,
                                                ROUTE_IFACE_LEN - 1);
                                        freeifaddrs(ifap);
                                        return 1;
                                }
                        }
                }
        }
        freeifaddrs(ifap);

        return 0;
}
```

and I connected one fortivpn and try to connect one more fortivpn.  
This is debug message of second one.
```
DEBUG:  Got Address: XXX.XXX.XXX.XX1
DEBUG:  Interface Name: lo
DEBUG:  Interface's Address Type: 17
DEBUG:  Interface Name: enp4s0f0
DEBUG:  Interface's Address Type: 17
DEBUG:  Interface Name: wlp3s0
DEBUG:  Interface's Address Type: 17
DEBUG:  Interface Name: wlx10feed2529d9
DEBUG:  Interface's Address Type: 17
DEBUG:  Interface Name: br0
DEBUG:  Interface's Address Type: 17
DEBUG:  Interface Name: vboxnet0
DEBUG:  Interface's Address Type: 17
DEBUG:  Interface Name: ppp0
DEBUG:  Interface's Address Type: NULL
DEBUG:  Interface Name: ppp1
DEBUG:  Interface's Address Type: NULL
DEBUG:  Interface Name: lo
DEBUG:  Interface's Address Type: 2
DEBUG:  Interface Name: br0
DEBUG:  Interface's Address Type: 2
DEBUG:  Interface Name: ppp0
DEBUG:  Interface's Address Type: 2
DEBUG:  Interface Addr: XXX.XXX.XXX.XX0
DEBUG:  Interface Name: ppp1
DEBUG:  Interface's Address Type: 2
DEBUG:  Interface Addr: XXX.XXX.XXX.XX1
INFO:   Interface ppp1 is UP.
```

If use original code, 'ppp_interface_is_up' select 'ppp0' because check only interface name with 'ppp' this point.
```
DEBUG:  Interface Name: ppp0
DEBUG:  Interface's Address Type: NULL
```

I think that order of interface list of 'getifaddrs' function's return value can be different each machine. so I think that checking address is more correct than interface name.

Mabin-J:
@mrbaseman 
Sorry. I thought you said that original logic is run normally caused my low english skill... 
^^;;;;;

mrbaseman:
On the weekend I found the time to test this with my lte modem. Without your patch openfortivpn uses ppp0 for both, the modem traffic uplink and the vpn channel, which of course does not work. 
With your patch it correctly allocates a ppp1 device but fails to set up routes:

INFO:   Connected to gateway.
INFO:   Authenticated.
INFO:   Got addresses: [10.212.XXX.YYY], ns [192.168.XXX.YYY, 8.8.8.8]
INFO:   Got addresses: [10.212.XXX.YYY], ns [192.168.XXX.YYY, 8.8.8.8]
INFO:   Got addresses: [10.212.XXX.YYY], ns [192.168.XXX.YYY, 8.8.8.8]
INFO:   Interface ppp1 is UP.
INFO:   Setting new routes...
WARN:   Could not set route to tunnel gateway (Invalid argument).
INFO:   Adding VPN nameservers...
INFO:   Cancelling threads...
INFO:   Setting ppp interface down.
INFO:   Restoring routes...
WARN:   Could not delete route through tunnel (No such device).
WARN:   Could not delete route to gateway (No such process).
INFO:   Removing VPN nameservers...
INFO:   Terminated pppd.
INFO:   Logged out.

it might be due to the fact that my Fortigate pushes a default route throgh the vpn channel (to ensure that the vpn also encrypts my surf traffic when I'm on an insecure network) - or I might have missed the latest commit (I'm not sure anymore, when I have pulled the source that I have used for testing). I'll re-test and give you an update.

mrbaseman:
ah, I just noticed that I have probably merged PR #88 into the code I was using for testing and I might have hit a problem with that code or some interference between the two pull requests. I definitely have to re-test.

adrienverge:
Hi, this looks good!

I'd be happy to merge once it is reported to work.

mrbaseman:
I don't have split routes configured on my fortigate, That means all traffic goes through the vpn connection. When I manually add a host route over ppp0 for my fortigate before starting openfortivpn this works for me. The traffic of the vpn channel goes over ppp0 and everything else is routed over the new ppp1. When split routes is configured this patch should work out of the box but I haven't tested that yet. It would be nice if the host route could be added automatically when starting openfortivpn.

mrbaseman:
I have just verified that when no default route is pushed by the fortigate this patch works fine. Adding a host route over ppp0 in the other case could be added in a separate commit.

adrienverge:
Thanks a lot @Mabin-J and @mrbaseman for testing!

Issue #85, opened 2017-01-03
diagnostic message colors invisible in gnome-terminal


RBKreckel:
I just discovered that OpenFortiVPN actually prints diagnostic messages, not just some gratuitous empty lines! The problem is that these diagnostic messages are hardly visible in gnome-terminal in the standard light system theme. Here is a screenshot.
![openfortivpn-in-gnome-terminal](https://cloud.githubusercontent.com/assets/24757097/21605817/c7da3428-d1ab-11e6-8b73-e40ed54c399a.png)

adrienverge:
Hi @RBKreckel,

I also have a light-background theme but I clearly see the debug lines in gnome-terminal:
![gnome-terminal](https://cloud.githubusercontent.com/assets/5244945/22325454/631c72a2-e3ae-11e6-8c46-33e85424a409.png)

I believe this is a problem with your current theme, because `\033[0;90m` colors (the ones used for `log_debug`) are meant to be seen. All standard light themes I have in gnome-terminal (black on white, black on light yellow, solarized light) display it correctly.

You could try to use one of those standard themes, or keep the one you have but go to Edit > Profile > Colors and change the 9th one of the palette (the one at bottom left) to something more contrasted with your background.

Hope this helps!

adrienverge:
Forget my previous message, I thought you were talking about debug messages, not info ones!

This is actually an issue. Will fix it.

RBKreckel:
Great. Thanks!

PR #91, opened 2017-01-17
fix for #85 diagnostic message colors invisible

mrbaseman:
switching the foreground color to white might not be a good idea on a bright background. Switching to bold looks nearly the same on a dark background and the characters stay visible on bright background as a fix for issue #85


mrbaseman:
I have also applied it to my new testing branch. I wasn't aware that github even notifies the existing pull requests about commits to other branches. I have just seen it after pushing, so sorry for the noise.

adrienverge:
Don't worry for the noise: these updates don't send any email notifications ;)

Thanks for this but I don't like having infos in bold while errors and warnings are thin.

What about simply using the default color for info messages?

```
{ "INFO:   ", "",  LOG_INFO}
```

mrbaseman:
sure. I have updated the pull request (maybe this is a bit overshooting for changing a single line of code)

adrienverge:
Thanks!

Issue #94, opened 2017-01-30
set up routes correctly when ppp0 interface is already there


mrbaseman:
as discussed in pull request #90 which implemented that a new ppp device is used when ppp0 is already there, I have noticed a problem when the vpn pushes a new default route over the new ppp1 interface.
In that particular case adding the gateway route fails in ipv4_set_default_routes with the following debug output:

DEBUG:  ip route add to A.B.C.D/255.255.255.255 via 0.0.0.0 dev ppp0
WARN:   Could not set route to tunnel gateway (Invalid argument).

Adding a host route over ppp0 before starting openvpn makes the tunnel work. The above warning still appears then, but the tunnel works. Comparing the routing entries I found that in this case the flags are sligtly different. Instead of 
	gtw_rt->rt_flags |= RTF_GATEWAY;
one would have to put:
	gtw_rt->rt_flags |=  RTF_HOST;
However, when I try to open a vpn connection to another gateway, which does not push a default route, the code is needed as it is. This is in the early stage, before the routes etc. are received, so I'm not sure how to handle this decision. Maybe we should add an option to the config file (something like is-gateway or host-route or additional-flags...) which is used here and allows us to control the setup of the routes. Any suggestions? I'm willing to implement and test this, but @adrienverge, I would like to have your opinion which path I should follow.

Mabin-J:
I think that changing [321st line of ipv4.c](https://github.com/adrienverge/openfortivpn/blob/5cb432beca51f777ff97b036d9352d46e08e488b/src/ipv4.c#L321) is enough for fix issue of warning message because 'ipv4_set_default_routes' function is only called when it is set 'default gateway mode(?)'.

Apart from this issue, I think that adding a route for VPN server is needed when it is set not only 'default gateway mode' but also 'split routes mode(?).'  
When I connect 'A.B.C.D' IP of VPN Server that contains 'A.B.C.0/24' route in 'split routing table,' Packets cannot be passed through VPN because 'A.B.C.0/24' route.
So, I have a plan that I'll insert logic that adding a default route for server's IP.

mrbaseman:
@Mabin-J: Thanks for your input. You are right: 'ipv4_set_default_routes' function is only called when the config on the Fortigate contains 
set split-tunneling disable
[...]

PR #95, opened 2017-02-01
set up routes correctly (pull request for #94)

mrbaseman:
This is a larger rework of the routing code, here a summary of the changes (for a detailed discussion see in issue #94 ):
- in ipv4_set_split_routes: add a host route to the gateway (basically copy&paste code from ipv4_set_default_routes)
- also in  ipv4_set_split_routes: merge in the two fixes for #25, namely add the gateway flag to these routes and corrected the test of the gateway ip (I have merged in pull request #88 here, for details see there). @adrienverge: if you want to merge this as a separate pull, please let me know and I can undo the changes in my branch.
- in ipv4_set_default_routes: change the route flag to host instead of gateway for the  route to the Fortigate. This is important when the existing default route is over a ppp device.
- in ipv4_restore_routes: remove the host route to the gateway which is now added in gateway mode, and free resources (just work that has to be done now that the above changes have been implemented).
- in ppp_interfaces_is_up: here are the changes from #90 which has already been merged. This actually was the starting point for the whole work. Since this is already upstream, these changes can be removed by rebase and squash of my whole pull request. I just did not do this yet, because I first thought that comments on the individual commits would stay intact, which was not the case.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -275,14 +275,46 @@ static int ipv4_set_split_routes(struct tunnel *tunnel)
 {
 	int i;
 	struct rtentry *route;
+	struct rtentry *gtw_rt = &tunnel->ipv4.gtw_rt;
+	struct rtentry *def_rt = &tunnel->ipv4.def_rt;
 	int ret;
+
+	route_init(def_rt);
+
+	// Back up default route
+	route_dest(def_rt).s_addr = inet_addr("0.0.0.0");
+	route_mask(def_rt).s_addr = inet_addr("0.0.0.0");
+
+	ret = ipv4_get_route(def_rt);
+	if (ret != 0) {
+		log_warn("Could not get current default route (%s).\n",
+		         err_ipv4_str(ret));
+		goto err_destroy;
+	}
+
+
+	// Set the default route as the route to the tunnel gateway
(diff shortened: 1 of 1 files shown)


mrbaseman:
I just noticed that the comments which I have placed on the individual commits all showed up above, which didn't make much sense. So I have removed them. So maybe I should just rebase and squash?

adrienverge:
Thanks @mrbaseman, this looks good!

Yes it's better to squash and have one commit for one logical change, but I would do it anyway before merging.

Most importantly, you need to explain the change *in the commit message itself*, so it is possible to understand why code changed by looking at the git history. This allows busting bugs more efficiently, avoid mistakes, enable useful `git blame`, etc. Also if this project moves from GitHub some day, we'll still have this information in git itself. (If you do a `git log` on this repo, you'll see that all non-trivial commits do this.)

About #88: do as you prefer, either include it as a separated commit here or wait for #88 to be merged.

mrbaseman:
Hi @adrienverge,

I have rebased and squashed the pull request and added the above explanations to the commit message. I have also inserted a few comments from #90. I have kept the changes of #88 in this pull request because it is also part of the fix for the routing issues. I have also included my comments from there in the commit message.

adrienverge:
Thanks!

PR #97, opened 2017-02-08
Fix to remain exist route & merge duplicated logic

Mabin-J:
Fix to remain route to vpn server when disconnect if route to vpn server is already exist before connect.

And merge duplicated logic for adding route to vpn server.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -237,6 +237,52 @@ static int ipv4_del_route(struct rtentry *route)
 	return 0;
 }
 
+int ipv4_protect_tunnel_route(struct tunnel *tunnel)
+{
+	struct rtentry *gtw_rt = &tunnel->ipv4.gtw_rt;
+	struct rtentry *def_rt = &tunnel->ipv4.def_rt;
+	int ret;
+
+	route_init(def_rt);
+
+	// Back up default route
+	route_dest(def_rt).s_addr = inet_addr("0.0.0.0");
+	route_mask(def_rt).s_addr = inet_addr("0.0.0.0");
+
+	ret = ipv4_get_route(def_rt);
+	if (ret != 0) {
+		log_warn("Could not get current default route (%s).\n",
+		         err_ipv4_str(ret));
+		goto err_destroy;
+	}
+
+
+	// Set the default route as the route to the tunnel gateway
+	memcpy(gtw_rt, def_rt, sizeof(*gtw_rt));
+	route_dest(gtw_rt).s_addr = tunnel->config->gateway_ip.s_addr;
+	route_mask(gtw_rt).s_addr = inet_addr("255.255.255.255");
+	gtw_rt->rt_flags |= RTF_HOST;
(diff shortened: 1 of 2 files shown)


src/ipv4.c
```diff
@@ -323,49 +335,14 @@ static int ipv4_set_split_routes(struct tunnel *tunnel)
...
 
 static int ipv4_set_default_routes(struct tunnel *tunnel)
 {
 	int ret;
 	struct rtentry *def_rt = &tunnel->ipv4.def_rt;
-	struct rtentry *gtw_rt = &tunnel->ipv4.gtw_rt;
 	struct rtentry *ppp_rt = &tunnel->ipv4.ppp_rt;
 
-	route_init(def_rt);
-	route_init(ppp_rt);
```
mrbaseman: I believe the `route_init(ppp_rt);` got lost here

Mabin-J: Oh... yes. my mistake...

adrienverge:
Looks good. Can you confirm it's tested and working before I merge?

mrbaseman:
Bon soir @adrienverge,

@Mabin-J has corrected the issue with route_init in the meantime, but I haven't tested this pull request  yet. The main reason: @Mabin-J  has found more issues on Mac OSX and also a double free error when adding the default route fails. He has proposed #99 which also implements the fix for the situation that the gateway route existed beforehand, but on the other hand does not merge the duplicate logic of the two modes in a new function. The two pull requests are to some extend orthogonal and I would hesitate to merge #97 in favor of #99. Maybe @Mabin-J can comment on this?

I haven't tested #99 either, but it looks good to me so far. I have spent some time trying to merge both approaches until I realized that the logic between the two modes has diverged too much to pull it out into a separate function in #99. If I find the time tomorrow, I can test and maybe I can get some results on OSX as well.

regards, Martin

Mabin-J:
I created #99 pull request because #99 is more important than this issue.  
And I think that this request need to be checked by @mrbaseman because I changed many codes that is written by @mrbaseman .

If you want to merge #99 and this both, I'll merge two pull request.

One more, I'm finding way to operate 'ipv4_get_route' in MacOS.  
Maybe 'ipv4_get_route' for MacOS will be completed until tomorrow.
I planned that I'll create another pull request for this but if you want to merge all my pull request, I'll merge all my request.

mrbaseman:
I have just verified that this works in my setup with split routes. If the host route to the Fortigate exists before starting openfortivpn, there is a warning that the route is already there, and the route is not deleted when the vpn channel is closed. When the route is not there before starting, the route is added and also removed again when the channel is terminated. 
@Mabin-J you are doing a great work here. From my point of view, either #97 or #99 can be merged, and the other one needs to be rebased (and the conflicts occurring then need to be resolved manually). I think it is up to @adrienverge to decide in which order he likes the fixes to be applied. And when we have another fix for `ipv4_get_route` in MacOS this would be really great.

adrienverge:
Thanks, squashing and merging

PR #99, opened 2017-02-09
segment error when openfortivpn fail to get default route

Mabin-J:
When openfortivpn fail to get default route, segment error is shown when disconnect it.  
This segment error caused 'def_rt' freed one more.

I found this error in MacOS. 
MacOS doesn't have '/proc' partition so openfortivpn cannot use 'ipv4_get_route' function in MacOS.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -381,11 +381,10 @@ int ipv4_set_tunnel_routes(struct tunnel *tunnel)
 {
 	int ret = ipv4_protect_tunnel_route(tunnel);
 
-	if (ret == 0) {
-		if (tunnel->ipv4.split_routes)
-			return ipv4_set_split_routes (tunnel);
-		else
-			return ipv4_set_default_routes (tunnel);
+	if (tunnel->ipv4.split_routes)
+		return ipv4_set_split_routes (tunnel);
+	else if (ret == 0) {
+		return ipv4_set_default_routes (tunnel);
 	} else {
 		return ret;
 	}
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -279,6 +279,7 @@ int ipv4_protect_tunnel_route(struct tunnel *tunnel)
 
 err_destroy:
 	route_destroy(def_rt);
+	tunnel->ipv4.route_to_vpn_is_added = 0;
 
 	return ret;
 }
@@ -400,28 +401,30 @@ int ipv4_restore_routes(struct tunnel *tunnel)
 		if (ret != 0)
 			log_warn("Could not delete route to vpn server (%s).\n",
 			         err_ipv4_str(ret));
-	} else {
-		log_debug("Route to vpn server is not added\n");
-	}
 
-	if (tunnel->ipv4.split_routes)
-		goto out;
+		if (tunnel->ipv4.split_routes)
+			goto out;
 
-	ret = ipv4_del_route(ppp_rt);
-	if (ret != 0)
-		log_warn("Could not delete route through tunnel (%s).\n",
-		         err_ipv4_str(ret));
+		ret = ipv4_del_route(ppp_rt);
+		if (ret != 0)
+			log_warn("Could not delete route through tunnel (%s).\n",
+			         err_ipv4_str(ret));
 
-	// Restore the default route
-	// It seems to not be automatically restored on all linux distributions
-	ret = ipv4_set_route(def_rt);
-	if (ret != 0)
-		log_warn("Could not restore default route (%s). Already restored?\n",
-		         err_ipv4_str(ret));
+		// Restore the default route
+		// It seems to not be automatically restored on all linux distributions
+		ret = ipv4_set_route(def_rt);
+		if (ret != 0) {
+			log_warn("Could not restore default route (%s). \
+Already restored?\n",
+			         err_ipv4_str(ret));
+		}
 
-	route_destroy(ppp_rt);
+		route_destroy(ppp_rt);
 out:
-	route_destroy(def_rt);
+		route_destroy(def_rt);
+	} else {
+		log_debug("Route to vpn server was not added\n");
+	}
 
 	return 0;
 }
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -280,6 +280,8 @@ int ipv4_protect_tunnel_route(struct tunnel *tunnel)
 err_destroy:
 	route_destroy(def_rt);
 	tunnel->ipv4.route_to_vpn_is_added = 0;
+	log_warn("Protecting tunnel route has faild. \
+But this can be working except for some cases.\n");
 
 	return ret;
 }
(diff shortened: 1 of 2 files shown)


mrbaseman:
this works for me on linux with split routes. I couldn't test Mac OS yet. However, the route to the gateway is still removed in my test: The `ipv4_get_route` works on linux and when it goes into the branch "Setting route to tunnel gateway..." in `ipv4_set_split_routes` there are two cases left in which another `tunnel->ipv4.route_to_vpn_is_added = 0;`  is needed. However, this could be added with an updated version of #97 later on.

adrienverge:
Hi @Mabin-J,

Could you be more precise on the problem? What segfaults exactly? If `/proc` isn't present, then `ipv4_get_route()` should nicely return `ERR_IPV4_SEE_ERRNO` isn't it?

On MacOS, is there a stable way to get the default route?

Why a different behavior between `ipv4_set_default_routes()` and `ipv4_set_split_routes()`?  With your change, in `ipv4_set_split_routes()`, it seems that you're using `def_rt` even though `ipv4_get_route(def_rt)` just failed: is it safe?

Also, you don't handle error codes correctly, so if `ipv4_get_route()` fails (on Linux, for another reason than `/proc`), `ipv4_set_route()` will be called while it shouldn't be, and `def_rt` won't be destroyed.

Mabin-J:
Apply merged #97 .

Mabin-J:
segment fault is caused that ```route_destroy(def_rt);``` is always called when disconnect.
(https://github.com/adrienverge/openfortivpn/blob/e212d0ef625d24d3b02561a22f8c02de731f7451/src/ipv4.c#L424)

when ```ipv4_protect_tunnel_route``` fail, ```route_destroy(def_rt);``` is called already in ```ipv4_protect_tunnel_route``` and when try to disconnect, ```route_destroy(def_rt);``` is called in ```ipv4_restore_routes```.

Mabin-J:
And.. I had a stable way to get the default route but I need more test.

src/ipv4.c
```diff
@@ -279,6 +279,9 @@ int ipv4_protect_tunnel_route(struct tunnel *tunnel)
 
 err_destroy:
 	route_destroy(def_rt);
+	tunnel->ipv4.route_to_vpn_is_added = 0;
+	log_warn("Protecting tunnel route is fail. \
+But this can be working except some case.\n");
```
mrbaseman: I'd suggest an improvement of the language here: 
"Protecting tunnel route has failed. \
But this can be working except for some cases."

src/ipv4.c
```diff
@@ -400,28 +402,30 @@ int ipv4_restore_routes(struct tunnel *tunnel)
...
+			         err_ipv4_str(ret));
+		}
 
-	route_destroy(ppp_rt);
+		route_destroy(ppp_rt);
 out:
-	route_destroy(def_rt);
+		route_destroy(def_rt);
+	} else {
+		log_debug("Route to vpn server is not added\n");
```
mrbaseman: "...was not added" would be better, because we are in the cleanup routine that is called at the end of the session here

mrbaseman:
what about the cleanup in these cases in ipv4_protect_tunnel_route on lines [272](https://github.com/Mabin-J/openfortivpn/blob/1bb1534cc93ec8430c803f9ac1be710e998e0750/src/ipv4.c#L272) and [276](https://github.com/Mabin-J/openfortivpn/blob/1bb1534cc93ec8430c803f9ac1be710e998e0750/src/ipv4.c#L276) ? shouldn't def_rt be freed in these cases, too? Maybe we should log the different warnings and `goto err_destroy:` for all failing cases of this function? (Caution, in the case of L276 additional curly braces are needed)
PS: I agree with your explanation for the segment error. This was my mistake when I introduced the cleanup of def_rt for the "split routes" case. I didn't think of the case that adding the route may have failed before. My apologies.

src/ipv4.c
Mabin-J: I fixed. Thank you.

src/ipv4.c
Mabin-J: I fixed. Thank you.

mrbaseman:
@Mabin-J did you have a chance to test your stable way to get the default route on Mac?

Mabin-J:
@mrbaseman Yes. I already tested that. Not only me but also my coworkers was using openfortivpn that is applied fixed code.

mrbaseman:
@Mabin-J I assume you are talking about your dev-branch here. Could you rebase that on the current master and create a pull request so that this could be reviewed more easily and maybe make its way into the upstream code? We are discussing here in a pull request about your hotfix-fail-route branch, but if I remember right this particular issue was fixed in the course of your pull request #97 - except for the still remaining issues with OSX

Mabin-J:
@mrbaseman, Oh, I missed understand.
I found way for get default route in MacOSX. 
It use parts of 'netstat' command's source in MacOSX.
but I was still testing.

And I already rebase this branch after merged #97 in 3 months ago. It is not applied only #110 #117.
I'll apply soon.

Mabin-J:
And, in MacOSX, since merged #95 and #97, 'split route mode' doesn't run correctly because 'ipv4_get_route()' doesn't run correctly. 
ef98b98 is for this issue.

mrbaseman:
I was looking into this today and have still found problems with the memory management. I hope that I can finish a solution (based on your pull plus a few more fixes) soon. 
About the Mac OSX port: Wouldn't it be easier to parse the output of route or netstat instead of pulling parts of the source into openfortivpn? Alternatively, isn't there anything like /proc/net/route in OSX, maybe below a different path?

Mabin-J:
@mrbaseman,
I'll check more source about memory management. but at least there is no more segment fault in Linux and MacOSX both.

And, I think that you are right. Parsing message of 'netstat -nr' is easier way. (in MacOSX, 'route' command doesn't have way for printing routing table) 
I just thought that getting route info directly from kernel is more correct than parsing message.

I know that kernel of MacOSX is based on BSD kernel. It doesn't have '/proc' partition. (FreeBSD doesn't have '/proc' partition, too and I know that most UNIX also doesn't have that.)

mrbaseman:
@Mabin-J before we do the same work twice, have a look at my version of this change: https://github.com/mrbaseman/openfortivpn/tree/hotfix-fail-route 

One problem was that the rt_dev entry got copied and the buffer from one entry was used in another, and also these pointers were not carefully checked and set to NULL when freed. This probably was correct at some point, but got broken when commits for different parts of the ipv4 code got merged together. Also, I have replaced the `goto out` jump in  ipv4_restore_routes by a more elegant `if` construct.

The problem that I mentioned yesterday was: `route_to_vpn_is_added` was not a reliable indicator to determine if route_destroy should be called or not, because there were all kinds of combinations possible: route entry already freed, but the flag not set, route entry still there and flag not set, etc. ... 
With the checks in `route_destroy` it can be safely called several times on the same entry now, without causing double free errors.

mrbaseman:
merged together with the above described changes

PR #102, opened 2017-02-13
Load OS trusted certificate stores

berenddeschouwer:
Load the OS trusted certificate stores using
the default paths that are compiled in to
the OS-shipped OpenSSL libraries.


berenddeschouwer:
This is necessary on stock RHEL 6,7 / Fedora 25 for CA-signed certificates.

PR #120, opened 2017-05-04
README.md: add 'Known Issue'

Mabin-J:
Since started support MacOSX, 'default route mode' isn't supported in MacOSX but it isn't noticed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,13 @@ Cmnd_Alias  OPENFORTIVPN = /usr/bin/openfortivpn
 described in #54, a malicious user could use `--ppp-plugin` and `--ppd-log`
 options to divert the program's behaviour.
 
+
+Known Issue
+-----------
+
+* 'Default Route Mode' doesn't work in Mac OS X
+
+
 Contributing
 ------------
 


mrbaseman:
would it be sufficient to modify `ipv4_get_route()` such that it works for OSX? I'd prefer fixing that because it also solves problems with the split route mode

mrbaseman:
Here is my proposal for a fallback of ipv4_get_route using the output of netstat on MacOSX: https://github.com/mrbaseman/openfortivpn/commit/07d866c3c934cc889563ad14e402a1cfa78fb609
Unfortunately, I can't really test this, because I don't own a Mac. I have written the code based on what I have found on the net. With a few tweaks I could make my compiler process the code on linux, but there might still be a few issues (e.g. different path of the headers or alike) in the real environment.... So, if anyone could test this on OSX, please?

Mabin-J:
@mrbaseman, Thank you for your comment.
I'll test when I arrive my office tomorrow.

mrbaseman:
I think we can close this, because we have a working ipv4_get_route on MacOSX now, which was just merged into master

Issue #119, opened 2017-04-27
Regression: Connection on OSX has broken since v1.3.0


SimonKagstrom:
With current git HEAD, I see these messages

```
INFO:   Setting new routes...
WARN:   Could not get current default route (No such file or directory).
INFO:   Adding VPN nameservers...
```

which seems to work with v1.3.0.

DimitriPapadopoulos:
The warning is from this piece of code in **ipv4.c**:
```
	ret = ipv4_get_route(def_rt);
	if (ret != 0) {
		log_warn("Could not get current default route (%s).\n",
		         err_ipv4_str(ret));
		goto err_destroy;
	}
```
Function `ipv4_get_route()` does seem to have changed much since commit 5cb432beca51f777ff97b036d9352d46e08e488b “Mac OS X port”.

It might help - or not - if you could run `openfortivpn` with at least options `-v -v`.

mrbaseman:
@mabin has a solution for this (see #99). However, we have to see how this fits into the current master branch.

mrbaseman:
Sorry, I meant @mabin-J in my previous comment

Mabin-J:
@DimitriPapadopoulos,
First of all, 'ipv4_get_route()' isn't changed since 5cb432b.

WARN message and broken 'split route mode' in MacOSX caused that #95, #97 is applied.
#95, #97 are for protecting route to VPN Server.
But caused 'ipv4_get_route()' cannot run correctly in MacOSX, 'split route mode' also broken.

mrbaseman:
We have just committed #99 together with a few other improvements around this topic with https://github.com/adrienverge/openfortivpn/commit/d94915c02e18989021ce8aada5db78d98056192c 

In 1.3.0 we did not catch the case in which your VPN gateway pushes a route in which its own public IP is contained. We try to handle this now, but as @Mabin-J explains, the `ipv4_get_route()` needed for this has never been implemented for MacOSX. The warning you see just tells you that the code could not find out how your current default route is configured. We are working on a solution for this. 

Before https://github.com/adrienverge/openfortivpn/commit/d94915c02e18989021ce8aada5db78d98056192c not only a warning was issued but openfortivpn also refused to continue establishing the connection. With this commit at least 'split route mode' should be working again, although the above described protection for the route to the VPN gateway does not yet work on MacOSX. You can, however, manually add such a host route before starting openfortivpn as a workaround if needed.

[...]

PR #121, opened 2017-05-10
README.md: modify 'macOS' part in 'Installing' Section

Mabin-J:
'brew' usage is changed.

When I ran 'brew install automake autoconf homebrew/versions/openssl101', brew showed this messages.
```
$ brew install automake autoconf homebrew/versions/openssl101
Updating Homebrew...
Error: No available formula with the name "homebrew/versions/openssl101" 
==> Searching for a previously deleted formula...
homebrew/versions/openssl101 was deleted from homebrew/versions in commit 4d7efbf:
  openssl101: remove "unacceptable" old version (https://github.com/Homebrew/homebrew-versions/issues/1528)

To show the formula before removal run:
  git -C "$(brew --repo homebrew/versions)" show 4d7efbf^:openssl101.rb

If you still use this formula consider creating your own tap:
  http://docs.brew.sh/How-to-Create-and-Maintain-a-Tap.html
```

I found that ['homebrew/versions'](https://github.com/Homebrew/homebrew-versions) is deprecated so I modified to possible to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,10 +86,10 @@ For other distros, you'll need to build and install from source:
 
   On macOS :
   ```
-  brew install automake autoconf homebrew/versions/openssl101
+  brew install automake autoconf openssl@1.0
   aclocal && autoconf && automake --add-missing
-  export CPPFLAGS="-I/usr/local/opt/openssl101/include"
-  export LDFLAGS="-L/usr/local/opt/openssl101/lib"
+  export CPPFLAGS="-I/usr/local/opt/openssl/include"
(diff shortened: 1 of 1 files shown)


mrbaseman:
Can someone confirm this? Just a few hours ago @SimonKagstrom has confirmed in #119 that the current master works fine for him. Do we have to consider different versions of MacOSX here?

Mabin-J:
@mrbaseman, I use 'Sierra' but I think that HomeBrew is same in any version of MacOSX.

I think that maybe HomeBrew is updated lately. because when I tried to install another package, HomeBrew was auto-updated.

Issue #104, opened 2017-02-27
otp commandline parameter


rexcze:
Hello,

-o or --opt= switch does nothing.  Output follows:

`openfortivpn <gateway-ip> -u user -p password --opt=123456`
`INFO:   Connected to gateway.`
`2factor authentication token:`

We are using version 1.3.0

mrbaseman:
There are different kinds of 2FA (See #79). Maybe the -o option doesn't work for all of them yet.

rexcze:
This issue appears with fortitoken.

Dne 27. 2. 2017 15:03 napsal uživatel "Martin Hecht" <
notifications@github.com>:

digmore:
I too was confused by the OTP handling. I looked into it and it turns out the "-o" switch was not implemented in the code. I've submitted a PR to fix this.

I notice also that your example says "--opt". If you try changing the spelling to "--otp" you might find it works. I spent a while trying to get the short switch to work, but once I switched to the longer one it worked first time.

mrbaseman:
I have just found this still open. The -o option is recognized in the meantime, and --otp should have worked before if spelled correctly. Can we close this issue?

BlueCase:
I have exactly the same issue as rexcze
```
openfortivpn -vvv --otp=111111
DEBUG:  Loaded config file "/etc/openfortivpn/config".
VPN account password: 
DEBUG:  Config host = "vpn.abc.de"
DEBUG:  Config realm = ""
DEBUG:  Config port = "443"
DEBUG:  Config username = "lalalala"
DEBUG:  Config password = "********"
DEBUG:  One-time password = "111111"
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Connected to gateway.
2factor authentication token: 
```

openfortivpn version: 1.5.0 (same with 1.4.0)
Token: FortiToken 200 HW OTP

Have someone a idea what could be the error?

DimitriPapadopoulos:
@BlueCase The original cause of this issue was double:
- `--otp` not implemented but it should be now ()
- use of `--opt` instead of `--otp`

This looks different. The one-time password is taken into account:
`DEBUG:  One-time password = "111111"
`

It would be better to create a new ticket wouldn't it?

BlueCase:
@DimitriPapadopoulos created #170

Issue #101, opened 2017-02-10
add a way to use openresolv


CircleCode:
when openresolv is available, please add a way to use it so that openfortivpn does not overwrite resolv.conf.

adrienverge:
Good idea, such a contribution would be welcomed!

CircleCode:
sadly, I have no skills in C.
Additionally, after thinking a bit, I think it would be much more open to use some scripts like openvpn.
These scripts would receive the routes and the list of dns servers, and could do whatever they want with it (for example, as i'm using dnsmasq, and I know the dns given by the tunel are only useful for some given domains, I would only register them for concerned domains)

CircleCode:
I made some tests about it, and it appears that the use of `--no-dns` and a `/etc/ppp/ip-up.d/`n script can almost do this. But I am a bit limited, and I think we could allow more flexibility by simply using the `ipparam` parameter: this param is passed to `/etc/ppp/ip-up.d/` and `/etc/ppp/ip-down.d/` scripts and can allow a more grained control of what to do for each vpn.
For me, an additional option to openforticlient (like `pppd-ipparam`) can be directly forwarded to the pppd command, and the user can use it in the way he wants in his own scripts.

mrbaseman:
I have implemented the ip-param option. @CircleCode could you test https://github.com/mrbaseman/openfortivpn/commit/d847932b0461e2324ad3f25e9c1c4213ed8179a1 please?

mrbaseman:
maybe we could include the ip-up.d/ and ip-down.d/ scripts at least as a sample or provide the recipe in the man page

CircleCode:
@mrbaseman thanks for this code.
Unfortunately, I'm on vacation right now, I'm not sure I will be able to test this before the next 2 weeks.
In any case, I'll comment here.
Thanks again.

mrbaseman:
@digmore has confirmed in #124 that the ipparam setting from my branch was working, so I have just merged this into master.
@CircleCode, first of all enjoy your vacation! When you are back, please let me know if the implementation of the ipparam setting fulfills your needs, too, or if you need some more flexibility. Right now, the parameter is just handed over unchanged to pppd by openfortivpn.

mrbaseman:
Just a remark: pppd sets a couple of environment variables (see man pppd). ipparam is currently passed through openfortivpn, but inside the ip-up / ip-down scripts one can replace variables in this string as follows:
`ipparam_raw=$6`
[...]

PR #124, opened 2017-05-18
Small improvements

digmore:
Hi, 

Great tool... thanks for sharing it.

This PR offers some small improvements I made in response to issues I had getting up and running.

Firstly, I was also confused by the OTP handling as in adrienverge/openfortivpn#104. It turns out the documentation mentions both "-o" and "--otp" switches, but only the latter was implemented in the code. I have added handling for the short option here. This may possibly close issue #104 as one option literally didn't exist and there is a spelling mistake in the second option in his example.

Secondly, I was having DNS issues due to the search option of resolv.conf not being supported. This information doesn't come back from my VPN end-point when the tunnel is established so it seems right that the client doesn't set it. As a workaround I added support to the configuration file for defining custom scripts to call on tunnel up and interface down events. For my case I manipulate this search parameter from a shell script called by openfortivpn when these events occur. I'm sure there would be other use cases; It seems to be a feature present in other VPN clients.

Please consider these for inclusion. C is not my most comfortable language so hopefully it is acceptable. It has tested okay over a few days on my machine.

Thanks

### DIFF
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -151,3 +151,9 @@ pppd-use-peerdns = 1
 insecure-ssl = 0
 .br
 cipher-list = HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4
+.br
+# These specify a script to run on VPN up/down events
+.br
+up-script = /usr/local/bin/custom_up.sh
+.br
+down-script = /usr/local/bin/custom_down.sh
--- a/src/config.c
+++ b/src/config.c
@@ -229,6 +229,10 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			cfg->insecure_ssl = insecure_ssl;
 		} else if (strcmp(key, "cipher-list") == 0) {
 			cfg->cipher_list = strdup(val);
+		} else if (strcmp(key, "up-script") == 0) {
+			cfg->up_script = strdup(val);
+		} else if (strcmp(key, "down-script") == 0) {
+			cfg->down_script = strdup(val);
 		} else {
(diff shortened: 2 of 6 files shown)


mrbaseman:
thanks for the hint about the -o option. Together with a few other minor fixes I have just landed this on master. 
About the main code change here: The up- and down-scripts: There is already an approach for resolvconf.d which makes use of the ip-up and ip-down scripts of pppd in /etc/ppp/ (see #101). 
Yesterday I have implemented the support for ipparam of pppd which could be used for the search parameter in resolv.conf, too.
I understand that your solution is more transparent to first time users, but I would like to keep openfortivpn a relatively thin client and not put too much features into it, especially when the same thing can be achieved with other means. 
Maybe we were just by coincidence working on the same problem and arrived at two different solutions. But if your approach has clear advantages, please explain and I would be happy to accept your pull request.
One more thing: @adrienverge has a strong interest in keeping the commits to the repository clean. The commits should be atomic and revertable if needed. This pull request is about two separate problems (-o option and the scripts), which makes it harder to squash all the commits (including the rebase and coding style fixes) into one single transaction. Now that I have already put the fix for the -o option into master, rebasing the branch should solve that problem, but I just wanted to mention this here.

digmore:
@mrbaseman, I appreciate for the tip about improving PRs.

I hadn't considered the interplay with pppd. I should have checked the other issues first. I managed to get this working using the ip-up.d/ip-down.d script approach with your ipparam branch. It's fine for me needs.

I'm going to close this PR as it's unnecessary. Thanks for the feedback..

mrbaseman:
thanks for confirming that the ipparam is working. So I have just merged my branch into master.

PR #126, opened 2017-05-23
Add openSUSE dependencies to README

nbirnel:
Nothing but adding another install dependencies for another distro to the README.


mrbaseman:
thanks, applied

PR #127, opened 2017-05-24
modify README.md (modify dependencies, etc)

Mabin-J:
- fix indent
- modify dependencies
- add some decoration

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,40 +1,39 @@
 openfortivpn
 ============
 
-openfortivpn is a client for PPP+SSL VPN tunnel services. It spawns a pppd
-process and operates the communication between the gateway and this process.
+openfortivpn is a client for PPP+SSL VPN tunnel services.  
+It spawns a pppd process and operates the communication between the gateway and 
+this process.
 
 It is compatible with Fortinet VPNs.
 
+
+
+--------
 Examples
 --------
 
 * Simply connect to a VPN:
-
   ```
   openfortivpn vpn-gateway:8443 --username=foo
   ```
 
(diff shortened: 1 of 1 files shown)


mrbaseman:
could you squash these commits, please, so that we have a single transaction in this PR?

Mabin-J:
OK. I'll squash that.

I think that two commit's checkpoint is different so I didn't squash.
(but 4bb5062 isn't same rule... that my fault. sorry.)

mrbaseman:
If you want #128 in here, could you please update the configure command line and the example sudo configuration and squash once more?

Mabin-J:
Oh, it's my mistake. I added description about #128 in commit message but I didn't add file...  :sweat_smile:

mrbaseman:
thanks, applied. I personally would put the configuration into /usr/local/etc then, but that would be the default for sysconfdir and theconfigure option could be dropped. So I decided to keep the configuration in /etc so that we have this option, which many users might find useful, in the examples.

Mabin-J:
I saw that put config file into /usr/local/etc and make symbolic link into /etc. 
But I don't know which way is common.

mrbaseman:
I'd say the default is that the sysconfig directory should be $prefix/etc. However, in some applications it is hard-coded as /etc, while the install script puts the configuration into $prefix/etc. In that case, a symlink in /etc pointing to the real location of the configuration would be a workaround. However, the better solution would be to set the location at configure time for both, the install script and for the program (btw, for openfortivpn #76 was about this topic)

PR #132, opened 2017-05-29
fix lock status when fail to spawn pppd in macOS.

Mabin-J:
When I run openfortivpn without elevate privileges in Linux,
it is finished normally but not in macOS.

It showed `INFO:   Cancelling threads...` and locked.
It didn't response when I push `Ctrl + C`

This is full message.
```
$ ./openfortivpn -c connect_info.config -v
DEBUG:  Loaded config file "connect_info.config".
VPN account password: 
DEBUG:  Config host = "123.123.123.123"
DEBUG:  Config realm = ""
DEBUG:  Config port = "443"
DEBUG:  Config username = "username"
DEBUG:  Config password = "********"
WARN:   This process was not spawned with root privileges, this will probably not work.
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Connected to gateway.
INFO:   Authenticated.
DEBUG:  Cookie: SVPNCOOKIE=[COOKIE]
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
WARN:   No gateway address, using interface for routing
DEBUG:  pppd_read_thread
DEBUG:  ssl_read_thread
DEBUG:  ssl_write_thread
DEBUG:  if_config thread
DEBUG:  gateway ---> pppd (12 bytes)
DEBUG:  pppd_write thread
DEBUG:  pppd ---> gateway (12 bytes)
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0

(repeat about 940 times...)

WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
DEBUG:  gateway ---> pppd (12 bytes)
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
WARN:   read returned 0
ERROR:  write: Input/output error
WARN:   read returned 0
INFO:   Cancelling threads...
WARN:   read returned 0

(Not response)

^C^C 

(Not response)

(I just ran `kill pid` in another terminal)

[1]    2201 terminated  ./openfortivpn -c connect_info.config -v
```

I checked something, and I found that `pppd` zombie process...
```
$ ps aux |grep ppp
username          2286   0.0  0.0        0      0   ??  Z     2:22PM   0:00.00 (pppd)
username          2288   0.0  0.0  2432804    780 s005  S+    2:22PM   0:00.00 grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn ppp
```
The 2286 process was remained until openfortivpn was killed.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -295,6 +295,9 @@ static void *pppd_write(void *arg)
 			          len - written);
 			if (n == -1) {
 				log_error("write: %s\n", strerror(errno));
+#ifdef __APPLE__
+				sem_post(&sem_if_config);
+#endif
 				goto err_free_buf;
 			}
 			written += n;


mrbaseman:
I have tried to reproduce this on linux, but I wasn't successful. Looking at the source I assume that the problem is a consequence of commenting out the lines https://github.com/Mabin-J/openfortivpn/blob/fix-lock-status-in-macos/src/io.c#L576 ff. for signal handling on OSX. Maybe we can find a better solution here than curing the symptoms? Any ideas welcome

DimitriPapadopoulos:
The source code has been introduced last year by commit 5cb432b _Mac OS X port_. Any clue why the signal handling code is disabled on MacOS X?

mrbaseman:
I think the comment in the source explains the intention for commenting it out on osx:
`// on osx this prevents the program from being stopped with ctrl-c`
the question is if there is a better solution than that to make handling of ctrl-c work on osx

Mabin-J:
I think that openfortivpn should be stopped when user push `Ctrl + C`.
And It is already works in Linux, macOS, both.

but openfortivpn fail to spawn `pppd`, it is finished normally in Linux, but it isn't finished in macOS.
`Ctrl + C` cannot work when that situation...

Mabin-J:
@DimitriPapadopoulos, I think that zombie process of `pppd` is clue.
because in Linux, openfortivpn didn't make zombie process same situation.

Mabin-J:
@mrbaseman, I don't know why this PR is failed in Linux...
That code is ran only in macOS...

Mabin-J:
I updated content.

mrbaseman:
@Mabin-J Not the pull request failed on linux. I tried to reproduce the **problem**, which the pull request aims to fix, because I was wondering, if (and why) the `#ifdef __APPLE__` is appropriate. 
My reasoning: is this a problem specific for osx, or is it a more general problem and you just happen to see this on osx?

While investigating that, I have noticed that parts of the signal handling is just commented out on osx. 

Therefore, my question: If my assumption is correct, that the problem appears because part of the signal handling is commented out, wouldn't it be better not to disable the code which does signal handling on linux, and maybe add some extra code instead to re-enable `Ctrl+C` on osx?

Mabin-J:
I tested after I removed [io.c#574](https://github.com/adrienverge/openfortivpn/blob/5cb432beca51f777ff97b036d9352d46e08e488b/src/io.c#L574) and [io.c#580](https://github.com/adrienverge/openfortivpn/blob/5cb432beca51f777ff97b036d9352d46e08e488b/src/io.c#L580) right now.

When I removed two line, openfortivpn wasn't break when I pushed `Ctrl + C` when it normal situation. (elevate privileges and normally spawned `pppd`)

And symptom is same when failed to elevate privileges and failed to spawn `pppd`.

I think that comment is wrong.

Mabin-J:
But you are right. 
I think that this symptom is caused zombie process 
so if zombie process problem is solved, maybe lock problem be solved.

Mabin-J:
But.. when lock situation, `pppd` cannot be killed with `kill -9 2286`... :(

mrbaseman:
so, to summarize:
- the problem does not occur on linux
- the problem is not caused by the fact that some parts of signal handling are skipped on osx ([io.c#L574](https://github.com/adrienverge/openfortivpn/blob/5cb432beca51f777ff97b036d9352d46e08e488b/src/io.c#L574) ff.), so my assumption above was wrong.
- the problem seems to be connected with the pppd zombie process, but we haven't found the root cause for this happening
so, the proposed fix (including the surrounding `#ifdef __APPLE__`) is an appropriate solution until the cause of the zombie process is found. Let's keep this in mind, and I'll merge this pull request.
Thanks for the contribution, testing, and the discussion

Issue #105, opened 2017-02-27
GCD semaphores might be unsafe when used in signal handlers


henninglive:
Openfortivpn [currently uses](https://github.com/adrienverge/openfortivpn/blob/master/src/io.c#L33-L45) GCD (Grand Central Dispatch) semaphores on OS-X. This is done to work around the fact the that OS-X doesn’t support [unnamed posix semaphores](http://man7.org/linux/man-pages/man7/sem_overview.7.html). I ran into the same problem while working on another project and while looking for workarounds I found multiple solutions suggesting substituting posix semaphores with GCD semaphores, but I was not able to find any documentation specifying if GCD semaphores are async-signal safe.

I did some further dinging and found the [source](https://github.com/apple/swift-corelibs-libdispatch) for GCD. As you can see [here](https://github.com/apple/swift-corelibs-libdispatch/blob/d04a0dff9f0ecf327d9fc7626f3013be4d9353f5/src/semaphore.c#L239-L251) and [here](https://github.com/apple/swift-corelibs-libdispatch/blob/d04a0dff9f0ecf327d9fc7626f3013be4d9353f5/src/semaphore.c#L220-L237), GCD uses `mach/semaphore.h` internally. Unfortunately, the [documentation](https://developer.apple.com/library/content/documentation/Darwin/Conceptual/KernelProgramming/synchronization/synchronization.html) for `mach/semaphore.h` suggest that they are not async-signal safe.
>Semaphores can be used any place where mutexes can occur. This precludes their use in interrupt handlers or within the context of the scheduler, and makes it strongly discouraged in the VM system. 

If this is correct, it is properly a bad idea to use GCD semaphores.  The currently solution might work most of the time, but specific timing might cause it to fail. [Here](https://docs.oracle.com/cd/E19455-01/806-5257/gen-26/index.html) is more info on Async-Signal Safety.

[Here](https://bugs.openjdk.java.net/browse/JDK-8129814) is another discussion about the same problem in OpenJDK. They suggest using [BSD kqueue/kevent](https://www.freebsd.org/cgi/man.cgi?kqueue).

In the other project I was working on we ended up using pipes, similar to [this](http://ck-hack.blogspot.no/2013/09/unnamed-semaphores-and-pososx.html). This should work on all posix compliant systems, unlike unnamed semaphores which are an optional part of the posix standard.

DimitriPapadopoulos:
Indeed _openfortivpn_ uses GCD semaphores:
```
[...]

PR #135, opened 2017-05-30
fix core cause of openfortivpn is locked when spawning `ppp` is failed.

Mabin-J:
This commit is related with #132 and 158f4894ed5164991168a66bba874a9a268d4bbb .

It seems that `pthread_cancel` cannot effect to `dispatch_semaphore_wait`.
(`dispatch_semaphore_wait` is used for `sem_wait` for macOS, [io.c#L41](https://github.com/adrienverge/openfortivpn/blob/158f4894ed5164991168a66bba874a9a268d4bbb/src/io.c#L41))
It seems that `dispatch_semaphore_wait` not use real semaphore.

If `pthread_cancel cannot` effect to `dispatch_semaphore_wait`, 
`if_config` should be returned at `sem_wait(&sem_if_config);`([io.c#L509](https://github.com/adrienverge/openfortivpn/blob/158f4894ed5164991168a66bba874a9a268d4bbb/src/io.c#L509)) but it isn't...
`openfortivpn` is locked that line.

#132 is not best solution because when run `sem_post(&sem_if_config);`([io.c#L299](https://github.com/adrienverge/openfortivpn/blob/158f4894ed5164991168a66bba874a9a268d4bbb/src/io.c#L299)), io.c#L510-L534 will be ran.

In Linux, when run `pthread_cancel(if_config_thread);`([io.c#L609](https://github.com/adrienverge/openfortivpn/blob/158f4894ed5164991168a66bba874a9a268d4bbb/src/io.c#L609)), 
`sem_wait(&sem_if_config);`([io.c#L509](https://github.com/adrienverge/openfortivpn/blob/158f4894ed5164991168a66bba874a9a268d4bbb/src/io.c#L509)) will be broken and `if_config` function will be returned immediately.

So I add code for breaking after `sem_wait` when `sem_post(&sem_stop_io);` is called.

I think that remain this logic before find alternative `dispatch_semaphore`.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -30,18 +30,31 @@
 #include <netinet/tcp.h>
 #include <pthread.h>
 #include <signal.h>
+
 #ifdef __APPLE__
+
 /* Mac OS X defines sem_init but actually does not implement them */
-#include <dispatch/dispatch.h>
+#include <mach/mach.h>
 
-typedef dispatch_semaphore_t	sem_t;
+typedef semaphore_t os_semaphore_t;
 
-#define sem_init(psem,x,val)	*psem = dispatch_semaphore_create(val)
-#define sem_post(psem)		dispatch_semaphore_signal(*psem)
-#define sem_wait(psem)		dispatch_semaphore_wait(*psem, \
-					DISPATCH_TIME_FOREVER)
+#define SEM_INIT(sem, x, value)	semaphore_create(mach_task_self(), sem, \
+								SYNC_POLICY_FIFO, value)
+#define SEM_WAIT(sem)			semaphore_wait(*sem)
+#define SEM_POST(sem)			semaphore_signal(*sem)
+#define SEM_DESTROY(sem)		semaphore_destroy(mach_task_self(), *sem)
 
-#define sem_destroy(psem)	dispatch_release(*psem)
+#else
+
+#include <semaphore.h>
+
+typedef sem_t os_semaphore_t;
+
+#define SEM_INIT(sem, x, value)	sem_init(sem, x, value)
+#define SEM_WAIT(sem)			sem_wait(sem)
+#define SEM_POST(sem)			sem_post(sem)
+#define SEM_DESTROY(sem)		sem_destroy(sem)
+
 #endif
 
 #include "hdlc.h"
@@ -116,7 +129,7 @@ static struct ppp_packet *pool_pop(struct ppp_packet_pool *pool)
 {
 	struct ppp_packet *first = pool->list_head;
 
-	//sem_wait(&pool->sem);
+	//SEM_WAIT(&pool->sem);
 	pthread_mutex_lock(&pool->mutex);
 	while (pool->list_head == NULL)
 		pthread_cond_wait(&pool->new_data, &pool->mutex);
@@ -130,9 +143,9 @@ static struct ppp_packet *pool_pop(struct ppp_packet_pool *pool)
 	return first;
 }
 
-static sem_t sem_pppd_ready;
-static sem_t sem_if_config;
-static sem_t sem_stop_io;
+static os_semaphore_t sem_pppd_ready;
+static os_semaphore_t sem_if_config;
+static os_semaphore_t sem_stop_io;
 
 /*
  * Thread to read bytes from the pppd pty, convert them to ppp packets and add
@@ -174,7 +187,7 @@ static void *pppd_read(void *arg)
 			continue;
 		} else if (first_time) {
 			// pppd did talk, now we can write to it if we want
-			sem_post(&sem_pppd_ready);
+			SEM_POST(&sem_pppd_ready);
 			first_time = 0;
 		}
 		off_w += n;
@@ -235,7 +248,7 @@ static void *pppd_read(void *arg)
 
 exit:
 	// Send message to main thread to stop other threads
-	sem_post(&sem_stop_io);
+	SEM_POST(&sem_stop_io);
 	return NULL;
 }
 
@@ -252,7 +265,7 @@ static void *pppd_write(void *arg)
 	FD_SET(tunnel->pppd_pty, &write_fd);
 
 	// Write for pppd to talk first, otherwise unpredictable
-	sem_wait(&sem_pppd_ready);
+	SEM_WAIT(&sem_pppd_ready);
 
 	log_debug("pppd_write thread\n");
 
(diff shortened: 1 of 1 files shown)


Mabin-J:
This PR's purpose is same with 158f4894ed5164991168a66bba874a9a268d4bbb so if possible, I want to merge 158f4894ed5164991168a66bba874a9a268d4bbb.

adrienverge:
Hi @Mabin-J,

Thanks for your help on making openfortivpn MacOS-compatible, it is appreciated!

However I don't understand the goal of your latest pull-requests. Could you open a GitHub issue to clearly state:
1) what are the current problems on MacOS,
2) your vision on how to fix them (for example by describing a set of pull-requests to merge).

This way @mrbaseman, @DimitriPapadopoulos and I can follow your contributions efficiently and merge them while avoiding regressions and bugs.

Also, please make sure your commits are clear (e.g. not `io.c: change code to solve #132`), documented if possible, atomic and revertable. This way in case of a bug, we can easily find the cause and a nice solution.

Thanks!

Mabin-J:
Hi @adrienverge. 

I'll do it.
Thanks for guide.

mrbaseman:
Hello @Mabin-J,

I would like to understand what exactly this patch does, and I believe I did, but please correct me if I got something wrong. If I understand your code correctly, you do 2 things:
- you revert #132 because it did not work correctly. It allowed to kill openfortivpn with `Ctrl + C` but pppd still went in zombie state, right?
- now you have found the following solution to avoid this zombie state: If I understand it right `pthread_cancel` does not cancel the threads on MacOS X and you work around this by setting `is_destroy` and calling `sem_post` which is defined as `dispatch_semaphore_signal` in the Apple code. This wakes up the hanging threads and the functions which were stuck in `sem_wait` check `is_destroy` and immediately exit. Did I get that right?

Apart from reassuring myself that I understood the change correctly, I would suggest to rename `is_destroy` to some a bit more specific name like `sem_cancel_pthreads` (or do you have a better idea?), and there is an unnecessary empty line added ([619](https://github.com/adrienverge/openfortivpn/pull/135/files#diff-592ee3fd43260e8d9bd24192211bce59R619)) which I would like to be removed before merge.

Thanks for your efforts in improving the MacOS X code in openfortivpn.

Mabin-J:
I checked [your comment](https://github.com/adrienverge/openfortivpn/issues/137#issuecomment-305261463) and I tested `semaphore_*` and it was succeed.

But I didn't test in Linux.
I'll test in Linux tomorrow.

Mabin-J:
Before #132, `openfortivpn` wasn't finished when I pushed `Ctrl + C`. it was only finished when I ran `kill pid`.
And after I ran `kill pid`, zombie process disappeared.

I think that Zombie process caused locked thread(if_config_thread).
When `if_config_tread` was finished correctly, zombie process didn't appear.

Mabin-J:
It works also in Linux correctly. :)

mrbaseman:
ok, this is a different approach now.

- Does it address #105? could you give a reference, please? In the commit message you have added a link to OpenJDK source, but in #105 in the context of JDK BSD kqueue / kevent are mentioned. I can't see the connection to your commit. Maybe I missed it?

- In the latest version you have introduced capital letter versions of the macros and you had to change the linux code as well. Is this really needed? Wouldn't the following work in the MacOS X code branch?
`#define sem_init(sem, x, value)	semaphore_create(mach_task_self(), sem, \`
`								SYNC_POLICY_FIFO, value)`
`#define sem_wait(sem)			semaphore_wait(*sem)`
`#define sem_post(sem)			semaphore_signal(*sem)`
`#define sem_destroy(sem)		semaphore_destroy(mach_task_self(), *sem)`
Maybe you tried that already but if this is not possible I would like to understand why.

Mabin-J:
* I think that using kqueue/kevent is very difficult because we need to make sem_* function with kqueue/kevent function. so I used `semaphore_*` function like OpenJDK.

* When I used `#define sem_*(sem) semaphore_*(*sem)`, It didn't works. I think that because `sem_*` function is already exist in macOS and those didn't work. it means, `sem_*` cannot be overrided. I think that maybe this reason, OpenJDK also used `SEM_*` define.

mrbaseman:
Ah, now I see, on [L1940](http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/8b16790cd73a/src/os/bsd/vm/os_bsd.cpp#l1940) and [L1905](http://hg.openjdk.java.net/jdk9/jdk9/hotspot/file/1ae9e84f68b3/src/os/bsd/vm/os_bsd.cpp#l1905) ff. of the links you have posted. They do it the same way as you have proposed for openfortivpn

mrbaseman:
I have just landed #134 on master and resolved the resulting conflict in io.c with https://github.com/adrienverge/openfortivpn/pull/135/commits/903d091ab1e93be9bdb7ca8a4214d73b049e4bc1 and https://github.com/adrienverge/openfortivpn/pull/135/commits/45f554535bd88427994cb6dbfe1bc78e66ac2090
@Mabin-J maybe you can squash these commits into your original proposal.

Mabin-J:
@mrbaseman, I finished.

mrbaseman:
@Mabin-J thanks. There is still discussion in #105. I'd still wait a bit for comments there.

mrbaseman:
Dear @Mabin-J 

in #105 we have come to the conclusion that this PR is a good step forward, not only to fix the locking issue here, but also as a solution for #105. I would kindly ask you to update the commit message once more, so that it also includes this conclusion. I'd suggest the following:

> io.c: fix core cause of openfortivpn is locked when spawning `pppd` has failed.
>
> This commit reverts the temporary fix from #132 (commit 158f489) and instead replaces GCD semaphores with Mach semaphores, following the lines of  http://hg.openjdk.java.net/jdk8u/jdk8u/hotspot/file/8b16790cd73a/src/os/bsd/vm/os_bsd.cpp.
>
> By doing so, this commit is also a solution for issue #105 (GCD semaphores might be unsafe when used in signal handlers): On MacOS X named POSIX semaphores, which are known to be async-signal-safe, are implemented upon Mach semaphores (https://books.google.de/books?id=K8vUkpOXhN4C&pg=PA1219&lpg=PA1210&focus=viewport&dq=semaphore_create&hl=de). An analysis of the use of semaphores in openfortivpn has lead to the conclusion that Mach semaphores are just as good as unnamed POSIX semaphores which unfortunately are not implemented in MacOS X.

Issue #87, opened 2017-01-05
Max number of split route


Valantin:
Hi in my environment I've 96 different route to add to administrative client
the max number of route in openfortivpn is "#define MAX_SPLIT_ROUTES 32"

Mabin-J:
Wow...
I modified 32 -> 64 in issue #110 because I needed 32-64 split route.
But your case, `MAX_SPLIT_ROUTES` is need to increase.
@mrbaseman, What do you think?

mrbaseman:
let's finish the signal topic first, then if automatic merge fails rebase the above pull request (the changes are in different files, so it should work without rebasing, but let's see...)

mrbaseman:
I have just landed #136 on the current master which solves this issue

PR #136, opened 2017-05-30
ipv4.h: increase 'MAX_SPLIT_ROUTES' 64 to 128 (Issue #87)

Mabin-J:


mrbaseman:
@Mabin-J could you rebase this branch on the current master, please, so that we have a cleaner history in the git repository after merging?

DimitriPapadopoulos:
@mrbaseman I think you can rebase while merging, at least there seems to be an option for that, see attached image.
Rebase and merge

Mabin-J:
@DimitriPapadopoulos, I think that @mrbaseman already know.
but I think that `Rebase and merge` makes one more unnecessary commit message.

Mabin-J:
I have a question.

Is 128 enough?
I think that some environment needs more routes. 
So, I think that to set more 128 or to check logic about this are better.

adrienverge:
Using the "rebase and merge" button is a good idea, as it keeps the git history cleaner and more readable.

That being said, I agree with @mrbaseman: rebasing the branch now makes sure there are no problems induced by the combination of latest changes on `master` and this PR.

DimitriPapadopoulos:
Perhaps even 128 won't be enough in some edge cases. It's probably best to have the code adapt to the size of the routing table instead of hardcoding it, but that's more involved. I suggest:
1. changing to 128 as a short-term solution,
2. changing the logic of the code to adapt to the size of routing tables.

Unless you already have the code for the second option - or can write it rapidly.

mrbaseman:
- I think 128 should be enough for most cases. The problem is that this number is used as a size of a static array, so it must be a constant in C. It makes sense to change the code in order to switch to a dynamically allocated array, but that's a bit more work, as it involves a mechanism to re-allocate more memory when needed and copying existing data structures to the new (larger) array.
- about rebasing: I have merged the current master into the pull request and noticed an extra commit. There were no conflicts, but just another commit for making sure that the change of a single number can land on the head of the git... I wasn't aware that I can do the "rebase" during the merge operation. Thanks for that hint. Anyhow, while I was in the office next door, you guys have been busy here, so that all that is left for me to do is to accept the commit :+1:

xarli64:
Hello, just to comment on the value, my company is currently announcing 125 routes (rfc1918 + a substantial number of public internal IPv4 routes). Fiting in a 128 entries table, but nearly full.

PR #144, opened 2017-06-12
Configure openssl

### LINKED ISSUE #143: Add openssl location to configure options
I've created a formula for homebrew, and the homebrew maintainer has asked if it is possible to add the openssl location to `configure` options instead of using the environment to specify it.

https://github.com/Homebrew/homebrew-core/pull/14344

Please consider adding it, together with an autogen.sh script to automate the generation of the autoconf/automake files

mnencia:
Closes #143

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -ex
+set -exu
 
 aclocal
 autoconf
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+set -ex
+
+aclocal
+autoconf
+automake --add-missing
--- a/configure.ac
+++ b/configure.ac
@@ -12,6 +12,99 @@ AC_GNU_SOURCE
 
 AM_SILENT_RULES([yes])
 
+AC_DEFUN([OPENSSL_DEVEL],
+[
+	dnl
+	dnl Set up configure script macros
+	dnl
+	AC_ARG_WITH(openssl,
+		[  --with-openssl=<path>     path containing OpenSSL header and library subdirs],
+		[OPENSSL_lib_check="$with_openssl/lib64 $with_openssl/lib $with_openssl/lib64/openssl $with_openssl/lib/openssl"
+		  OPENSSL_inc_check="$with_openssl/include $with_openssl/include/openssl"],
+		[OPENSSL_lib_check="/usr/local/openssl/lib64 /usr/local/openssl/lib /usr/local/lib64/openssl /usr/local/lib/openssl /opt/openssl/lib64 /opt/openssl/lib /usr/local/opt/openssl/lib64 /usr/local/opt/openssl/lib /usr/lib64/openssl /usr/lib/openssl /usr/local/lib64 /usr/local/lib /usr/lib64 /usr/lib"
+		  OPENSSL_inc_check="/usr/local/openssl/include /usr/local/include/openssl /opt/openssl/include /usr/local/opt/openssl/include /usr/local/include/openssl /usr/local/include /usr/include/openssl /usr/include"])
+	AC_ARG_WITH(openssl-lib,
+		[  --with-openssl-lib=<path> directory path of OpenSSL library],
+		[OPENSSL_lib_check="$with_openssl_lib $with_openssl_lib/lib64 $with_openssl_lib/lib $with_openssl_lib/lib64/ssl $with_openssl_lib/lib/ssl"])
+	AC_ARG_WITH(openssl-include,
+		[  --with-openssl-include=<path> directory path of OpenSSL headers],
+		[OPENSSL_inc_check="$with_openssl_include $with_openssl_include/include $with_openssl_include/include/ssl"])
+
+	dnl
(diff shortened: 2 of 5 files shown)


autogen.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -ex
```
adrienverge: Can you add the `u` option here? (`set -eux`)

DimitriPapadopoulos:
[Travis CI configuration file](https://travis-ci.org/adrienverge/openfortivpn/jobs/241958559/config) `.travis.yml` should be updated as well:
```
script:
  - ./tests/lint/run.sh
  - >
      aclocal && autoconf && automake --add-missing
      && ./configure --prefix=/usr --sysconfdir=/etc
      && make
```

autogen.sh
mnencia: It should. Just point it to the installation prefix

mnencia: I've added it together with the travis CI update

mnencia:
The --with-openssl option should work on also on Linux. Just point it to the openssl installation prefix.

mnencia:
Unfortunately this doesn't work out of the box. I'll debug it asap.

mnencia:
It fails to detect the location of openssl libraries

```
root@precise:/tmp/openfortivpn# dpkg -S ssl.so
libssl1.0.0: /lib/x86_64-linux-gnu/libssl.so.1.0.0
```

mrbaseman:
I have noticed that openfortivpn indeed does not link against openssl 1.1.0. I'm still looking for a way to check the openssl version in the configure script. 
I have found another source which might be helpful : the [autoconf-archive](https://www.gnu.org/software/autoconf-archive/ax_check_openssl.html)

mnencia:
I've added the OpenSSL version check to the configure

DimitriPapadopoulos:
@mrbaseman Just wondering, what does this mean exactly?
> I have noticed that openfortivpn indeed does not link against openssl 1.1.0.

Does this mean that _openfortivpn_ should be compatible with _openssl_ 1.1.0 but currently isn't? If so should _openfortivpn_ be fixed? And if so in the source code or in the build infrastructure only?

mrbaseman:
@DimitriPapadopoulos on the long term we should support openssl 1.1.x, but for the current issue I just would check if the user who supplies a path to a non-standard openssl installation tries to use a supported one. Most distributions ship patched versions of 1.0.0, 1.0.1 or 1.0.2, but this will also change when upstream support for these ends. A few enterprise versions probably will still backport patches like they currently do for 0.9.8 but that's probably not the standard case.
In order to support openssl 1.1.0 the code has to be changed. I think the problem is that most structures have been made opaque, but I need a closer look at the compiler errors and compare them with the FAQ or the openssl mailing list archives to tell for sure.

configure.ac
```diff
@@ -10,8 +10,123 @@ AM_INIT_AUTOMAKE([foreign subdir-objects])
...
+	dnl
+	case "$ac_cv_openssl_lib" in
+	  /usr/lib) ;;
+	  *) LDFLAGS="$LDFLAGS -L${ac_cv_openssl_lib}" ;;
+	esac
+	CPPFLAGS="$CPPFLAGS -I${ac_cv_openssl_inc}"
+	AC_MSG_CHECKING([that we can build OpenSSL programs])
+	AC_COMPILE_IFELSE(
+		[AC_LANG_PROGRAM([#include <openssl/ssl.h>],
+			[init_openssl_library()])],
```
mrbaseman: I have noticed that the init function should read SSL_library_init() - and also this check seems not to work. It silently goes passes and just writes out a warning into config.log:
> warning: implicit declaration of function 'init_openssl_library' [-Wimplicit-function-declaration]

mrbaseman:
@mnencia thanks for adding the version check. I hope that I find more time for a closer look next week

configure.ac
mnencia: I may be wrong, but I think that we could completely remove this check, because there already is the library check some lines below:

```
AC_CHECK_LIB([ssl], [SSL_connect], [], [AC_MSG_ERROR([Cannot find libssl.])])
```

mnencia: I decided to keep this test because it is useful to detect and eventual compiler misconfiguration.

mnencia:
I've fixed the openssl library function name that we use in the configure check.

mrbaseman:
@mnencia Great! Thanks a lot! I have just tested building against a few openssl installations and it works nicely. I think we can merge this into master. The homebrew formula however refers to version 1.3.1 which doesn't contain this pull request. I think it makes sense to tag a new release 1.3.2 then. @adrienverge and @DimitriPapadopoulos what do you think?

DimitriPapadopoulos:
There have been quite a few changes for Mac OS X. Perhaps it's time for a new release indeed.

adrienverge:
Sure, we can release a new version once it's merged.

@mnencia Can you squash your last commit into f81c28a?

mnencia:
Last commit squashed into  	3408e41

adrienverge:
Thanks!

v1.4.0 released.

PR #149, opened 2017-07-02
Routes improvements (maybe)

martinetd:
Hi,

I've had most of these commits laying around for a while; some could be a bit messy so feel free to be picky. I'm not 100% happy with the last one in particular myself :)

Comments/rework requests welcome

### DIFF
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -9,7 +9,9 @@ openfortivpn \- Client for PPP+SSL VPN tunnel services
 [\fB\-u\fR \fI<user>\fR]
 [\fB\-p\fR \fI<pass>\fR]
 [\fB\-\-realm=\fI<realm>\fR]
+[\fB\-\-set-routes=<bool>\fR]
 [\fB\-\-no-routes\fR]
+[\fB\-\-set-dns=<bool>\fR]
 [\fB\-\-no-dns\fR]
 [\fB\-\-ca-file=\fI<file>\fR]
 [\fB\-\-user-cert=\fI<file>\fR]
@@ -57,11 +59,15 @@ VPN account password.
 Connect to the specified authentication realm. Defaults to empty, which
 is usually what you want.
 .TP
-\fB\-\-no-routes\fR
-Do not try to configure IP routes through the VPN when tunnel is up.
+\fB\-\-set-routes=\fI<bool>\fR, \fB\-\-no-routes\fR
+Set if we should try to configure IP routes through the VPN when tunnel is up. If used multiple times, the last one takes priority.
+
+\fB\-\-no-routes\fR is the same as \fB\-\-set-routes=\fI0\fR.
 .TP
-\fB\-\-no-dns\fR
-Do not add VPN nameservers in /etc/resolv.conf when tunnel is up.
+\fB\-\-set-dns=\fI<bool>\fR, \fB\-\-no-dns\fR
+Set if we should add VPN nameservers in /etc/resolv.conf when tunnel is up. If used multiple times, the last one takes priority.
+
+\fB\-\-no-dns\fR is the same as \fB\-\-set-dns=\fI0\fR.
 .TP
 \fB\-\-ca-file=\fI<file>\fR
 Use specified PEM-encoded certificate bundle instead of system-wide store to
--- a/src/config.c
+++ b/src/config.c
@@ -56,7 +56,7 @@ int add_trusted_cert(struct vpn_config *cfg, const char *digest)
  * @params[in] str  the string to read from
  * @return          0 or 1 if successful, < 0 if unrecognized value
  */
-static int strtob(const char* str)
+int strtob(const char* str)
 {
 	if (str[0] == '\0') {
 		return 0;
--- a/src/config.h
+++ b/src/config.h
@@ -114,6 +114,7 @@ struct vpn_config {
 	free((cfg)->cipher_list);
 
 int add_trusted_cert(struct vpn_config *cfg, const char *digest);
+int strtob(const char* str);
 
 int load_config(struct vpn_config *cfg, const char *filename);
 
--- a/src/main.c
+++ b/src/main.c
@@ -22,7 +22,7 @@
 #include "tunnel.h"
 #include "userinput.h"
 
-#define USAGE \
+#define usage \
 "Usage: openfortivpn [<host>:<port>] [-u <user>] [-p <pass>]\n" \
 "                    [--realm=<realm>] [--otp=<otp>] [--no-routes]\n" \
 "                    [--no-dns] [--pppd-no-peerdns] [--pppd-log=<file>]\n" \
@@ -31,9 +31,10 @@
 "                    [--user-key=<file>] [--trusted-cert=<digest>]\n" \
 "                    [--use-syslog] [-c <file>] [-v|-q]\n" \
 "       openfortivpn --help\n" \
-"       openfortivpn --version\n"
+"       openfortivpn --version\n" \
+"\n"
 
-#define HELP \
+#define help_options \
 "Client for PPP+SSL VPN tunnel services.\n" \
 "openfortivpn connects to a VPN by setting up a tunnel to the gateway at\n" \
 "<host>:<port>. It spawns a pppd process and operates the communication between\n" \
@@ -85,7 +86,10 @@
 "                                to be even more verbose.\n" \
 "  -q                            Decrease verbosity. Can be used multiple times\n" \
 "                                to be even less verbose.\n" \
-"\n" \
+"\n"
+
+
+#define help_config \
 "Config file:\n" \
 "  Options can be taken from a configuration file. Options passed in the\n" \
 "  command line will override those from the config file, though. The default\n" \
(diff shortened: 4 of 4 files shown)


DimitriPapadopoulos:
@martinetd Thank you for these contributions. Separate merge requests might be easier to discuss:
* About the 1st commit: Is the length of string literals limited by the C standard, and is so how many characters can a string literal store? Out of curiosity, is this really more efficient, wouldn't the compiler optimize the multiple `USAGE` buffers away? That said the patch looks good, this is just about the comments.
* About the routing stuff: I'm afraid someone else will have to look into this - I'm not comfortable with that - but I am wondering whether all this needs to be in _openfprtivpn_. Could part of this stuff be done in an `/etc/ppp/ip-up.d/` script? Am I making sense at all?
* `tests/lint/run.sh` doesn't seem to like the last commit, is this something you can fix?

martinetd:
Happy to split things a bit, will do tomorrow evening.


* 1st commit: if you extend the string a bit more, you get this warning/error: src/main.c:238:4: error: string length ‘4338’ is greater than the length ‘4095’ ISO C99 compilers are required to support [-Werror=overlength-strings]
I'm sure gcc can deal with more, it's just probably a good point to split the string at.

Regarding optimization, I've just compiled the old code again and the "usage" string is indeed there twice (you can check with `strings`, or `objdump -j .rodata -s openfortivpn`). It's the only string that was used twice, so the overhead is small enough, but it is a gain :)

* Routing stuff, well, it doesn't *need* to be in _openfortivpn_ - I used to run it with --no-route and manually set the routes myself. I think more could benefit from the manual route specification, but anyone able to list their own routes for themselves could likely deal with it manually as well.
The default route as half internet though is fairly standard (it's what openvpn does for example) and I think could be useful for everyone.
I'll split this part off so we can discuss more lenghtly.

* Sorry, was waiting for travis to run lint for me as I didn't remember how to run it myself, but it turns out I can't access travis results either for some obscure reason. I've seen your pull request on my end, I don't like the idea of moving all the strings back to please lint (especially since that's not leaving much room for the string itself), I'll try to see how to fix that...

mrbaseman:
Hello, 

- I have hit the define-limit in the past. I believe that's one of the reasons why usage and help were two separate defines already, so +1 from my side for this commit
- I'm also fine with the changes of --set-routes and --set-dns
- the changes in the routing code revert improvements which we have implemented recently. The current mechanism adds a host route to the vpn gateway to keep the tunnel running even when the vpn pushes a route to its own network. In tunnel mode we remove the default route and restore it again when the vpn connection is closed. 
- Personally, I can't see the advantage of the half internet  route at the moment (perhaps I should find out more about this topic first. Do you have a reference at hand?). Anyhow, I would like to see this behavior enabled with an extra command line option instead of again changing the default behavior.
- For the custom routes I would vote for a solution with an ip-up script. This could for example be passed via --pppd-ipparam (if not hard-coded in the ip-up script).

martinetd:
Sorry for the slow reactions, been a busy week... I didn't know you could push to forks without fork owner interaction!

Anyway,

* if the help string split and set-routes/dns are fine I guess it's probably best to just merge that for now so that leaves less on the plate, do you mind if I massage your coding style fix in my commits so I can split the PR?

* The routing code should not revert anything about the improvements you made? I had actually written the same code a while before, so I had to rewrite it as I rebased forward (which is actually why I finally bothered pushing), but it still sets a /32 route to the tunnel endpoint as far as I can see.
I've just double-checked, can you point me to a part that looks wrong?

* The half internet route has two advantages:
 First is small, but basically means you don't have to remove default route, so you don't have to set it back up when we quit - less risk of restoring an old route that is no longer relevant if the user moved.
 Second is what I'm interested about, which is that the tunnel will survive a dhcp renew with that. If you just overwrite the default route, the dhcp renew will overwrite it back and the tunnel will no longer be used at this point - having the default route configured as two /1 subnets lets it stay around even with a new lease.
I know openvpn does the same (gateway-redirect def1) but haven't been able to find much reference about it; it's just a bit easier to use and behaves the same (/1 routes are more specific so will be used by default)

* I'll drop the last commit about custom routes, and use pppd-ipparam for myself; hadn't thought of using --pppd-ipparam (I want to be able to change behavior solely with invocation params).
Although it's a shame the server most can't control can push split-routes but not the user; it's just a nitpick at this point :)

Thanks!

mrbaseman:
@martinetd, I believe it is the default setting for pull requests that you grant the maintainers of the base fork the right to commit to the branch that constitutes the pull request. You could uncheck that box, in principle. I just thought I can try to make the PR pass the travis checks, simply by running `tests/lint/run.sh | patch -p1`.  Unfortunately, this ended up in several lines being too long due to the fact how astyle treats indentation for `static const char[]` instead of `#define`. Please feel free to update the changes again, and sqash the coding style commit (or drop mine and replace it with your variant).

About the routing code: Since recently we add a host route to the gateway through the previously existing interface. Some users had the problem that their VPN gateway pushes a route to a network which contains their own public IP. This was also not always done correctly in the default routing mode. Also, we store the existing default route which we restore when openfortivpn exits. The information is stored in ipv4_config but now you remove ppp_rt and def_rt from the structure. In the case of dhcp renewals for the default routing mode the half internet route approach may be more suitable, but for the case where the public IP of the gateway is part of a route received in split routes mode the existing mechanism with a host route and restoring the routing table at the end is better. Therefore, I would like  to add the half internet route approach with an extra command line switch.

I didn't get your final remark 
> Although it's a shame the server most can't control can push split-routes but not the user...

Maybe something gets lost when translating twice. Perhaps I can get the point when you express this in French? (or you rephrase it a bit?)

martinetd:
(ok for style stuff, will repush during the weekend)

Routing code- I don't see what using half internet or default gateway changes about either the tunnel endpoint or the client's public IP being in it:
 - We already protect the /32 route to the gateway, and still do.
 - I fail to see how having the public IP within the pushed route subnet affects things even more, mainly because it should not care (like, really, was there an issue I could look at?), but this is about split routes and not default route mode?
I only remove ppp_rt and def_rt in a separate patch because they are no longer used at all (since we do not overwrite the default route, it is still there when we exit - there is no need to restore it. The half internet routes we add are tied to the ppp interface and will automatically get dropped when the interface is destroyed, just as split routes are handled - there is no code that removes all the split routes at cleanup.
The code is really simpler, less error-prone (can't restore an old default route that changed during the time openfortivpn was running), and should work for everyone. I'm happy to make it an option, but I'll eat a hat if there is anything newly broken with that - unless they're already playing with /1 routes everything that was more specific than default is still more specific than a /1, and nothing changes really.

Final remark- I don't think language can help weird thinking pattern ;) I'm just mildly complaining that the last custom split route patch was using the same code servers use to push split route anyway (and that I can't control what my server wants to push), you can ignore that.

martinetd:
I've repushed just the help message split and set-routes patches for now, also added a trivial fix to pppd options (I guess not many are using these from command line!)

I still think half-internet routes is an improvement (all the original patches are still in my master branch), but I'm doing that in a pppd script now so there is no real use for me and I won't push further. If anything I just wish fedora's packaging of pppd would use ip-up.d like everyone else instead of a single ip-up/ip-up.local script -_-

DimitriPapadopoulos:
@martinetd @mrbaseman I'm specifically interested in this:
> ... the tunnel will survive a dhcp renew with that.

Is this currently not true? It might explain some issues I see from time to time. Is _openfortivpn_ supposed to survive a DHCP lease renew (I hope so)? Does it survive such a renew in practice? Are there timeouts involved here, such as the renew should not take more than _n_ seconds?

martinetd:
From my experience, openfortivpn (the program) will survive a renew but the default route to ppp0 will be dropped and replaced by the dhcp default route.
This might be due to my setup however (I usually notice this when roaming, the dhcp client will get the same IP so sesssions etc will survive but dhcpcd/dhclient might be a brand new one and it can behave differently from a persistent dhcp client that just renews a lease) - so ultimately I might be fixing a problem I created... But I don't think I'm the only one with that problem :)
It's easy enough to check if you're in that case. The next time you see connections in the tunnel hanging, you can check and see if openfortivpn is still running/there still is a ppp interface, but 'ip route' will show another interface for the default route.

I think it should survive as long as the renew keeps the same IP. The tunnel connection has no reason to be dropped and things work fine for me with half-internet routes as I use now.

mrbaseman:
the half internet routes are not part of this PR anymore? My intention was not to reject them. I just didn't want to change the default behavior once more and rather add this as a separate option. Would you like to add this? Or open a separate PR once this one is merged? Or shall I add the command line option around your routing improvement? I'm fine with either of the three options.

martinetd:
I removed the "more difficult" commits a few comments ago ("I've repushed just the help message split and set-routes patches for now, also added a trivial fix to pppd options[…]"), with the intention of either pushing a new PR or opening an issue for the half-internet routes once the easy bits were merged.

Everything is still in my master branch so it's still accessible if you'd like to merge everything right now, but I am happy to discuss this part longer unless you think it's already good.
I'm not sure I understood the third option, if it's referring to the --custom-routes new option I thought everyone agreed to not take this patch in favor of a pppd script?

mrbaseman:
I would like to add the half internet route feature, but I would like to have a command line switch to select this instead of the current behavior. But, agreed, let's put this into a separate pull request. I'll merge the current one, now.

martinetd:
Ok, thanks.

I'm a bit busy right now, happy to let you pick the half internet route from my master and do the command-line/config switch, or I will do it over the weekend if you're not in a hurry.
I don't mind either way.

mrbaseman:
I have cherry-picked your half-internet-routes commit and added the config-file/command-line switch in the above PR.

Issue #145, opened 2017-06-22
support openssl 1.1.x


mrbaseman:
In the long term we should support openssl 1.1

Currently, when linking against openssl 1.1.x,  compilation doesn't succeed and therefore we have added a check in the configure script to preserve users from running configure successful and failing afterwards at compile time

mrbaseman:
I have just seen that openssh is discussing the same topic on their [devel list](https://lists.mindrot.org/pipermail/openssh-unix-dev/2017-June/thread.html) and also OpenVPN has made an effort to [support 1.1](https://www.mail-archive.com/openvpn-devel%40lists.sourceforge.net/msg14946.html). I'm not expecting that approaches can be applied 1:1, but for reference of more general discussions and a first impression of the consequences of the [API changes](https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes) the former links could be good resources, the latter one being the official documentation.

rahra:
What exactly is the problem with OpenSSL 1.1?
I just compiled against 1.1.0f and it compiles and the client works.

mrbaseman:
I have seen the following error messages: 
> src/io.c:68:13: error: ‘lock_callback’ defined but not used [-Werror=unused-function]
>  static void lock_callback(int mode, int type, char *file, int line)
>              ^
> src/io.c:75:22: error: ‘thread_id’ defined but not used [-Werror=unused-function]
>  static unsigned long thread_id()
>                       ^
> cc1: all warnings being treated as errors
> Makefile:532: recipe for target 'src/openfortivpn-io.o' failed
> make: *** [src/openfortivpn-io.o] Error 1

and indeed with https://github.com/adrienverge/openfortivpn/commit/3531e2bf29f7dba8af78f1e647a04e63067255d6 they go away and the code at least compiles. 
When I have opened this issue I didn't have a deeper look if it was such a small fix or if larger code areas need changes after the first warnings are fixed.
Also, I haven't tested yet if the connection actually works. It might also be something to test against different versions of FortiOS on the server side.

rahra:
Yes, but these errors are not related to openssl.
It is the code checking option "-Werror" of gcc. These compiler errors occur because the functions are declared "static" and are not called within this file because they are callback functions for openssl.

fabiopicchi:
[...]

Issue #75, opened 2016-11-04
configure option --sysconfdir is not honored during runtime


mrbaseman:
the configure option --sysconfdir is honored during installation and the config file is correctly placed there. However, when running openfortivpn without command line options it searches the config in the hard coded place /etc/openfortivpn/config, even when --prefix or --sysconfdir at configure time points somewhere else. This could be improved by passing sysconfdir somehow as a macro to the preprocessor. I have been playing around for a while with autoconf, but didn't find a robust solution, yet.

adrienverge:
@mrbaseman Thanks for pointing this problem out. I'm open to contributions towards this.

mrbaseman:
the pull request has been merged, thank you. Closing this issue.

PR #151, opened 2017-07-10
Some random fixes

lkundrak:
Here's a couple of commits I'd like to see in openfortivpn.

In particular, I care about the "otp" in configuration file commit. There's one bugfix that looks semi-serious.

Sorry for posting unrelated patches in a bulk -- feel free to cherry-pick at your convenience.


mrbaseman:
Hi @lkundrak,
thanks for these comits. Let me comment on them to start up discussion:
* thanks for the Openssl 1.1.0 port - nice to see that it's just a few lines. I have already opened #145 for this but expected a lot more to do. This is something where we probably should have a few confirmations that it works, but I guess we are not in a hurry here. In configure.ac there is a hard-coded version check now. Your remove that with the next commit, but just in case someone wants to test this commit separately, then a modified configure script is needed, too.
* pkg-config for detecting OpenSSL: I'm a bit in a hurry and didn't test it yet: Does it provide a configure option to the user to specify a non-default location of the ssl library? That was the main reason for the lengthy autoconf macro added in #143 , but if your much more elegant solution provides the same functionality, I'm happy to merge that one
* imo. `$prefix` should default to the empty string and when evaluating `$sysconfdir` it should contain the configure option `--prefix`, or, if not supplied, just not appear there. In previous versions (before #75) openfortivpn was always looking for its configuration in /etc, regardless of the installation prefix, which was not correct behavior, too. But maybe there are still be cases left where the outcome is not what one would expect. If I recall correctly, there is also a line in the Makefile or in the headers that's fed with the result of this `eval`
* thanks for the http-fix - maybe a call to memset would be better portable?
* --otp as a config option really in a file? How would that work? It must be a file created on the fly e.g. in a ramdisk where otp from user input is filled into a template before starting up openfortivpn, right?
* about dropping -Wall and -Werror I would let the final decision up to @adrienverge

adrienverge:
Thanks for these fixes! Can we keep `-Wall` (to report potential problem) and put `-Werror` in `make distcheck` (to ease maintainers' work)?

lkundrak:
> thanks for the Openssl 1.1.0 port - nice to see that it's just a few lines. I have already opened #145 for this but expected a lot more to do. This is something where we probably should have a few confirmations that it works, but I guess we are not in a hurry here. In configure.ac there is a hard-coded version check now. Your remove that with the next commit, but just in case someone wants to test this commit separately, then a modified configure script is needed, too.

Well, I've tested this.

I fact, I'm a bit concerned about whether the code is correct for OpenSSL 1.0 in the first place, because [1] claims that "using a single SSL object in two threads, one each for reading and writing" is wrong, and I think that's IIRC exact
ly what openfortivpn does (?). If that is the case, then we perhaps should consider replacing the threads altogether with a poll(2) loop.

[1] https://www.openssl.org/blog/blog/2017/02/21/threads/

In any case, OpenSSL 1.1 doesn't make things any worse in this respect, regardless what the new semantics of locking is (since the important part seems to be that now OpenSSL takes care of platform native locking specifics by itself as opposed to letting the application deal with the locking primitives).

> pkg-config for detecting OpenSSL: I'm a bit in a hurry and didn't test it yet: Does it provide a configure option to the user to specify a non-default location of the ssl library? That was the main reason for the lengthy autoconf macro added in #143 , but if your much more elegant solution provides the same functionality, I'm happy to merge that one

I think it works well in such cases in general. See the remark about PKG_CONFIG_PATH in the commit message, which I believe should be used.

> imo. $prefix should default to the empty string and when evaluating $sysconfdir it should contain the configure option --prefix, or, if not supplied, just not appear there.

That's not how it works at configure time. What you say makes sense at build time.

> In previous versions (before #75) openfortivpn was always looking for its configuration in /etc, regardless of the installation prefix, which was not correct behavior, too. But maybe there are still be cases left where the outcome is not what one would expect. If I recall correctly, there is also a line in the Makefile or in the headers that's fed with the result of this eval

Yes. I'm replacing that with a -D on a command line determined at build time as opposed by a config.h #define.

> thanks for the http-fix - maybe a call to memset would be better portable?

Well, given we build with -std=gnu99 by default, so there's not much point in that. If find the initializer to look slightly better.

> --otp as a config option really in a file? How would that work? It must be a file created on the fly e.g. in a ramdisk where otp from user input is filled into a template before starting up openfortivpn, right?

Yes, precisely as you say (and as is described in the commit message). Here's the NM plugin patch: https://bugzilla.gnome.org/show_bug.cgi?id=777404

> Thanks for these fixes! Can we keep -Wall (to report potential problem) and put -Werror in make distcheck (to ease maintainers' work)?

Okay. Updating the branch soon.

lkundrak:
Merging this now. I believe I addressed the concerns and hopefully there's nothing controversial there.

mrbaseman:
the configure options --with-openssl and alike just introduced with #144 have been replaced by the $PKG_CONFIG_PATH now. They were introduced on request of the Homebrew maintainer [here](https://github.com/Homebrew/homebrew-core/pull/14344) - however, I fail to see where the formula makes use of this option. Maybe Homebrew is fine with pkg_config as well, or we should re-introduce the configure option again for compatibility reasons of the next release. I just did not yet merge this PR before finding this out (but maybe it's just a matter of getting a confirmation that homebrew is fine with the latest commits)

PR #152, opened 2017-07-21
Fix bug in pppd_run forking code

terdei:
The error handling of execvp in the forked child is bad.
Instead of returning into the original forked process, it should
exit immediately after execvp. This only happens if execvp fails
for some reason, which is probably a very rare situation.
But anyway it is better to be fixed.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -144,7 +144,8 @@ static int pppd_run(struct tunnel *tunnel)
 
 		close(tunnel->ssl_socket);
 		execvp(args[0], args);
-		exit(errno);
+		fprintf(stderr, "execvp: %s\n", strerror(errno));
+		exit(EXIT_FAILURE);
 	}
 
 	// Set non-blocking
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -143,10 +143,8 @@ static int pppd_run(struct tunnel *tunnel)
 		assert (i < sizeof (args) / sizeof (*args));
 
 		close(tunnel->ssl_socket);
-		if (execvp(args[0], args) == -1) {
-			log_error("execvp: %s\n", strerror(errno));
-			return 1;
-		}
+		execvp(args[0], args);
+		exit(errno);
 	}
 
 	// Set non-blocking
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -167,7 +167,14 @@ static int pppd_terminate(struct tunnel *tunnel)
 	close(tunnel->pppd_pty);
 
 	log_debug("Waiting for pppd to exit...\n");
-	waitpid(tunnel->pppd_pid, NULL, 0);
+	int status;
+	if (waitpid(tunnel->pppd_pid, &status, 0) == -1) {
+		log_error("waitpid: %s\n", strerror(errno));
+		return 1;
+	}
+	if (WIFEXITED(status)) {
+		log_debug("waitpid: exit status code %d", WEXITSTATUS(status));
+	}
 
 	return 0;
 }


DimitriPapadopoulos:
Don't we still want to log the error?

Something like:
```
execvp(args[0], args)
int errcode = errno;
log_error(... ...);
exit(errcode);
```

terdei:
Hi Dimitri!

log_error uses a mutex, that will become invalid when the new process is forked. It might be in a locked state at the time of forking, and in this case,  log_error in the child will wait forever.
So I think it is safer to skip logging in this case.

DimitriPapadopoulos:
I see... Then would a mere printf() make sense or is this too much hassle for not benefit? What happens if `/usr/sbin/pppd` cannot be run or is not available for some reason? Will any diagnostic be printed or not?

DimitriPapadopoulos:
Also do you mind if we keep changes to the code as minimal as possible - although I understand the _if_ is not really useful here!):
```
		if (execvp(args[0], args) == -1) {
			exit(errno);
		}
```

terdei:
I think it is safer not to use an if block around exit.
In case pppd can not be launched, then the exit code (sent by the exit call after execvp) will be received by waitpid in pppd_terminate. We could extend that to log the appropriate error message there.

DimitriPapadopoulos:
OK then. I think I've seen the useless _if_ in other software too - but it remains useless!

Adding a log in the parent process looks like a nice addition.

DimitriPapadopoulos:
I have read [elsewhere](https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html) that `errno` should not be used for exit status, among other reasons because its value might exceed 255. Doesn't this apply here? If it does, how to send information about the cause of the _exec()_ failure to the parent process?

DimitriPapadopoulos:
Said otherwise [process exit status](https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html) (as defined in [_sysexit.h_](http://sourceware.org/git/?p=glibc.git;a=blob;f=misc/sysexits.h) for C programmers or the [Advanced Bash-Scripting Guide](http://tldp.org/LDP/abs/html/exitcodes.html) for shell programmers) and C [system error number](https://www.gnu.org/software/libc/manual/html_node/Checking-for-Errors.html)  (`errno`) are different beasts and their values should not be mixed.

terdei:
Yes, that is correct. Do you have any idea how the `execvp` failure code could be sent back to the invoking process?
I checked the `errno` constants on Linux: all those that are mentioned as a possible error code for the `execve` syscall are under 255. So on Linux, it could be safe to use `exit(errno)` in our particular case.
I know this is just Linux, and other platforms might have different constant values, of course.

Other option could be to simply print the error message to STDERR, and then do `exit(1)`.

DimitriPapadopoulos:
I have added some error logging around _waitpid()_ to this pull request.

However I'm afraid I am unable to predict possible values of `errno` after _execvp()_, which could perhaps be translated to process exit status codes with similar semantics. Perhaps a simple `EXIT_FAILURE` would be sufficient...

mrbaseman:
I just wanted to propose this
```
		execvp(args[0], args);
		log_error("execvp: %s\n", strerror(errno));
		exit(1);
```
and noticed that you were describing the same in words ;-)

DimitriPapadopoulos:
@mrbaseman Unfortunately _log_error()_ is not an option because it uses a mutex. See above.

mrbaseman:
oh, I missed that part of the discussion, and your [commit](https://github.com/adrienverge/openfortivpn/pull/152/commits/0aecddf3cadea57889baf236090c37bfe0a55f9f) looks better then.

DimitriPapadopoulos:
@terdei Your suggestion looks like a good solution:
```
		execvp(args[0], args);
		fprintf(stderr, "execvp: %s\n", strerror(errno));
		exit(EXIT_FAILURE);
```

An alternative would be to rewrite _do_log()_ as follows and use the mutex-less _unprotected_do_log()_ after _execvp()_:
```
void do_log(int verbosity, const char *format, ...)
{
	va_list args;

	pthread_mutex_lock(&mutex);

	unprotected_do_log(verbosity, format, args);

	pthread_mutex_unlock(&mutex);
}
```
Is that safe enough?

terdei:
@DimitriPapadopoulos I changed pppd_run to print the error to STDERR before exiting. Also updated the branch with master.
What do you think, can these changes be pulled in to the master?

DimitriPapadopoulos:
Looks good to me. Hopefully it compiles on Mac OS too, I don't have a Mac to test.

PR #153, opened 2017-07-21
Add script option

terdei:
The script specified will be run after tunnel buildup and before teardown. Very similar to OpenVPN's up/down options.
The script will receive info about the tunnel as VPN_* environment variables.
Use cases are implementing split DNS, and/or overriding gateway specified routes.

After implementing this feature for myself, I noticed #101, and the option to use pppd ip-up/ip-down scripts for this task. So this is an alternative to that.

### DIFF
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -22,6 +22,7 @@ openfortivpn \- Client for PPP+SSL VPN tunnel services
 [\fB\-\-pppd-log=\fI<file>\fR]
 [\fB\-\-pppd-plugin=\fI<file>\fR]
 [\fB\-\-pppd-ipparam=\fI<string>\fR]
+[\fB\-\-script=\fI<path to script>\fR]
 [\fB\-c\fR \fI<file>\fR]
 [\fB\-v|\-q\fR]
 .br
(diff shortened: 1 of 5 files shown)


DimitriPapadopoulos:
I have modified your branch so that it passes the _lint_ test.

terdei:
Thanks for fixing that. I have seen the lint errors in travis, but have not had the time so far to fix it.

lkundrak:
The commits should be logical units with the commit message describing the nature of the change and the reasoning behind it.

You seem to have split a single change into multiple commits and didn't provide explanations, which make it sort of difficult to review the changes. Perhaps they could just be squashed together?

terdei:
@lkundrak : I agree with you, my merge request contained only a sinlge commit, but turned out that it had lint errors. Dimitri fixed this, so the other commits came from this.
And then, later the master was updated, so I merged the change to the new master.
Question is: how to fix this. Shall I close this mr, and create a new one on top of the latest master, also taking care of lint checks?
I am happy to do this, if there is a chance that my change will be merged in. But I would not waste my time, if the change gets rejected for other reasons.
Pls. let me know how to move on.

DimitriPapadopoulos:
@terdei While a `--script` option is probably easier for end-users, using _pppd_ ip-up/ip-down scripts is probably better for Linux packagers. See also new option `--pppd-call` introduced by #270 which paves the way to using _openfortivpn_ as a regular user but requires a _pppd_ configuration file in `/etc/ppp` anyway.
I'd rather avoid multiple methods to run scripts. How do you feel about this? Am I missing a use case where `--script` makes a difference?

mrbaseman:
can we close this one? We have added some information on our [wiki](https://github.com/adrienverge/openfortivpn/wiki) how to use the already existing pppd mechanism

mrbaseman:
what shall we do with this one? One would probably have to squash and rebase the changes manually, but as stated already before, similar functionality is provided by pppd ip-up/down scripts. 
Therefore, can we close this one?

mrbaseman:
Oh, I have scrolled over my [earlier message](https://github.com/adrienverge/openfortivpn/pull/153#issuecomment-439734912) - so sorry for asking the same question again. Given the fact that nobody has replied in a month I assume we can close it.

PR #158, opened 2017-07-27
Add half-internet-routes as an option to influence how default routing is handled

mrbaseman:
This merges the routing improvement proposed by @martinetd in #149 together with a command line option to select this type of behavior explicitly. The corresponding config file option is also added and the man page and help message are updated to reflect the new option


martinetd:
Haven't had a chance to test yet but it looks good, thanks for the option-ization.
I'll remove this part from my ppp script and test at some point but please don't wait for me :)

Issue #139, opened 2017-06-08
dynamic number of split routes


mrbaseman:
in the past MAX_SPLIT_ROUTES has been increased more and more (#110, #136) to address border cases. In #136 we have agreed on first setting it to 128 to address the current request in #87 and in another step modify the code such that it can dynamically increase the number of routes. A mechanism to increase the array while saving its content is needed to achieve this. Also, the dynamically allocated memory should be freed at the end.

mrbaseman:
the solution for this has already been merged, so closing this issue

PR #163, opened 2017-08-16
dynamic allocation of memory for split route array

mrbaseman:
proposed fix for #139

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -510,10 +510,19 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
                              char *gateway)
 {
 	struct rtentry *route;
-	char env_var[22];
+	char env_var[24];
 
-	if (tunnel->ipv4.split_routes == MAX_SPLIT_ROUTES)
-		return ERR_IPV4_NO_MEM;
+	if ((tunnel->ipv4.split_rt == NULL)
+	    || ((tunnel->ipv4.split_routes % STEP_SPLIT_ROUTES) == 0)) {
+		void *new_ptr
+		        = realloc(
+		                  tunnel->ipv4.split_rt,
+		                  (size_t) (tunnel->ipv4.split_routes + STEP_SPLIT_ROUTES)
+		                  * sizeof(*(tunnel->ipv4.split_rt))
+		          );
+		if (new_ptr == NULL) return ERR_IPV4_NO_MEM;
+		tunnel->ipv4.split_rt=new_ptr;
+	}
 
 	sprintf(env_var, "VPN_ROUTE_DEST_%d", tunnel->ipv4.split_routes);
 	setenv(env_var, dest, 0);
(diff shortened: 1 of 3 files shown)


src/ipv4.c
```diff
@@ -510,10 +510,19 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
                              char *gateway)
 {
 	struct rtentry *route;
-	char env_var[22];
-
-	if (tunnel->ipv4.split_routes == MAX_SPLIT_ROUTES)
-		return ERR_IPV4_NO_MEM;
+	char env_var[24];
+
```
mrbaseman: Just noticed that I forgot to check for MAX_SPLIT_ROUTES here

src/tunnel.c
```diff
@@ -486,10 +486,6 @@ int run_tunnel(struct vpn_config *config)
 	struct tunnel tunnel;
 
 	memset(&tunnel, 0, sizeof(tunnel));
-#ifdef __APPLE__
-	// initialize value
-	tunnel.ipv4.split_routes = 0;
-#endif
 	tunnel.config = config;
```
mrbaseman: This initialzation should happen as part of the memset above.

src/ipv4.c
```diff
@@ -510,10 +510,19 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
                              char *gateway)
 {
 	struct rtentry *route;
-	char env_var[22];
-
-	if (tunnel->ipv4.split_routes == MAX_SPLIT_ROUTES)
-		return ERR_IPV4_NO_MEM;
+	char env_var[24];
```
mrbaseman: Increase string length for very large numbers of routes

src/ipv4.h
```diff
@@ -76,7 +77,7 @@ struct ipv4_config {
 	struct rtentry	def_rt; // default route
 	struct rtentry	gtw_rt; // route to access VPN gateway
 	struct rtentry	ppp_rt; // new default route through VPN
-	struct rtentry	split_rt[MAX_SPLIT_ROUTES]; // split VPN routes
+	struct rtentry	*split_rt; // split VPN routes
```
mrbaseman: Do we have to free this explicitly at the end?

DimitriPapadopoulos:
I would like to suggest changing:
```
if (tunnel->ipv4.split_routes == MAX_SPLIT_ROUTES) return ERR_IPV4_NO_MEM;
```
to:
```
if (tunnel->ipv4.split_routes == MAX_SPLIT_ROUTES)
    return ERR_IPV4_NO_MEM;
```

See for example [Linux kernel coding style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation):

> Don’t put multiple statements on a single line unless you have something to hide:
> ```
> if (condition) do_this;
>   do_something_everytime;
> ```

DimitriPapadopoulos:
Same goes for:
```if (new_ptr == NULL) return ERR_IPV4_NO_MEM;```

src/ipv4.h
DimitriPapadopoulos: It's probably better to free explicitly - might be handy in case we add the feature where _openfortivpn_ reconnects automatically after a time-out.

mrbaseman:
@DimitriPapadopoulos thanks for reviewing. I have addressed your points and then squashed and merged this PR.

PR #165, opened 2017-08-19
Add pppd ifname option

jisaitua:
When having multiple VPNs it desirable to give meaninful names to the interfaces. This pull request adds the ifname option to the pppd daemon.


src/tunnel.c
```diff
@@ -139,6 +139,10 @@ static int pppd_run(struct tunnel *tunnel)
 			args[i++] = "ipparam";
 			args[i++] = tunnel->config->pppd_ipparam;
 		}
+		if (tunnel->config->pppd_ifname) {
+			args[i++] = "ifname";
+			args[i++] = tunnel->config->pppd_ifname;
+		}
 		// Assert that we didn't use up all NULL pointers above
```
mrbaseman: you should add two more NULL entries in the args[] array above just in case anyone uses all the possible arguments to pppd

mrbaseman:
the man page should also mention this new option

DimitriPapadopoulos:
@jisaitua Just trying to understand the use case here. What is the rationale for this patch? Is it needed when **running** multiple VPNs **at the same time**? I've never tested it because this is often not possible. For example once connected to my organization's FortiGate, all other outgoing VPN connections are blocked by routing rules and firewalls.

ziesemer:
@DimitriPapadopoulos - One example for which I'm looking for this:  As a consultant, I may connect to a different client network every day.  Each has a different configuration, and different firewall rules I want in my own iptables for each.  Instead of always being `ppp0`, I'd like to be able to do `vpn-client1`, `vpn-client2`, etc.  I could then keep appropriate rules for each that I don't need to worry about reloading each time.  I could then even do a wildcard on `vpn-*`, without needing to worry about other `ppp` connections.  In some rare occurrences, I need to use one VPN to get to another.  Most of the other VPN clients I've used (OpenVPN, OpenConnect, etc.) also support using a specific name for a VPN interface.

DimitriPapadopoulos:
@ziesemer Makes sense indeed. I'll attempt to rebase the patch at some point and merge it. Are you willing to test it?

ziesemer:
@DimitriPapadopoulos - Sure thing.

jisaitua:
@DimitriPapadopoulos We maintain many VPN connections with different customers and we need to monitor the state of those connections. When setting meaningful names for each interface (i.e: vpn-customerA), we can have a direct relationship with the name of the interface and the customer, which makes things easier for us. It is also useful for the maintenance of our iptables rules.

DimitriPapadopoulos:
@jisaitua and @ziesemer I have updated the `pppd-ifname-option` branch and added the two missing NULLs in `args`. Can you give this branch a try and verify it works for you?

ziesemer:
@DimitriPapadopoulos - 0a52999 seems to work perfectly, thanks!

Just to note, the man page will also need to be updated.

DimitriPapadopoulos:
Unfortunately the `ifname` option does not look like a standard _pppd_ option.

See [PPPD(8) manual page](https://ppp.samba.org/pppd.html#sect4), "[Naming PPP interfaces](https://unix.stackexchange.com/questions/145692/naming-ppp-interfaces#311226)", this [`AdrianBan:ppp-debian`](https://github.com/AdrianBan/ppp-debian) patch, or the "[Changing interface names in pppd, where is the ifname option?](https://lists.ubuntu.com/archives/ubuntu-users/2009-March/178455.html)" thread of discussion of the _ubuntu-users_ mailing list.

DimitriPapadopoulos:
@mrbaseman The `ifname` option is not part of the vanilla _pppd_ release but a patch has been added by Debian and appears to be used by most recent distributions. Checking whether the `ifname` option is available at runtime is overkill - I wouldn't like to spend time on this. Do you agree it is not worth checking at run time and that an error message from _pppd_ if it doesn't support the option is sufficient?

DimitriPapadopoulos:
@jisaitua and @ziesemer It looks like the recommended way of renaming a device is to use _udev_:
* [Rename Network Interface using udev in Debian](http://www.debianhelp.co.uk/udev.htm)
* [Understanding the Device Renaming Procedure](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/networking_guide/sec-understanding_the_device_renaming_procedure)

On the other hand this does not seem to work well with _pppd_:
* [Renaming NextG PPP](http://forums.whirlpool.net.au/archive/957998#rr19170038)

Can you comment on that? Do you have experience with _udev_ and _pppd_?

mrbaseman:
- If `pppd` shows a message that it doesn't understand the `ifname` parameter at runtime I think this is sufficient. I can't test this, because my pppd understands it. 

- In the man page we should put a note that not all pppd implementations understand it

DimitriPapadopoulos:
@mrbaseman Unfortunately, after applying pull request #210 and adding a random bogus option such as `"dsdzedzedzedzed"`to `args`, the error messages are not very helpful:
```
INFO:   Cancelling threads...
ERROR:  pppd: An immediately fatal error of some kind occurred, such as an essential system call failing, or running out of virtual memory.
INFO:   Terminated pppd.
```
It's the best I can do for now, until I find a way to:
* either undo the effect of `forkpty()` and get the child process to print to `stderr`,
* or read the error output of the child process from the parent process and print it to the parent's `stderr`.

Both will probably be necessary for good error reporting, both for cases where `execvp()` is unable to start `pppd`, and cases where `pppd` starts and exits, for example because of a bad option.

I nevertheless suggest we apply the patch. I am working on `pppd` error reporting anyway in other pull requests.

mrbaseman:
Ok, fine with me. However, it seems we have a merge conflict now. Anyhow, I think we should squash the various commits.

DimitriPapadopoulos:
Then please commit whatever you want first. I'll manage the conflicts afterwards.

DimitriPapadopoulos:
OK, I will resolve conflicts and squash commits.

DimitriPapadopoulos:
Merging... I suggest I open a new issue for the `stderr` issue.

PR #183, opened 2017-09-25
Fix warning from static analysis tool scan-build

DimitriPapadopoulos:
This static analysis tool emits this single warning.

Once fixed, I could have Travis run _scan-build_ automatically.

Also the code was "prettyfied" for consistency between `pool_pop()` and `pool_push()`.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -138,10 +138,10 @@ static void pool_push(struct ppp_packet_pool *pool, struct ppp_packet *new)
  */
 static struct ppp_packet *pool_pop(struct ppp_packet_pool *pool)
 {
-	struct ppp_packet *first = pool->list_head;
+	struct ppp_packet *first;
 
-	//SEM_WAIT(&pool->sem);
 	pthread_mutex_lock(&pool->mutex);
+
 	while (pool->list_head == NULL)
 		pthread_cond_wait(&pool->new_data, &pool->mutex);
 


src/io.c
```diff
@@ -138,10 +138,10 @@ static void pool_push(struct ppp_packet_pool *pool, struct ppp_packet *new)
  */
 static struct ppp_packet *pool_pop(struct ppp_packet_pool *pool)
 {
-	struct ppp_packet *first = pool->list_head;
+	struct ppp_packet *first;
```
adrienverge: Wouldn't `struct ppp_packet *first = NULL;` be safer?

DimitriPapadopoulos: I don't know. This is a pretty simple function, `first` is set a couple lines below:
```
	struct ppp_packet *first;

	pthread_mutex_lock(&pool->mutex);

	while (pool->list_head == NULL)
		pthread_cond_wait(&pool->new_data, &pool->mutex);

	first = pool->list_head;
```

Besides the scope of static analysis tools is precisely to catch such errors, so they should warn us if a variable was to be used before being initialized.

In any case I tried to be consistent between `pool_pop()` and `pool_push()`. If we add `NULL` to `pool_pop()`, we should add `NULL` to `pool_push()`.

adrienverge: My bad, I misread this commit. There is absolutely no problem with your change, since `first` is set on line 148.

> Besides the scope of static analysis tools is precisely to catch such errors, so scan-build should warn us if a variable was to be used before being initialized.

Yes, and using `scan-build` would be an appreciated change. Thanks for the work!

PR #185, opened 2017-09-28
Fix Coverity defects

DimitriPapadopoulos:


src/ipv4.c
```diff
@@ -848,10 +848,10 @@ int ipv4_del_nameservers_from_resolv_conf(struct tunnel *tunnel)
 
 	ret = 0;
 
-err_free:
-	free(buffer);
 err_close:
 	fclose(file);
+err_free:
+	free(buffer);
```
adrienverge: Why?

In case `fstat(fileno(file), &stat)` fails, the program will jump to `err_close`, close the file and free `buffer` (which is problem since `buffer` is `NULL`).

DimitriPapadopoulos: It's not a problem because `free(NULL)` is safe as documented in the man page:
> If ptr is NULL, no operation is performed.

On the contrary `fclose(NULL)` is not safe:
> The behaviour of fclose() is undefined if the stream parameter is an illegal pointer, or is a descriptor  already  passed to a previous invocation of fclose().

Said otherwise, I could add `if (buffer)` or `if (file)` in front of `free()` and `fclose()` to be on the safe side, but that's not needed any more because:
* `free()` already checks `NULL` arguments internally,
* with the new order of the `err_close` and `err_free` labels, `fclose()` will never be called with a `NULL` argument.

adrienverge: Hi,

You're right, it would not be a real problem for `free()`, but all along the
program's code we use the common way of stacking / unstacking `goto`s:

```c
void foo()
{
    if (!doA())
        goto exit;      // nothing to clean up
    if (!doB())
        goto cleanupA;  // will clean up doA
    if (!doC())
        goto cleanupB;  // will clean up doB then doA

    /* everything has succeeded */
    return;

cleanupB:
    undoB();
cleanupA:
    undoA();
exit:
    return;
}
```

This way of ordering `goto`s is classic and widely used, I'd prefer we stick to
it for clarity. What do you think?

DimitriPapadopoulos: Sure, we can do that, but need to change to:
```
err_free:
	free(buffer);
err_close:
	if (file)
		fclose(file);
```

DimitriPapadopoulos:
Forced new the proposed changes into the `coverity` branch.

src/ipv4.c
adrienverge: Yes, totally right!

DimitriPapadopoulos:
After today's last changes:
* Coverity defects have been eliminated.
* I have tested this last version and it works just fine for me.

I plan on merging into `master`, then creating a `coverity_scan` for automating scanning as documented in [Travis CI Integration ](https://scan.coverity.com/travis_ci).

adrienverge:
Sounds good to me!

Can you squash "Fix Coverity defect" commits into one, for clarity?

DimitriPapadopoulos:
I would rather keep commits as small and specific as possible - each commit fixes a different type of defect. I find it easier to understand more detailed logs months after the fact. But then if you really prefer a single commit it's also OK with me.

adrienverge:
No it's good for me, I understand your point!

PR #189, opened 2017-10-06
Print message associated to pppd exit status code

### LINKED ISSUE #177: Improve pppd debugging
Link pppd numerical exit status with human readable error messages as documented in the [PPPD(8) manual page](https://ppp.samba.org/pppd.html#sect12).

DimitriPapadopoulos:
Fixes #177.


DimitriPapadopoulos:
@mrbaseman Does this look good? I hope it can help debugging some "I/O error" occurrences.

mrbaseman:
Yes, it looks good to me. A few remarks:
The messages are only documented in the pppd man page, not in the header. Therefore, we have to include them ourselves if we want to print them out in clear text. 
I hope the error codes have the same meaning on all platforms, but we still print out the integer value, so in case there are implementations where different values are used one can still look up their meaning in the man page.
I wasn't completely sure if the arithmetic to calculate the array size is correct for arrays of strings, but a short test showed me that it indeed works (the entries in the array actually are pointers to the string constants).

Issue #92, opened 2017-01-19
Connection dies with 'Protocol violation with EOF' after some time


dwt:
Not sure what to make of it - the tunnel comes online and works for a short time, then dies with these error messages:

```
DEBUG:  Error reading from SSL connection (Protocol violation with EOF).
INFO:   Cancelling threads...
INFO:   Setting ppp interface down.
INFO:   Restoring routes...
INFO:   Removing VPN nameservers...
DEBUG:  Waiting for pppd to exit...
INFO:   Terminated pppd.
INFO:   Closed connection to gateway.
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Logged out.
```

Running master from 19. of January 2017 on Mac OS X 10.12.2 (16C68).

What further information can I give to help you debug this problem?

mrbaseman:
The connection over which the tunnel was running was stable during that time? Any log entries of pppd which could provide a hint?

dwt:
The connection was stable during that time, I'm not really sure how to reproduce this, it seems that the tunnel breaks sometimes very few seconds after establishing (the first time I connected) then it worked flawlessly for half an hour (till I hang up). The last time I used it, I couldn't get any connection to work for longer than 30 seconds before it broke.

Today I only got these messages out of pppd (and IMO they look fine?). The first connect failed, second and third where fine and manually closed. (I don't really see that first connect in the log though, not sure why)

```
Standard	10:57:00.903531 +0100	pppd	publish_entry SCDSet() failed: Success!
Standard	10:57:00.903579 +0100	pppd	publish_entry SCDSet() failed: Success!
Standard	10:57:00.904576 +0100	pppd	pppd 2.4.2 (Apple version 838) started by root, uid 0
Standard	10:57:00.905492 +0100	pppd	Connect: ppp0 <--> /dev/ttys001
Standard	10:57:01.387849 +0100	pppd	local  IP address 172.16.0.4
Standard	10:57:01.387951 +0100	pppd	remote IP address 1.1.1.1
Standard	10:57:01.388102 +0100	pppd	Committed PPP store
Standard	10:57:01.388942 +0100	pppd	Committed PPP store
Standard	10:59:27.025956 +0100	pppd	Modem hangup
Standard	10:59:27.027486 +0100	pppd	Connection terminated.
Standard	10:59:27.027635 +0100	pppd	LCP close (User request).
[...]

PR #190, opened 2017-10-06
added --loop option for automatic reconnects

mrbaseman:


src/main.c
```diff
@@ -255,6 +258,17 @@ int main(int argc, char **argv)
...
 			}
 			if (strcmp(long_options[option_index].name,
+			           "loop") == 0) {
+				long int loop = strtol(optarg, NULL, 0);
+				if (loop < 0) {
+					log_warn("Bad loop option: " \
+					         "\"%s\"\n", optarg);
+					break;
+				}
+				cfg.loop = loop;
```
DimitriPapadopoulos: A `long` is cast into an `int`. I'm not programming much in C these days, but couldn't this be an issue if `loop` is insanely large?

src/main.c
```diff
@@ -380,6 +394,11 @@ int main(int argc, char **argv)
 
 	if (run_tunnel(&cfg) == 0)
 		ret = EXIT_SUCCESS;
+	while ((ret == EXIT_SUCCESS) && (cfg.loop!=0) && (get_sig_received()==0)) {
+		sleep(cfg.loop);
+		if (run_tunnel(&cfg) != 0)
+			ret = EXIT_FAILURE;
+	}
```
DimitriPapadopoulos: Less code, but I don't know if it is readable. What do you think?
```
	ret = EXIT_SUCCESS;
	while ((ret == EXIT_SUCCESS) && (cfg.loop!=0) && (get_sig_received()==0)) {
		if (run_tunnel(&cfg) != 0)
			ret = EXIT_FAILURE;
		else:
			sleep(cfg.loop);
	}
```
or:
```
	while ((cfg.loop!=0) && (get_sig_received()==0)) {
		if (run_tunnel(&cfg) != 0) {
			ret = EXIT_FAILURE;
			break;
		}
		sleep(cfg.loop);
	}
```

src/main.c
mrbaseman: very large values don't make much sense here. The value is used as argument for `sleep(unsigned int)`, so even in this call, negative values are (probably) cast to large positive numbers. The worst that can happen is that _openfortivpn_ waits a very long time before trying to reconnect, which gives a user experience nearly as if loop was left with the default value of 0. 

I don't feel that a code change is needed here, but I'm open for suggestions. Maybe some comments of [this discussion](https://stackoverflow.com/questions/7021725/converting-string-to-integer-c) are helpful

src/main.c
mrbaseman: yes, I was looking for something like this, but first didn't get the signal handling right and after I had introduced `get_sig_received()` I forgot to come back to this loop condition. Something with `do { ... } while ( )` might be another solution that doesn't need a `break;`

mrbaseman:
A general remark: Supporting automatic reconnect was an idea which came up in one of the issues about unstable connections. I think they can mostly be ascribed MTU problems or other problems in the lower layers. Connections that are dropped after hours most probably hit the idle timeout configured on the vpn gateway. Does it still make sense to implement automatic reconnects from the client side?

DimitriPapadopoulos:
Right, `do { ... } while ( )` might be better.

As for your general remark, I'm not so fond of automatically reconnecting either. Sure, it might make sense when you hit the gateway timeout - which I believe is ~8 hours by default **whether the connection is idle or not**. On the other hand many FortiGate devices are configured to accept only tokens / one-time passwords. Reconnecting doesn't make sense in such cases.

Whether the feature is useful or not, I think I'd rather keep it outside _openfortivpn_ and write a simple shell wrapper that restarts _openfortivpn_, or use a crontab, or use the `respawn` facility of SysV init / _upstart_ / _systemd_. To that purpose:
* We should verify that _openfortivpn_ exit values are meaningful. But then I don't think it's possible to discriminate between a FortiGate timeout and MTU issues for example.
* We could start a wiki and provide a sample script, or describe how to use the _init_ system and `respawn` for that purpose.

src/main.c
DimitriPapadopoulos: That's not an important issue. Let's reconsider only if it raises a Coverity issue.

src/main.c
DimitriPapadopoulos: Right, `do { ... } while ( )` might be better. But don't we need a break anyway to avoid the last `sleep()` call which should occur after the `run_tunnel()` call?

mrbaseman: I have just updated the branch with https://github.com/adrienverge/openfortivpn/pull/190/commits/d489dd7d22eb3821c5dc283afd20529fe7385325 which addresses this (not sure if my `if`s are nicer than your `break`s though...)

mrbaseman:
A wiki that provides some extra hints and helper scripts is a good idea. I have seen the request for init scripts or ipup scripts for pppd several times already. But I'm using two factor authentication with Tokens as well, so I can't run it purely script based either...

mrbaseman:
the reauthentication feature was requested and successfully tested in #92.  @DimitriPapadopoulos I have just seen that I hadn't yet properly addressed the cast of a long to an unsigned int. Now I have done so and also a bit of cleanup. Are you ok with merging this small patch into master (after rebasing and squashing it into a single commit)?

DimitriPapadopoulos:
Looks good to me.

To tell the truth I would prefer to keep the reconnection mechanism outside of _openfortivpn_ (a shell wrapper ?) but if there is a widespread need it's OK with me - I am of course biased by my own experience and needs.

mrbaseman:
I think we have a couple of requests for this feature. Only, I'm not so happy with calling the command line option "--loop". Maybe there are better proposals? If not, I would rebase and merge this in the near future. (I was ill and on vacation and had a considerable backlog in my mailbox, now I'm ill again, so my activities here are a bit slowed down... )

adrienverge:
Like Dimitri I'd rather use a wrapper to relaunch openfortivpn, but since there are multiple demands for this, let's do it!

> Only, I'm not so happy with calling the command line option "--loop". Maybe there are better proposals?

Maybe `--auto-reconnect`?

DimitriPapadopoulos:
Perhaps [`--persistent`](http://www.thesaurus.com/browse/persistent) as suggested in #251? Here are a few links to other synonyms:
* [loop](http://www.thesaurus.com/browse/loop)
* [never-ending](http://www.thesaurus.com/browse/never-ending)
* [continuous](http://www.thesaurus.com/browse/continuous)
* [nonstop](http://www.thesaurus.com/browse/nonstop)

Most probably we should reuse the vocabulary used in _systemd_ and other _init_ systems. See for example [How To Configure a Linux Service to Start Automatically After a Crash or Reboot](https://www.digitalocean.com/community/tutorials/how-to-configure-a-linux-service-to-start-automatically-after-a-crash-or-reboot-part-1-practical-examples), [Autostart using systemd](https://home-assistant.io/docs/autostart/systemd/) or [Ensure systemd services restart on failure](https://jonarcher.info/2015/08/ensure-systemd-services-restart-on-failure/):
* respawn
* restart
* reload
* auto...

adrienverge:
> Perhaps --persistent as suggested in #251?

Sounds good too :+1:

adrienverge:
Nice job!

Should I release v1.7.0? Or do you guys plan to merge other features soon?

mrbaseman:
I would wait with the new release for the 2fa issue in #255 to land, and maybe #254 as well. Both are not very big changes, but at least the 2fa in combination with network manager applet is an important one.

Issue #150, opened 2017-07-04
[Feature request] support proxy


ymartineau:
hi,

do you already support corporate proxy (standard, no SOCKS, without authentication) or do you plan to add that kind of feature?

thank you

mrbaseman:
currently, there is no built-in proxy support. The simple approach without authentication should not be too hard to implement. Something like [described here](https://stackoverflow.com/questions/3127127/how-to-open-socket-thru-proxy-server-in-net-c/3127176#3127176) needs to be implemented in the tcp_connect function in tunnel.c (just linking this here as a reminder for myself or anyone else who likes to create a pull request)

mrbaseman:
could you test [this implementation](https://github.com/mrbaseman/openfortivpn/tree/proxy_support) please? When connecting to the vpn server it first looks at the environment variable `PROXY` which should contain `proxy_host:proxy_port`. If this variable is set, _openfortivpn_ tries to connect to this proxy and tell it to forward the traffic to the vpn_host and vpn_port specified on the command line or in the config file of _openfortivpn_. Documentation is still to be done.

mrbaseman:
@ymartineau did you have a chance to test the proposed implementation for proxy support?

ymartineau:
hi, I didn't had the time yet. I work from home today and vpn over vpn is not supported in our network, I'll try next week. Thank you

ymartineau:
I would say it doesn't work:

I exported the PROXY environment variable. I checked it before with http_proxy, https_proxy and a wget.

Then, I ran openfortivpn and had a connection timeout.

[root@localhost bin]# ./openfortivpn -v --insecure-ssl
DEBUG:  Loaded config file "/home/centos/programs/openfortivpn-local/etc/openfortivpn/config".
DEBUG:  Config host = "1.2.3.4"
DEBUG:  Config realm = ""
DEBUG:  Config port = "5678"
DEBUG:  Config username = "johndoe"
DEBUG:  Config password = "********"
ERROR:  connect: Connection timed out
INFO:   Closed connection to gateway.
ERROR:  connect: Network is unreachable
INFO:   Could not log out.
[root@localhost bin]# 

I ran a tcpdump and saw that connection was still done directly to the vpn, not through the proxy. I did not check your code.

mrbaseman:
thanks for testing. Obviously I have to correct my code. I thought the socket to the proxy was already open and the connect to the gateway is forwarded through the proxy, but it seems not to be the case.
[...]

PR #192, opened 2017-10-12
Proxy support

mrbaseman:
this branch implements the support proxy as requested in #150

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -323,24 +323,122 @@ static int get_gateway_host_ip(struct tunnel *tunnel)
  */
 static int tcp_connect(struct tunnel *tunnel)
 {
-	int ret, handle;
+	int ret, handle, bytes_read, curr_pos;
 	struct sockaddr_in server;
+	const struct addrinfo hints = { .ai_family = AF_INET };
+	struct addrinfo *result = NULL;
+	char *env_proxy, *proxy_host, *proxy_port, *response;
+	char request[128];
 
 	handle = socket(AF_INET, SOCK_STREAM, 0);
 	if (handle == -1) {
 		log_error("socket: %s\n", strerror(errno));
 		goto err_socket;
 	}
+	env_proxy = getenv("https_proxy");
+	if (env_proxy == NULL)
+		env_proxy = getenv("HTTPS_PROXY");
+	if (env_proxy == NULL)
+		env_proxy = getenv("all_proxy");
+	if (env_proxy == NULL)
+		env_proxy = getenv("ALL_PROXY");
+	if (env_proxy != NULL) {
+		// protect the original environment from modifications
+		env_proxy = strdup(env_proxy);
+		// get rid of a trailing slash
+		if (env_proxy[strlen(env_proxy) - 1] == '/')
+			env_proxy[strlen(env_proxy) - 1] = '\0';
+		// get rid of a http(s):// prefix in env_proxy
+		proxy_host = strstr(env_proxy, "://");
+		if (proxy_host == NULL)
+			proxy_host = env_proxy;
+		else
+			proxy_host += 3;
+		// split host and port
+		proxy_port = index(proxy_host, ':');
+		if (proxy_port != NULL) {
+			proxy_port[0] = '\0';
+			proxy_port++;
+			server.sin_port = htons(strtol(proxy_port, NULL, 10));
+		} else {
+			server.sin_port = htons(tunnel->config->gateway_port);
+		}
+		// get rid of a trailing slash
+		if (proxy_host[strlen(proxy_host) - 1] == '/')
+			proxy_host[strlen(proxy_host) - 1] = '\0';
+		log_debug("proxy_host: %s\n", proxy_host);
+		log_debug("proxy_port: %s\n", proxy_port);
+		server.sin_addr.s_addr = inet_addr(proxy_host);
+		// if host is given as fqhn we have to do a dns lookup
+		if (server.sin_addr.s_addr == INADDR_NONE) {
+			ret = getaddrinfo(proxy_host, NULL, &hints, &result);
 
+			if (ret) {
+				if (ret == EAI_SYSTEM)
+					log_error("getaddrinfo: %s\n", strerror(errno));
+				else
+					log_error("getaddrinfo: %s\n", gai_strerror(ret));
+				goto err_connect;
+			}
+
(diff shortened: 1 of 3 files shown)


DimitriPapadopoulos:
Can you fix the Codacy warning?

DimitriPapadopoulos:
@mrbaseman Part of the patch is not strictly following the [Linux kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html), typically:
> ### 3) Placing Braces and Spaces
> [...]
> ### 3.1) Spaces
> [...]
> 
> Use one space around (on each side of) most binary and ternary operators, such as any of these:
> `=  +  -  <  >  *  /  %  |  &  ^  <=  >=  ==  !=  ?  :`

I don't want to be pedantic though - only use this comment if you feel it it useful. Alternatively commit it as is and, if you agree, I can change the code later, before a release for example.

DimitriPapadopoulos:
By the way there's a [_checkpatch_ Perl script](https://wintergreenworks.wordpress.com/2012/05/30/checkpatch-linux-coding-style-check/) to check patches and files in the Linux kernel tree.

adrienverge:
@DimitriPapadopoulos is right, let's try to use the same coding rules consistently across the code. For example I find `bytes_read=read(handle,&(request[curr_pos++]),2);` not very readable.

> By the way there's a checkpatch Perl script to check patches and files in the Linux kernel tree.

I didn't know this script. Maybe it could be used in our CI!

DimitriPapadopoulos:
@adrienverge Unfortunately I find this script way too verbose. I had to leave out lots of errors, some of which I'm really not interested in.

In any case:
* We need to fix the existing code first.
* We might want to disregard out a few errors types.

adrienverge:
@DimitriPapadopoulos Sure, no problem.

mrbaseman:
our `lint` check was already happy with the current patch, but `checkpatch.pl` finds a couple more problems. I'll correct them.

src/tunnel.c
```diff
@@ -322,9 +326,22 @@ static int tcp_connect(struct tunnel *tunnel)
...
 		}
+		// get rid of a trailing slash
+		if (proxy_host[strlen(proxy_host) - 1] == '/')
+			proxy_host[strlen(proxy_host) - 1] = '\0';
 		log_debug("proxy_host: %s\n", proxy_host);
 		log_debug("proxy_port: %s\n", proxy_port);
 		server.sin_addr.s_addr = inet_addr(proxy_host);
+		// if host is given as fqhn we have to do a dns lookup
+		if (server.sin_addr.s_addr == INADDR_NONE) {
+			host = gethostbyname(proxy_host);
```
DimitriPapadopoulos: While _gethostname()_ is perfect right now, please note for future reference that pull request #77 suggests switching to _getaddrinfo()_. It makes sense as [_gethostname()_ is obsolete and not part of POSIX anymore](https://github.com/adrienverge/openfortivpn/pull/77#issuecomment-335367313).

mrbaseman:
Yes, I'm aware of that, but on line [271](https://github.com/adrienverge/openfortivpn/blob/6a32acd5b62fe6c8fb832212dd48a59c27967851/src/tunnel.c#L271) of tunnel.c `gethostbyname` is used too. When adding ipv6 support in the future we have to change both places.

mrbaseman:
I'll check how #77 needs to be updated when #192 is landed and push an appropriate commit to that branch (or vice versa, depending on which pr gets merged first).

DimitriPapadopoulos:
@mrbaseman Looks good to me. You just need to update the man page after commit 28cd7f0 _support dns lookup for proxy and tolerate trailing slashes_:
```
It checks if one of the environment variables
B https_proxy HTTPS_PROXY all_proxy ALL_PROXY
is set which are supposed to contain a string of the format 
.br
.B http://[ip]:[port]
.br
where
.B [ip]
is the ip of the proxy server in dotted decimal notation and
.B [port]
```
Indeed one may now use a DNS name instead of an IPv4 address.

DimitriPapadopoulos:
Also in **tunnel.c** and unless I'm missing something you can change:
```
		proxy_str = strdup(env_proxy);
		env_proxy = proxy_str;
```
into:
```
		env_proxy = strdup(env_proxy);
```

DimitriPapadopoulos:
And finally according to [The Open Group Base Specifications Issue 6
IEEE Std 1003.1, 2004 Edition](http://pubs.opengroup.org/onlinepubs/009695399/functions/bzero.html) you should use _memset_() instead of _bzero_():
> ### APPLICATION USAGE
> 
> The [_memset_()](http://pubs.opengroup.org/onlinepubs/009695399/functions/memset.html) function is preferred over this function.
> 
> For maximum portability, it is recommended to replace the function call to _bzero_() as follows:
> 
> ```
> #define bzero(b,len) (memset((b), '\0', (len)), (void) 0)
> ```
> 
> ### FUTURE DIRECTIONS
> 
> This function may be withdrawn in a future version.
> 
> ### CHANGE HISTORY
> 
> First released in Issue 4, Version 2.
> 
> #### Issue 5
> Moved from X/OPEN UNIX extension to BASE.
> 
> #### Issue 6
> 
> This function is marked LEGACY.

I cannot find _bzero_() in [The Open Group Base Specifications Issue 7
IEEE Std 1003.1-2008, 2016 Edition](http://pubs.opengroup.org/onlinepubs/9699919799/).

DimitriPapadopoulos:
Also you've replaced `gethostbyname` with `getaddrinfo` in `AC_CHECK_FUNCS` (I had forgotten that) but you can also add `ntohs` next to `inet_ntoa`.

mrbaseman:
I have squashed and rebased this branch and updated the commit message now. @DimitriPapadopoulos and @adrienverge do you think we can merge it like this?

DimitriPapadopoulos:
Looks good to me.

Also proxy support seems complicated to me. Am I correct that there are different proxies out there with variations in the protocol? In any case, since it already works for two users, I suggest you merge this pull request so that it's available out there in the field and can be tested by more users.

adrienverge:
If it's good to both of you then it's good to me!

mrbaseman:
@DimitriPapadopoulos well, there are standard http proxies, proxies that require authentication in order to use them, socks proxies... This one is the simplest implementation (without authentication), but with a little bit of tolerance about line endings and http headers, where the proxy might not exactly follow the standard

Issue #130, opened 2017-05-26
default config is overwritten by each install


mrbaseman:
There is a sample config file in the package and `make install` copies that file over an existing one.
Actually, this is what `make install` is supposed to do: replace all files of previous installations by up-to-date ones. For the config file this might not be what people want.
There are workarounds:
- not using the default config file for any useful connection data, but copy it and invoke openfortivpn with  `--config` 
- keep a backup of the default config file and put that one in place after each installation of a new version of openfortivpn

When searching for a solution with autoconf I have found the recommendation **not** to install config files directly, but install templates of them to some place like /usr/share. An additional goody could be an installation hook which checks if the default config file exists, and if not it would copy the template. If the default config file already exists it would just skip the copy operation. We already have such a hook to adjust the file permissions.

PR #193, opened 2017-10-13
preserve existing config during install, this solves #130

mrbaseman:
I have decided to keep the sample config file in the etc directory, but append a ".sample" extension and copy it to the right location if it doesn't exist yet. I think this is less effort than maintaining a separate path, e.g. $prefix/share/openfortivpn or so just for a single template file which contains only a few lines


DimitriPapadopoulos:
This issue has been bothering me as well. Thank you for looking into it!

I'm not sure I understand the logic though. Can you walk me through it? My understanding is:
```
	install etc/openfortivpn/config.sample

	if $(sysconfdir)/openfortivpn/config does not exist
	then
		copy config.sample in config (⇒ both "config.sample" and "config" exist)
	else
		delete config.sample (⇒ both "config.sample" and "config" are deleted)
	endif
```
Is that correct?

mrbaseman:
oh, yes, your understanding is **nearly** correct. The current commit works like this:
`       install etc/openfortivpn/config.sample

	if $(sysconfdir)/openfortivpn/config does not exist
	then
		copy config.sample in config (⇒ both "config.sample" and "config" exist)
	else
		delete config.sample (⇒ "config.sample" is deleted, but "config" is preserved with its previous content)
	endif`
So, I should
- either delete config.sample also when config did **not** exist in the first place (I simply should have put the `rm` after the `endif` instead into the `else` branch)
- or keep config.sample in **both** cases (so the user can have a look which entries are supposed to be in config). This option makes also sense to me (so we would just drop the `rm`)

which option would you prefer?

DimitriPapadopoulos:
Yes, looks good:
* Always copy `config.sample`.
* Copy `config` only when it does not already exist (do not overwrite).

By the way, it looks like files such as `config.sample` often live in `/usr/share/doc` - and are sometimes named `config.template`. I can find no clear rule though. Please do as you see fit.

mrbaseman:
I have the impression that `/usr/share/doc` is more a place for documentation which is not in man or info format. To me `/usr/share/"package"` seems to be more appropriate. I had a look at my Ubuntu installation: Some packages create even another a subdirectory "examples" or "templates" and remove the suffix. Some have a suffix ".template", some others use ".sample". 
I guess the data structure below /usr/share/ depends on many factors (history of the package, number and type of files, personal taste of the authors,...). For _openfortivpn_ I think with currently only one file the best is to put it directly to `$(prefix)/share/openfortivpn`, and you are right, there are more ".template" files below /usr than ".sample" files, so I have renamed the file in another commit.

DimitriPapadopoulos:
Looks perfect. Thank you for looking into this.

mrbaseman:
Thank you for your input with your comments. Four eyes always see more than two.

mrbaseman:
I have merged this now and rebased my other branches. I hope that I didn't mess up the github repo ;-)

adrienverge:
> I have decided to keep the sample config file in the etc directory, but append a ".sample" extension and copy it to the right location if it doesn't exist yet

Thanks, good idea!

PR #196, opened 2017-10-16
Update README.md

mrbaseman:
add openSUSE / SLE to the list of distributions that ship openfortivpn, and update the solus link

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,7 @@ Examples
 Installing
 ----------
 
-openfortivpn is packaged for [Fedora](https://admin.fedoraproject.org/pkgdb/package/rpms/openfortivpn/), [openSUSE / SLE](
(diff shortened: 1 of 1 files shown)


mrbaseman:
I have just noticed that even SUSE provides an rpm for openfortivpn (and while I'm at it, I would like to point the solus link to the openfortivpn package instead of the general solus web site)

adrienverge:
Even better!

DimitriPapadopoulos:
Perhaps the Suse link should be:
https://software.opensuse.org/package/openfortivpn
instead of this temporary link (depends on user _Tomcat42_ as long as the package remains "unstable"):
https://software.opensuse.org/download.html?project=home%3ATomcat42&package=openfortivpn

mrbaseman:
even better, the link that @DimitriPapadopoulos has found is much nicer than the one that I have found

PR #197, opened 2017-10-16
Is the pppd path valid?

DimitriPapadopoulos:
If file "/usr/sbin/pppd" does not exist, print a specific message.
It happens often enough that PPP is not installed, we need something
specific for that.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -93,7 +93,16 @@ static int pppd_run(struct tunnel *tunnel)
 	int amaster;
 #ifndef __APPLE__
 	struct termios termp;
+#endif
 
+	static const char pppd_path[] = "/usr/sbin/pppd";
+
+	if (access(pppd_path, F_OK) != 0) {
+		log_error("%s: %s.\n", pppd_path, strerror(errno));
+		return 1;
+	}
+
+#ifndef __APPLE__
 	termp.c_cflag = B9600;
 	termp.c_cc[VTIME] = 0;
 	termp.c_cc[VMIN] = 1;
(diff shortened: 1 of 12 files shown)


src/tunnel.c
```diff
@@ -93,7 +93,16 @@ static int pppd_run(struct tunnel *tunnel)
 	int amaster;
 #ifndef __APPLE__
 	struct termios termp;
+#endif
+
+	const char pppd_path[] = "/usr/sbin/pppd";
+
+	if (access(pppd_path, F_OK) != 0) {
```
adrienverge: Maybe `X_OK` is more appropriated?

DimitriPapadopoulos:
A possible problem with `X_OK` is:
> The check is done using the calling process's real UID and GID, rather than the effective IDs as is done when actually attempting an operation (e.g., open(2)) on the file. Similarly, for the root user, the check uses the set of permitted capabilities rather than the set of effective capabilities; and for non-root users, the check uses an empty set of capabilities.

Not certain what it means at this point, given that:
```
$ ls -l /usr/sbin/pppd
-rwsr-xr-- 1 root dip 390888 Jan 29  2016 /usr/sbin/pppd
$ 
```
Any clue?

adrienverge:
On my Fedora system, `pppd` is executable by anybody:
```
ls -l /usr/sbin/pppd
-rwxr-xr-x 1 root root 392736 11 févr.  2017 /usr/sbin/pppd
```

But you're right, let's avoid problems on system with exotic permissions. `F_OK` is fine!

mrbaseman:
this works (if I move my pppd elsewhere, I get the clear error message), but...
`src/tunnel.c: In function ‘pppd_run’:
src/tunnel.c:120:4: warning: initialization discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]`
`    pppd_path, "38400", "noipdefault", "noaccomp",`
`    ^`

DimitriPapadopoulos:
Indeed. This is a C feature. While _execvp_() will not modify the strings in the `argv` array, you cannot express that in C. So we have to cast from `const char*` to `char *` before calling _execvp_().

See for example:
* [comp.lang.c FAQ](http://c-faq.com/ansi/constmismatch.html]
* [Why is argv parameter to execvp not const?](https://stackoverflow.com/questions/10456043/why-is-argv-parameter-to-execvp-not-const)
* [Creating a char* const * for execvp()](https://stackoverflow.com/questions/35386791/creating-a-char-const-for-execvp)

DimitriPapadopoulos:
Had to delete/recreate the DimitriPapadopoulos:missing_pppd branch...

DimitriPapadopoulos:
Had to delete/recreate the DimitriPapadopoulos:missing_pppd branch...

PR #198, opened 2017-10-16
Linux kernel coding style

DimitriPapadopoulos:
Use checkpatch.pl to identify and fix a few issues.
For now I'm leaving lots of issues as is because I feel
it would modify existing code way too much without clear
benefit.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -44,7 +44,7 @@ static void url_encode(char *dest, const char *str)
 		else if (*str == ' ')
 			*dest++ = '+';
 		else {
-			static char hex[] = "0123456789abcdef";
+			static const char hex[] = "0123456789abcdef";
 
 			*dest++ = '%';
 			*dest++ = hex[*str >> 4];
@@ -337,8 +337,8 @@ int try_otp_auth(struct tunnel *tunnel, const char *buffer, char **res)
 {
 	char data[256];
 	char path[40];
-	char tmp [40];
-	char prompt [80];
+	char tmp[40];
+	char prompt[80];
 	const char *t = NULL, *n = NULL, *v = NULL, *e = NULL;
 	const char *s = buffer;
 	char *d = data;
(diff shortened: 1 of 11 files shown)


src/http.c
```diff
@@ -156,7 +156,8 @@ int http_receive(struct tunnel *tunnel, char **response)
 
 			if (header_size) {
 				/* We saw the whole header, let's check if the
-				 * body is done as well */
+				 * body is done as well
+				 */
```
adrienverge: Don't fully agree with this one, but why not.

DimitriPapadopoulos: I have to say I don't agree either.
We can leave it out if you want to.

adrienverge: In this case, let's leave it like it was!

mrbaseman:
I personally prefer to put more braces than strictly needed. 
If you later want to update the `if` or `else` branch, you can not forget to put braces around when they are already there.
I agree that the rules for spaces make the code better readable, but I would love a tool that corrects my imperfectly formatted code ;-)
@DimitriPapadopoulos did you create this with the `--fix` option or by correcting each item by hand?

DimitriPapadopoulos:
@mrbaseman I was horrified by the number of issues reported by _checkpatch.pl_. So no, I didn't use the `--fix` option. I fixed manually the errors and warnings that made most sense to me and skipped suggestions that would change the existing code base too much without clear benefit.

DimitriPapadopoulos:
@mrbaseman I do understand you point, it is valid of course. On the other hand I'd rather use an existing style - with its advantages and pitfalls. The [Linux kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html#placing-braces-and-spaces) is specific about this:
> Do not unnecessarily use braces where a single statement will do.
> ```
> if (condition)
> 	action();
> ``` 
> and
> ```
> if (condition)
> 	do_this();
> else
> 	do_that();
> ```
> This does not apply if only one branch of a conditional statement is a single statement; in the latter case use braces in both branches:
> ```
> if (condition) {
> 	do_this();
> 	do_that();
> } else {
> 	otherwise();
> }
This is just a suggestion. If you really want to we could use braces even with single statements. However the current code is a mixture of single statements with/without braces.

mrbaseman:
This really makes sense. Thanks for explaining and for patience. When I read the coding style, most of the rules feel clear and logic, but when writing and debugging code I must notice that I haven't assimilated them yet enough.

PR #199, opened 2017-10-17
Does /usr/sbin/pppd exist?

DimitriPapadopoulos:
If /usr/sbin/pppd does not exist, print a specific message. When users
build from source, it happens often enough that PPP is not installed.

The existing frprintf(stderr, ...) call just after execvp() currently
doesn't work, probably because of the prior forkpty() call. If we get
it to work someday, we can move this after the failed call to execvp().
On the other hand a very specific message in case /usr/sbin/pppd is
missing makes sense, as this occurs quite often.

Because of limitations of C const-correctness, the execpv signature is:
	char *const argv[]
but execvp() really doesn't change its arguments. Therefore it is valid
to pass an array of:
	const char *const
then cast the array back to:
	char *const *
just before calling execvp().

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -93,7 +93,16 @@ static int pppd_run(struct tunnel *tunnel)
 	int amaster;
 #ifndef __APPLE__
 	struct termios termp;
+#endif
 
+	static const char pppd_path[] = "/usr/sbin/pppd";
+
+	if (access(pppd_path, F_OK) != 0) {
+		log_error("%s: %s.\n", pppd_path, strerror(errno));
+		return 1;
+	}
+
+#ifndef __APPLE__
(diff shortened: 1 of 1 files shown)


mrbaseman:
looks good to me. for the `fprintf(stderr, ...` you have proposed an alternative in https://github.com/adrienverge/openfortivpn/pull/152#issuecomment-318321587 
Now that we have noticed that the `fprintf`doesn't work, shall we try the rewrite of `do_log`, maybe in a separate pr?

DimitriPapadopoulos:
Yes, it's worth trying the alternative. However I'm not certain the alternative works either - because of the prior _forkpty()_ call.

In any case it's worth trying to do something about it. I'll open a new ticket.

PR #200, opened 2017-10-18
Does /usr/sbin/pppd exist?

DimitriPapadopoulos:
If /usr/sbin/pppd does not exist, print a specific message. When users
build from source, it happens often enough that PPP is not installed.

The existing frprintf(stderr, ...) call just after execvp() currently
doesn't work, probably because of the prior forkpty() call. If we get
it to work someday, we can move this after the failed call to execvp().
On the other hand a very specific message in case /usr/sbin/pppd is
missing makes sense, as this occurs quite often.

Because of limitations of C const-correctness, the execpv signature is:
	char *const argv[]
but execvp() really doesn't change its arguments. Therefore it is valid
to pass an array of:
	const char *const
then cast the array back to:
	char *const *
just before calling execvp().

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -93,7 +93,16 @@ static int pppd_run(struct tunnel *tunnel)
 	int amaster;
 #ifndef __APPLE__
 	struct termios termp;
+#endif
 
+	static const char pppd_path[] = "/usr/sbin/pppd";
+
+	if (access(pppd_path, F_OK) != 0) {
+		log_error("%s: %s.\n", pppd_path, strerror(errno));
+		return 1;
+	}
+
+#ifndef __APPLE__
 	termp.c_cflag = B9600;
 	termp.c_cc[VTIME] = 0;
 	termp.c_cc[VMIN] = 1;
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Something's wrong! Can't merge this pull request:
* #199: Pull request appears to have worked but GitHub shows my `missing_pppd` branch as not integrated.
* #197: I just messed it up and was unable to rebuild a clean branch.

mrbaseman:
- #197 by mistake contained https://github.com/adrienverge/openfortivpn/pull/197/commits/e8f055ced193865d73b183b6e0a4cb5a94003a18 which looks like applying the linux coding style (meant to be in #198), then https://github.com/adrienverge/openfortivpn/pull/197/commits/bf8027d6ed5e8afadf2c8c90fc83cc19ac4c003d couldn't be applied anymore.
- with #199 I don't see any clear reason for the failure. Your Logo is not displayed, and in the raw patch it is From: Dimitri Papadopoulos <DimitriPapadopoulos(at)noreply.users.github.com> whereas older patches are From: Dimitri Papadopoulos Orfanos <dimitri.papadopoulos(at)gmail.com>
- recently I had to clean up my branches, too. My mistake was that I did a `git merge upstream/master`  which caused a subsequent `git rebase upstream/master` to fail. Possible solution: `git rebase -i $(git merge-base master upstream/master)` and drop the duplicate commits, or (if no uncommitted changes are there) start over again with `git pull -f` (I don't know if it was a similar issue in your case, which has lead to this strange behavior)

PR #202, opened 2017-10-19
Free cert_whitelist

DimitriPapadopoulos:


src/config.h
```diff
@@ -114,13 +114,14 @@ struct vpn_config {
 			free((cfg)->cert_whitelist); \
 			(cfg)->cert_whitelist = tmp; \
 		} \
```
mrbaseman: `cert_whitelist` is freed in the loop above. After the loop `cert_whitelist` is `NULL` so that the additional `free` has no action

DimitriPapadopoulos: My wrong indeed. I'm sorry. Will fix that.

DimitriPapadopoulos: I really should had waited for a review of the pull request.

DimitriPapadopoulos: Actually I'm at a loss how to update branch `jisaitua:pppd-ifname-option` in pull request #165 to be able to merge it. I've been trying different things without success. Git messages don't make any sense. Got tired of it and convinced I need an additional commit in `master` - but it doesn't work. Any clue how to update this pull request?

PR #204, opened 2017-10-19
Define and use the ARRAY_SIZE macro

DimitriPapadopoulos:
Borrowed from the Linux kernel (include/linux/kernel.h).
ARRAY_SIZE(x) returns the number of elements in an array.


mrbaseman:
looks good to me

PR #209, opened 2017-10-20
Reorganize pppd options for clarity

DimitriPapadopoulos:


src/tunnel.c
```diff
@@ -118,21 +118,35 @@ static int pppd_run(struct tunnel *tunnel)
...
+
 		// Dynamically get first NULL pointer so that changes of
 		// args above don't need code changes here
 		int i = ARRAY_SIZE(args) - 1;
 		for (; args[i] == NULL; i--)
 			;
 		i++;
 
+		// Coverity detected a defect, actually a false positive:
+		//  CID 196857: Out-of-bounds write (OVERRUN)
```
mrbaseman: the reason is probably that coverity doesn't recognize that at least one `NULL` entry will remain in `args[]`, so that the `for` loop above will be executed at least once. 
BTW: `while(args[i] == NULL) i--;` would perhaps be easier to read than the `for` loop

mrbaseman:
I would add the explanation for the false positive also in the comment, and maybe replace the `for` loop by a `while` (see my comment in the code). Apart from that it looks good.

src/tunnel.c
DimitriPapadopoulos: Indeed `args` is not constant and Coverity is unable to infer that `args` *will* still contain NULL elements.

Good idea about the `while`, will do.

PR #210, opened 2017-10-20
Better error handling of pppd exit status

DimitriPapadopoulos:
Use log_error() instead of log_debug() in case of pppd failure.
In case of abnormal exit, print the termination signal.


mrbaseman:
looks good to me. 
For connecting stderr of the child to the parent process, I have found a suggestion with pipes: https://jineshkj.wordpress.com/2006/12/22/how-to-capture-stdin-stdout-and-stderr-of-child-program/ but it seems one has to copy the data explicitly https://stackoverflow.com/questions/2500344/forkpty-socket

DimitriPapadopoulos:
OK, then I suggest I commit this change, then create a separate new issue to better handle `stderr` in the future.

PR #211, opened 2017-10-20
execvp() -> execv()

DimitriPapadopoulos:
We provide the full path to /usr/sbin/pppd. No need to search the PATH
with execvp(), execv() is enough.


mrbaseman:
You are right. A nice explanation about the exec functions family, which I actually found better than the man page: https://stackoverflow.com/questions/5769734/what-are-the-different-versions-of-exec-used-for-in-c

Issue #201, opened 2017-10-18
print error status of execv() and /usr/sbin/pppd


DimitriPapadopoulos:
It looks like `fprintf(stderr, ...)` doesn't print in this piece of code of _tunnel.c_:
```
		execv(args[0], args);
		fprintf(stderr, "execv: %s\n", strerror(errno));
		exit(EXIT_FAILURE);
```

See issue #194 where no error message is printed despite the missing `/usr/sbin/pppd` executable.

An additional error message has been added in pull request #199 before a `forkpty()` call, with the rationale that the `forkpty()` might be responsible for the `fprintf(stderr, ...)` failure.

Now if we find a solution to print an error message just after the `execv()` call, we might revert #199.

For the record, this piece of code originates in pull request #152  where I had added a [comment](https://github.com/adrienverge/openfortivpn/pull/152#issuecomment-318321587) suggesting an alternative implementation:
> An alternative would be to rewrite _do_log()_ as follows and use the mutex-less _unprotected_do_log()_ after _execvp()_:
> ```
> void do_log(int verbosity, const char *format, ...)
> {
> 	va_list args;
> 
> 	pthread_mutex_lock(&mutex);
> 
> 	unprotected_do_log(verbosity, format, args);
> 
> 	pthread_mutex_unlock(&mutex);
> }
> ```
I doubt this alternative implementation helps - as I see it the real problem is the `forkpty()` call. It's nevertheless a good idea to implement it with a different goal: consistency in error output.

DimitriPapadopoulos:
This is clearly an effect of `forkpty()`.

The child process operates in a pseudoterminal. If there's an error in the child process - for example when executing a program with execv() - I cannot find a way to access / print to the initial pre-fork stderr.

DimitriPapadopoulos:
#### Extended goals

Make sure error messages are logged and printed to `stderr` in both these cases:
1. `execv()` fails to execute `/usr/sbin/pppd` (for example because PPPD is not installed on the system),
2. `/usr/sbin/pppd` is executed but fails (for example because option `ifname` is not implemented in the available version of PPPD): we already retrieve the exit status code and print error messages (#189) but these messages are not always clear, so it would be better to intercept the `stderr` output of `/usr/sbin/pppd` in the parent process and reprint (perhaps after filtering) to the parent process' `stderr`.

#### How to achieve that?

See #210.

DimitriPapadopoulos:
[...]

PR #212, opened 2017-10-21
F_OK -> X_OK

DimitriPapadopoulos:
We need to catch as many errors as possible early,
before calling forkpty(). After the call, messages
printed to stderr will be redirected to a pseudo-
terminal and lost.

At the same time discard existing attempts to print
to stderr after calling forkpty().

Not sure I understand the ACCESS(2) man page entirely:
	The check is done using the calling process's
	real UID and GID, rather than the effective IDs
	as is done when actually attempting an operation
	(e.g., open(2)) on the file. Similarly, for the
	root user, the check uses the set of permitted
	capabilities rather than the set of effective
	capabilities; and for non-root users, the check
	uses an empty set of capabilities.
 	This allows set-user-ID programs and capability-
	endowed programs to easily determine the invoking
	user's authority. In other words, access() does
	not answer the "can I read/write/execute this
	file?" question. It answers a slightly  differ‐
	ent  question: "(assuming  I'm  a setuid binary)
	can the user who invoked me read/write/execute
	this file?", which gives set-user-ID programs
	the possibility to prevent malicious users from
	causing them to read files which users shouldn't
	be able to read.
	If the calling process is privileged (i.e., its
	real UID is zero), then an X_OK check is successful
	for a regular file if execute permission is enabled
	for any of the file owner, group, or other.

On the other hand I have checked that X_OK works on Ubuntu
16.04 where:
$ /bin/ls -l /usr/sbin/pppd
-rwsr-xr-- 1 root dip 390888 janv. 29  2016 /usr/sbin/pppd
$

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -41,7 +41,6 @@
 #include <util.h>
 #endif
 #include <sys/wait.h>
-#include <assert.h>
 
 #include "http.h"
 #include "log.h"
@@ -100,8 +99,13 @@ static int pppd_run(struct tunnel *tunnel)
 
 	static const char pppd_path[] = "/usr/sbin/pppd";
 
-	if (access(pppd_path, F_OK) != 0) {
-		log_error("%s: %s.\n", pppd_path, strerror(errno));
+	/*
+	 * Catch as many errors as possible and print error messages before
+	 * invoking forkpty(). After the call stderr will be directed to a
+	 * pseudoterminal.
+	 */
+	if (access(pppd_path, X_OK) != 0) {
+		log_error("%s: %s\n", pppd_path, strerror(errno));
 		return 1;
 	}
 
@@ -175,8 +179,6 @@ static int pppd_run(struct tunnel *tunnel)
 			args[i++] = "ifname";
 			args[i++] = tunnel->config->pppd_ifname;
 		}
-		// Assert that we didn't use up all NULL pointers above
-		assert(i < ARRAY_SIZE(args));
 
 		close(tunnel->ssl_socket);
 		execv(args[0], (char *const *)args);


mrbaseman:
I think the main point is that _openfortivpn_ **is** a privileged process (you have started it with root privileges, right?) - that's why it works. 
The interesting question would be if there are cases in which X_OK would return false even if `pppd` was executable for _openfortivpn_. Without having tested it, I would assume that if you `chmod u+s _openfortivpn_` as root, and call it as a regular user, who is not in group `dip` it would probably work with F_OK, but not with X_OK. I would expect similar issues with `sudo` configurations. The calling user might be allowed to call `sudo openfortivpn` but he might not be allowed to run `pppd` directly.
Does it make sense? 
For reference: We have started discussing this a little bit in #197

DimitriPapadopoulos:
It works for me with `sudo`- without being member of `dip`. Also as far as I can understand `sudo` is mandatory because running `/usr/sbin/pppd` is not the only requirement. See README:
> openfortivpn needs elevated privileges at three steps during tunnel set up:
> * when spawning a `/usr/sbin/pppd` process;
> * when setting IP routes through VPN (when the tunnel is up);
> * when adding nameservers to `/etc/resolv.conf` (when the tunnel is up).

mrbaseman:
ah ok.  sudo(8):

> ...When sudo executes a command, the security policy specifies the execution
> environment for the command.  Typically, the real and effective user and
> group and IDs are set to match those of the target user, ...

so, when called with `sudo` the process is also privileged in the context of ACCESS(2)

What remains would be an _openfortivpn_ binary with SUID root premissions in conjunction with a pppd not executable by the calling user. This is a corner case that probably doesn't need to work, because as you point out, the user would anyway also need the permissions to update the routing table and to modify `/etc/resolv.conf`. On the other hand _openfortivpn_ doesn't drop privileges, so the user might gain those privileges due to the SUID bit.

Anyhow, our README suggests `sudo` rather than setting the suid bit. If not directly invoked as root, `sudo` is the better solution, since it allows more detailed configuration e.g. who is allowed to call it.

DimitriPapadopoulos:
I have created a small test program:
```
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>

static const char pppd_path[] = "/usr/sbin/pppd";

int main(void) {
	if (access(pppd_path, X_OK) != 0)
		fprintf(stderr, "%s: %s.\n", pppd_path, strerror(errno));
}
```

DimitriPapadopoulos:
```
$ ls -l /usr/sbin/pppd
-rwsr-xr-- 1 root dip 390888 Jan 29  2016 /usr/sbin/pppd
$ 
$ $ id
uid=12345(user) gid=12345(user) groupes=12345(user),27(sudo)
$ 
$ gcc -o test test.c
$ ls -l ./test
-rwxrwxr-x 1 user user 8848 Oct 23 18:04 ./test
$ 
$ ./test
/usr/sbin/pppd: Permission denied.
$ 
$ sudo ./test
$ 
$ sudo chown root:user ./test
$ sudo chmod u+s,g-w,o=r ./test
$ 
$ ./test 
/usr/sbin/pppd: Permission denied.
$ 
```
So indeed the test doesn't work in this corner case.

Perhaps we should drop this change and try to fix `stderr` output as suggested by the links you've reported in #210.

DimitriPapadopoulos:
I think it's best to close this pull request and work on `stderr` output in #201 , both if `execv()` fails or if it succeeds but `/usr/sbin/pppd` exits with error

See #210 for ideas on how to best handle this.

PR #213, opened 2017-10-21
Call _exit() instead of exit() after fork()

DimitriPapadopoulos:
See for example [Why use _exit rather than exit in the child branch of a fork? ](http://www.unixguide.net/unix/programming/1.1.3.shtml) in the Unix programming FAQ.


mrbaseman:
good catch. This is the only `exit()` which is affected. The fork never reaches the one in `main.c`.

PR #220, opened 2017-10-23
Better error handling of pppd exit status

DimitriPapadopoulos:
Use log_error() instead of log_debug() in case of pppd failure.
In case of abnormal exit, print the termination signal.

Use shorter variable names for readability.


DimitriPapadopoulos:
See #210 which for some reason got discarded.

PR #225, opened 2017-11-01
Fix Coverity defects and other issues

DimitriPapadopoulos:


mrbaseman:
looks good to me. I did check that the received data fits into the buffer, but I have missed that the terminating '\0' character might get overwritten. Thanks for spotting this issue.

DimitriPapadopoulos:
Actually I hadn't spotted it either. Coverity emitted a half-false positive, in the sense that it complained about a buffer overflow because it cannot infer the string is null-terminated thanks to memset(), but it did force me to investigate which is how I found this terminal '\0' issue.

The Coverity false positive hasn't go away though. Perhaps we could re-write this piece of code at some point: if it's not obvious to Coverity and its programmers, it won't be obvious to future maintainers. I was thinking of a dedicated function to read the response line by line, and we read lines until the contain 200, or something like that.

mrbaseman:
hmm... one could replace the `memset` by an assignment of the terminating `'\0'` inside of the loop if that makes Coverity happy

Concerning the dedicated function to read the response here a few things I have already been looking into:
- I was searching for a best practice how to read line by line from a socket but unfortunately [there is no concept of line in socket transmissions](https://stackoverflow.com/questions/9828752/read-line-by-line-from-a-socket-buffer). 
- To me the [suggested implementation](https://stackoverflow.com/questions/9828752/read-line-by-line-from-a-socket-buffer) seemed a bit lengthy, so I have extracted from there what I thought I need. but maybe a separate function like this could be an option.
- Probably one has to read single characters and rely on the internal buffering, because we don't know the type of line endings of the proxy server a priori and [one can not push back (unget) characters to a scoket](https://stackoverflow.com/questions/8687547/unread-unget-to-a-bsd-socket). 
- It's also not easily possible to put another buffering layer into the IO stack here, because we have to hand over the socket to openssl right after the headers have been received,
- and we may not accidentally consume any byte of the ssl handshake in advance otherewise the handshake will fail. For the same reason we also may not leave a character of the http header in the socket (e.g. a newline or carriage return when we did not correctly detect the type of line endings used by the proxy)

DimitriPapadopoulos:
Indeed the fact that you cannot unget a character never helps. That said I'm not sure I have grasped the logic of how to handle the HTTP response from the proxy server. I've read part of the [HTTP/1.1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1) specs:
> ## 6 Response
> 
> After receiving and interpreting a request message, a server responds with an HTTP response message.
> ```
>        Response      = Status-Line               ; Section 6.1
>                        *(( general-header        ; Section 4.5
>                         | response-header        ; Section 6.2
>                         | entity-header ) CRLF)  ; Section 7.1
>                        CRLF
>                        [ message-body ]          ; Section 7.2
> ```
> ### 6.1 Status-Line
> 
> The first line of a Response message is the Status-Line, consisting of the protocol version followed by a numeric status code and its associated textual phrase, with each element separated by SP characters. No CR or LF is allowed except in the final CRLF sequence.
> `Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF`

So my I'm wondering:
* The HTTP/1.1 standard requires CR+LF for line breaks. You attempt to accommodate proxy servers that use LF+CR or plain LF instead. Does this happen in practice, proxy servers with different line ending schemes?
* An empty header line precedes the message body. That's precisely what you're looking for in the _while_():  `"\r\n\r\n"`, `"\n\r\n\r"`, `"\n\n"`. So the whole point is to read until and only until we hit the double CR+LF sequence (or its variants if we have to support different line ending schemes).
* In the same way reading a line means reading until we see any of`"\r\n"`, `"\n\r"`, `"\n" (CR+LF or isr variants). I think writing a function that does precisely that should be easy. That would be our line reading function. Note that you have to consume an extra character (at least in the first line) if you're talking to an `\n` proxy server and need to make sure it isn't an `"\n\r"` proxy server. But again are there non-standard proxy servers out there using something different from the standard CR+LF?
* Once we have that line reading function, we just need to parse differently each line: first line should be parsed in more detail, rest of the header can probably be ignored, empty line means the SSL handshake starts with the next character, so we must stop reading and return.
* What happens if the proxy returns something else than 200? Shouldn't we print an error and return an error?

DimitriPapadopoulos:
According to the [RFC2616 HTTP/1.1 standard](https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2):
> HTTP/1.1 defines the sequence CR LF as the end-of-line marker for all protocol elements except the entity-body (see appendix 19.3 for tolerant applications). The end-of-line marker within an entity-body is defined by its associated media type, as described in section 3.7.
> ```
> CRLF           = CR LF
> ```

The standard gives some [latitude for compatibility with broken HTTP software](https://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html#sec19.3):
> The line terminator for message-header fields is the sequence CRLF. However, we recommend that applications, when parsing such headers, recognize a single LF as a line terminator and ignore the leading CR.

So lines endings can be either of `\r\n` and `\n`. Have you heard `\n\r` on any server? Additionally I would argue that proxy servers nowadays are compliant and use `\r\n` - would you trust your HTTP traffic to a non-compliant proxy server?

DimitriPapadopoulos:
From the [RFC7230 HTTP/1.1 standard]:
> The normal procedure for parsing an HTTP message is to read the start-line into a structure, read each header field into a hash table by field name until the empty line, and then use the parsed data to determine if a message body is expected.

About the size of the `request` buffer used to read the HTTP response header:

The [RFC7230 HTTP/1.1 standard] states:
> Various ad hoc limitations on request-line length are found in practice.  It is RECOMMENDED that all HTTP senders and recipients support, at a minimum, request-line lengths of 8000 octets.

[and](https://tools.ietf.org/html/rfc7230#section-3.2.5):
> #### 3.2.5.  Field Limits
> 
> HTTP does not place a predefined limit on the length of each header field or on the length of the header section as a whole, as described in Section 2.5. Various ad hoc limitations on individual header field length are found in practice, often depending on the specific field semantics.
> A server that receives a request header field, or set of fields, larger than it wishes to process MUST respond with an appropriate 4xx (Client Error) status code. Ignoring such header fields would increase the server's vulnerability to request smuggling attacks (Section 9.5).
> 
> A client MAY discard or truncate received header fields that are larger than the client wishes to process if the field semantics are such that the dropped value(s) can be safely ignored without changing the message framing or response semantics.

It looks like 8000 bytes is enough in practice, especially when communicating with a proxy server. That could simplify parsing the response header.

mrbaseman:
First of all, the proxy code is not perfect. We can improve it in several aspects. Adding authentication at the proxy is also still to be done, socks proxies are another popular family of proxies which do not (yet support). But first to answer your questions:
- I have tried a couple public proxies. Unfortunately they refused to connect to ports different than 80 and 443, so I couldn't test much further. I'm sure there were different line endings, `\r\n` and plain `\n`, but I'm not sure anymore about `\n\r`. I have added it because we have to distinguish between two types anyway. Maybe this is unnecessary - it is/was used in [Risc OS](https://en.wikipedia.org/wiki/Newline) but perhaps not when talking over the network.
- we can increase the buffer to 8192 octets, although I doubt that this is necessary for the case of a proxy. I have also seen recommendations that clients should never rely on a fixed buffer size and allocate more if needed. Anyhow, the worst that can happen is that the end of the header is not found and _openfortivpn_ just refuses to establish a connection with that proxy, but with 8192 I believe we are on the safe side (if we aren't already).
- About the status code: In principle any 2xx code should be considered as successful, whereas I have often seen 4xx codes when the proxy refused to connect to specific ports. On the other hand I have only seen 200 for success, decorated with different Reason-Phrases like 'OK' or 'OK connection established' etc. We can be more strict to the standard here and look for the 2xx code on the first line after the first blank. I'm not sure if there might be applications that violate that part of the specification and use tabs or two blanks between the fields. 
- I think we can trust any proxy implementation as long as we receive a valid server certificate for the ssl connection that either matches the host name to which we connect or it matches a digest that we already trust beforehand. The proxy is just something on the way through the insecure internet, which ssl is designed to deal with (provided that it is used correctly, but the proxy code doesn't touch the ssl part at all).
Finally, should we move this discussion to a new issue?

DimitriPapadopoulos:
Sure, will open a new issue for further modifications of the proxy code.

Issue #222, opened 2017-10-23
configure checks for functions and headers 


mrbaseman:
AC_CHECK_FUNCS in configure.ac checks for the existence of a couple of functions, but the list seems to be incomplete. I have extracted a list of calls to functions or function-like calls to macros for instance, and removed those defined in the source itself. However, a quite lengthy list remains: 
`assert atoi cast_addr close connect CRYPTO_set_id_callback CRYPTO_set_locking_callback execv exit _exit fclose fcntl fflush fopen forkpty fprintf fputs free freeaddrinfo freeifaddrs freopen fwrite getaddrinfo getchar getenv getopt_long htons index inet_addr inet_ntoa isatty malloc memcpy memmem memmove memset open openlog OPENSSL_free pclose popen printable_char printf pthread_cancel pthread_cond_init pthread_cond_signal pthread_cond_wait pthread_join pthread_mutexattr_init pthread_mutexattr_setrobust pthread_mutex_destroy pthread_mutex_init pthread_mutex_lock pthread_mutex_unlock pthread_sigmask puts read realloc rewind  select setenv sigaddset sigemptyset signal snprintf socket sprintf SSL_CTX_free SSL_CTX_new SSL_CTX_set_options  SSL_free SSL_get_error SSL_get_peer_certificate SSL_get_verify_result SSL_library_init SSL_load_error_strings SSL_new SSL_read SSL_set_mode SSL_set_verify SSL_shutdown  SSL_write strcasestr strcat strchr strcmp strcpy strdup strerror strlen strncasecmp strncat strncpy strsignal strstr strtob strtok strtok_r strtol syslog system tcsetattr usleep va_end va_start vprintf vsnprintf vsyslog write`
If I add the list to AC_CHECK_FUNCS the configure run performs a number of checks and adds the result to the Makefile: 
[...]

PR #227, opened 2017-11-07
Improve autoconf

mrbaseman:
this branch is to implement the improvements to the autoconf system as discussed in #222 

It is not yet ready to be merged, but in order to see how travis reacts on those changes I make it a pull request already

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,12 +56,13 @@ For other distros, you'll need to build and install from source:
 
 1.  Install build dependencies.
 
-    * RHEL/CentOS/Fedora: `gcc` `automake` `autoconf` `openssl-devel`
-    * Debian/Ubuntu: `gcc` `automake` `autoconf` `libssl-dev`
-    * Arch Linux: `gcc` `automake` `autoconf` `openssl`
-    * Gentoo Linux: `net-dialup/ppp`
-    * openSUSE: `gcc` `automake` `autoconf` `libopenssl-devel`
-    * macOS(Homebrew): `automake` `autoconf` `openssl@1.0`
+    * RHEL/CentOS/Fedora: `gcc` `automake` `autoconf` `openssl-devel` `pkg-config`
+    * Debian/Ubuntu: `gcc` `automake` `autoconf` `libssl-dev` `pkg-config`
+    * Arch Linux: `gcc` `automake` `autoconf` `openssl` `pkg-config`
+    * Gentoo Linux: `net-dialup/ppp` `pkg-config`
+    * openSUSE: `gcc` `automake` `autoconf` `libopenssl-devel` `pkg-config`
+    * macOS(Homebrew): `automake` `autoconf` `openssl@1.0` `pkg-config`
+    * FreeBSD: `automake` `autoconf` `libressl` `pkgconf` (port still in progress)
 
     On Linux, if you manage your kernel yourself, ensure to compile those modules:
     ```
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,20 +1,20 @@
 #!/bin/sh
 set -exu
 
-if ! type aclocal >/dev/null 2>/dev/null ; then
-  echo "aclocal not found - please install automake and autoconf" >&2
-  exit 1
-fi
 if ! type autoconf >/dev/null 2>/dev/null ; then
   echo "autoconf not found - please install it" >&2
   exit 1
 fi
+
 if ! type automake >/dev/null 2>/dev/null ; then
   echo "automake not found - please install it" >&2
   exit 1
 fi
 
-aclocal
+if type aclocal >/dev/null 2>/dev/null ; then
+  aclocal
+fi
+
 autoconf
 automake --add-missing
 
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,6 +1,21 @@
 #!/bin/sh
 set -exu
 
+if ! type aclocal >/dev/null 2>/dev/null ; then
+  echo "aclocal not found - please install automake and autoconf" >&2
+  exit 1
+fi
+if ! type autoconf >/dev/null 2>/dev/null ; then
+  echo "autoconf not found - please install it" >&2
+  exit 1
+fi
+if ! type automake >/dev/null 2>/dev/null ; then
+  echo "automake not found - please install it" >&2
+  exit 1
+fi
+
 aclocal
 autoconf
 automake --add-missing
+
+echo "now you can run ./configure && make to build openfortivpn" >&2
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 #                                               -*- Autoconf -*-
 # Process this file with autoconf to produce a configure script.
 
-AC_PREREQ([2.68])
+AC_PREREQ([2.63])
 AC_INIT([openfortivpn], [1.6.0])
 AC_CONFIG_SRCDIR([src/main.c])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
@@ -9,6 +9,8 @@ AM_INIT_AUTOMAKE([foreign subdir-objects])
 # Checks for programs.
 AC_PROG_CC
 AC_GNU_SOURCE
+m4_ifndef([PKG_PROG_PKG_CONFIG], [m4_fatal([Please install pkg-config.])])
+PKG_PROG_PKG_CONFIG
 
 # Helps support multiarch by setting 'host_os' and 'host_cpu'
 AC_CANONICAL_HOST
@@ -21,7 +23,7 @@ AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthr
 AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])
 
 # Checks for header files.
-AC_CHECK_HEADERS([arpa/inet.h fcntl.h netdb.h netinet/in.h stddef.h stdint.h stdlib.h string.h strings.h sys/ioctl.h sys/socket.h unistd.h])
+AC_CHECK_HEADERS([arpa/inet.h assert.h ctype.h errno.h fcntl.h getopt.h ifaddrs.h limits.h mach/mach.h netdb.h net/if.h netinet/in.h netinet/tcp.h net/route.h pty.h semaphore.h signal.h stdarg.h stddef.h stdint.h stdio.h stdlib.h string.h strings.h sys/ioctl.h syslog.h sys/socket.h sys/stat.h sys/time.h sys/types.h sys/wait.h termios.h unistd.h util.h])
 
 # Checks for typedefs, structures, and compiler characteristics.
 AC_C_INLINE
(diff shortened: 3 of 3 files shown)


src/ipv4.h
```diff
@@ -18,6 +18,8 @@
 #ifndef _OPENFORTIVPN_IPV4_H
 #define _OPENFORTIVPN_IPV4_H
 
+#include <sys/types.h>
+#include <sys/socket.h>
```
DimitriPapadopoulos: Out of curiosity, what are `sys/types.h` and `sys/socket.h` needed for? Perhaps they should be added to source files instead of this header file.
Is this related to `rtentry`?

src/log.c
```diff
@@ -54,7 +54,7 @@ void init_logging(void)
 	is_a_tty = isatty(STDOUT_FILENO);
 
 	pthread_mutexattr_init(&mutexattr);
-#ifndef __APPLE__
+#ifdef HAVE_PTHREAD_MUTEXATTR_SETROBUST
```
DimitriPapadopoulos: Excellent! It would be great if all \_\_APPLE\_\_ ifdef's could eventually be changed the same way.

src/ssl.h
```diff
@@ -41,6 +41,9 @@
  * this pragma selectively suppresses this type of warnings in clang
  */
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
+#ifndef ERESTART
 /*
  * ERESTART is one of the recoverable errors which might be returned.
  * However, in OSX this constant is not defined in errno.h so we define
```
DimitriPapadopoulos: Perhaps the documentation should refer to (Open)BSD in addition to OS X.

autogen.sh
```diff
@@ -17,3 +17,5 @@ fi
 aclocal
 autoconf
 automake --add-missing
+
+echo "now you can run ./configure && make to build openfortivpn" >/dev/stderr
```
DimitriPapadopoulos: In Bourne shell, shouldn't we use `>&2` instead of `>/dev/stderr`?

I believe `>/dev/stderr` is **not** portable, and wasn't even aware it can work on Linux...

README.md
```diff
@@ -57,7 +57,7 @@ For other distros, you'll need to build and install from source:
 1.  Install build dependencies.
 
     * RHEL/CentOS/Fedora: `gcc` `automake` `autoconf` `openssl-devel`
-    * Debian/Ubuntu: `gcc` `automake` `autoconf` `libssl-dev`
+    * Debian/Ubuntu: `gcc` `automake` `autoconf` `libssl-dev` `pkg-config`
```
DimitriPapadopoulos: I suspect `pkg-config` should be added to build dependencies of all platforms: it has to be possible to install `automake`/`autoconf` without `pkg-config` on most if not all platforms, even though that might occur seldom.

src/ssl.h
```diff
@@ -41,6 +41,9 @@
  * this pragma selectively suppresses this type of warnings in clang
  */
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
DimitriPapadopoulos: This is not specific to OS X as far as I can see, rather specific to the Clang compiler - perhaps a specific version of it? Clang might be used instead of GCC on Linux as well.

Change `#ifdef __APPLE__` to `__clang__`.

src/ipv4.h
mrbaseman: yes, this was following the synopsis of https://www.freebsd.org/cgi/man.cgi?query=rtentry&sektion=9 but in the meantime I have noticed that OpenBSD and FreeBSD, compared to linux, both use a different implementation for `rtentry`

DimitriPapadopoulos: Like OS X perhaps?

If so perhaps we can re-use the OS X impelmentation with BSDs? Then the `#ifdef __APPLE__` could be changed to something like `#ifndef HAVE_RTRENTRY` or something like this. Is there something like this in _autoconf_?

Otherwise, yes, the additional headers should be guarded (if possible) by something like `#ifdef __BSD__`.

DimitriPapadopoulos: The [`configure.in` from _tcpdump_](https://opensource.apple.com/source/tcpdump/tcpdump-79.60.1/tcpdump/configure.in.auto.html) might be a good source of inspiration.

mrbaseman: I think the situation for OSX is different. We provide the definition of `rtentry` for OSX. For BSDs there is already a struct with that name and slightly different fields. With C we can't simply extend or overwrite that structure, but I still have the hope that I can define a few macros accordingly. Once the code compiles I can put around some `#ifdef` constructs, hopefully smart ones that use the outcome of configure tests

README.md
mrbaseman: I have added pkg-config as dependency, but how do I mark this request in github as solved?

autogen.sh
mrbaseman: same here, I have changed it to `>&2` as requested

mrbaseman:
@DimitriPapadopoulos You have approved my changes. That's what I meant. Github was still showing in red "Changes requested", and now it shows "Changes approved" in green.

shall I move the BSD support to a separate branch so that we can merge part of this already? I think I have to squash at least some of the commits anyway so that we have a clean history in the master branch with atomic changes.

DimitriPapadopoulos:
Yes, it's perhaps best to move BSD support in a separate branch.
You could then squash the autoconf improvements in a single commit.
BSD support would a different commit.

mrbaseman:
I have moved BSD support to my [bsd_support](https://github.com/mrbaseman/openfortivpn/commits/bsd_support) branch and squashed everything else here

DimitriPapadopoulos:
I recommend the test for `/usr/bin/pppd` is removed from `autogen.sh` because this is a run-time requirement, not a build-time requirement.
We have been emitting a run-time error message when `/usr/bin/pppd` is missing since version 1.6.0.

mrbaseman:
I have done so and squashed once again

DimitriPapadopoulos:
Sorry. One last point I had forgotten:
* Should we perhaps check for `pkg-config` in `autogen.sh`?
* Why check for `aclocal`? Isn't `aclocal` part of `automake`? On Ubuntu 16.04 LTS:
```
$ dpkg -S /usr/bin/aclocal-1.15 
automake: /usr/bin/aclocal-1.15
$ 
dpkg -S /usr/bin/automake-1.15 
automake: /usr/bin/automake-1.15
$ 
```

mrbaseman:
you are right, `aclocal` is part of `automake`, but I have the impression that this has not always been in the same package (e.g. [this thread](https://askubuntu.com/questions/45480/how-do-i-install-aclocal) makes me believe that aclocal has been in a separate dev-package before). 
The package names, especially when it comes to devel packages, depend on the distribution, so it is hard to give a correct advice what to install. I have chosen to recommend to install `automake` and `autoconf` in `autogen.sh`, hoping that other dependencies are automatically installed. Maybe we can leave it like this until someone comes along and tells us "Hey, I have installed both, but `autogen.sh` didn't tell me that on my linux distribution [...] I have to install devel-package-xyz which contains `aclocal`"

For pkg-config I have already added [this check](https://github.com/adrienverge/openfortivpn/pull/227/commits/809bde25af430e00c5ba27d1c12c42a4d1f5593b#diff-67e997bcfdac55191033d57a16d1408aR12) in `configure.ac`

I have already moved the check for `sys/mutex.h` to the [bsd_support](https://github.com/mrbaseman/openfortivpn/commits/bsd_support) branch, `sys/types.h` is always included [in hdlc.h](https://github.com/adrienverge/openfortivpn/blob/master/src/hdlc.h#L24) and `sys/socket.h` is included also on MacOSX [in ipv4.h](https://github.com/adrienverge/openfortivpn/blob/master/src/ipv4.h#L25) and in the long term we are aiming to get rid of `#ifdef __some_os__` statements in favor of `#ifdef _AC_HAS_SOME_HEADER` so we should keep such header checks in `configure.ac` and react on the result of those checks.

DimitriPapadopoulos:
Also, from the [automake manual](https://www.gnu.org/software/automake/manual/automake.html#Future-of-aclocal):

> ### 6.3.6 The Future of `aclocal`
> 
> `aclocal` is expected to disappear. This feature really should not be offered by Automake. Automake should focus on generating `Makefile`s; dealing with M4 macros really is Autoconf’s job. The fact that some people install Automake just to use `aclocal`, but do not use `automake` otherwise is an indication of how that feature is misplaced.
> 
> The new implementation will probably be done slightly differently. For instance, it could enforce the `m4/`-style layout discussed in [Local Macros](https://www.gnu.org/software/automake/manual/automake.html#Local-Macros).
> 
> We have no idea when and how this will happen. This has been discussed several times in the past, but someone still has to commit to that non-trivial task.
> 
> From the user point of view, `aclocal`’s removal might turn out to be painful. There is a simple precaution that you may take to make that switch more seamless: never call `aclocal` yourself. Keep this guy under the exclusive control of `autoreconf` and Automake’s rebuild rules. Hopefully you won’t need to worry about things breaking, when `aclocal` disappears, because everything will have been taken care of. If otherwise you used to call `aclocal` directly yourself or from some script, you will quickly notice the change.

My point is that it's best not to mess with `aclocal` and be totally agnostic about it...

mrbaseman:
but `autogen.sh` in its current version on master [explicitly calls](https://github.com/adrienverge/openfortivpn/blob/master/autogen.sh#L4) `aclocal`. As an alternative we could change that to 
```
if type aclocal >/dev/null 2>/dev/null ; then
  aclocal
fi
```
i.e. if `autogen.sh` finds `aclocal` happily call it and if not simply ignore it

DimitriPapadopoulos:
By the way, you're of course correct when using the preexisting check for `pkg-config` in _autoconf_ instead of rolling out your own in `autogen.sh`.

Back to `aclocal`: I understand we shouldn't be calling `aclocal` explicitly. Not sure why we need to call it at this point, except that according to the following work flow diagram `aclocal` automatically generates `aclocal.m4` from `configure.ac` (since `acinclude.m4`, `m4/*.m4` files and `acsite.m4` are missing). I understand `autoconf` should be able to do without `aclocal.m4`, or perhaps generate it itself, if properly configured in `configure.ac`.
aclocal work flow diagram

DimitriPapadopoulos:
Based on the [autoconf manual](https://www.gnu.org/software/autoconf/manual/html_node/Making-configure-Scripts.html), `aclocal.m4` looks like it's optional:
> ## 3 Making `configure` Scripts
> [...]
> Files used in preparing a software package for distribution, when using just Autoconf:
> ```
>      your source files --> [autoscan*] --> [configure.scan] --> configure.ac
>      
>      configure.ac --.
>                     |   .------> autoconf* -----> configure
>      [aclocal.m4] --+---+
>                     |   `-----> [autoheader*] --> [config.h.in]
>      [acsite.m4] ---'
>      
>      Makefile.in
> ```

mrbaseman:
I have just made the funny observation that without explicitly calling `aclocal` in a freshly cloned repository the built-in `PKG_PROG_PKG_CONFIG` macro is not found when cloning from my branch, with the current master it bails out due to undefined `AM_INIT_AUTOMAKE` and `AM_SILENT_RULES`

mrbaseman:
I have implemented my suggestion of https://github.com/adrienverge/openfortivpn/pull/227#issuecomment-347183354 in https://github.com/adrienverge/openfortivpn/pull/227/commits/9303a53217dea0dca9cd4b3ca80a914185da098c now. Maybe we leave this as a separate commit. @DimitriPapadopoulos would this be ok for you to merge?

DimitriPapadopoulos:
Please do merge.

Let's just keep in mind this `aclocal` issue, perhaps some day we can avoid calling it explicitly.

PR #240, opened 2017-11-28
Bourne shell

DimitriPapadopoulos:
* Standard stdout/stderr redirection.
* The last message is informational, not an error. Use sdtout, not stderr.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,21 +1,21 @@
 #!/bin/sh
 set -exu
 
-if ! type autoconf >/dev/null 2>/dev/null ; then
+if ! type autoconf >/dev/null 2>&1 ; then
   echo "autoconf not found - please install it" >&2
   exit 1
 fi
 
-if ! type automake >/dev/null 2>/dev/null ; then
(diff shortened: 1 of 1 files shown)


autogen.sh
```diff
@@ -1,21 +1,21 @@
 #!/bin/sh
 set -exu
 
-if ! type autoconf >/dev/null 2>/dev/null ; then
+if ! type autoconf >/dev/null 2>&1 ; then
```
mrbaseman: is this more portable? I thought `2>&1` was bash specific syntax

DimitriPapadopoulos: It's actually the standard syntax for Bourne shell and the most portable across `ksh`, `zsh`, `bash`, etc.

DimitriPapadopoulos: See for example [sh(1) man page](https://docs.oracle.com/cd/E86824_01/html/E54763/sh-1.html) or [_Bourne Shell Scripting_](https://en.wikibooks.org/wiki/Bourne_Shell_Scripting/Files_and_streams#Redirecting_standard_error_.28and_other_streams.29)Wikibook.

PR #242, opened 2017-12-07
Use X509_check_host instead of explicit CN match.

Lobstros:
Hi,

When establishing a connection, the Common Name field in the target's certificate is explicitly checked, which results in an (overridable) failure when the hostname is not in the CN, but is in the Subject Alternative Name field (e.g. in multi-domain certs).

I've changed this to use openssl native X509_check_host instead--[the wiki](https://wiki.openssl.org/index.php/Hostname_validation) reports this is available in 1.0.2+/1.1+, and it adds an additional dependency on <openssl/x509v3.h>. Hopefully that's ok?

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -39,6 +39,7 @@
 #include <net/if.h>
 #include <arpa/inet.h>
 #include <openssl/err.h>
+#include <openssl/x509v3.h>
 #ifndef __APPLE__
 #include <pty.h>
 #else
@@ -490,14 +491,8 @@ static int ssl_verify_cert(struct tunnel *tunnel)
 		return 1;
 	}
 
-	subj = X509_get_subject_name(cert);
-
 	// Try to validate certificate using local PKI
-	if (subj
-	    && X509_NAME_get_text_by_NID(subj, NID_commonName, common_name,
-	                                 FIELD_SIZE) > 0
-	    && strncasecmp(common_name, tunnel->config->gateway_host,
-	                   FIELD_SIZE) == 0
+	if (X509_check_host(cert, common_name, FIELD_SIZE, 0, NULL)
 	    && SSL_get_verify_result(tunnel->ssl_handle) == X509_V_OK) {
 		log_debug("Gateway certificate validation succeeded.\n");
 		ret = 0;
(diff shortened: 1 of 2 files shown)


Lobstros:
Drat! Apparently the CI builds against OpenSSL 1.0.1f. Would you prefer to keep it this way? If so I can maybe try to implement explicitly reading and matching SAN fields against the hostname.

mrbaseman:
you could add `openssl/x509v3.h` to `AC_CHECK_HEADERS` in [configure.ac](https://github.com/adrienverge/openfortivpn/blob/master/configure.ac#L26) and wrap your code into a section `#ifdef HAVE_OPENSSL_X509V3_H`. Then it would be used only when built against Openssl >= 1.0.2 and otherwise fall back to the existing code.

Lobstros:
Thanks for the suggestion--ended up using HAS_X509_CHECK_HOST instead (since the .h file is still present in older versions). That needed a new line of AC_CHECK_FUNCS so it's not fatal when the function isn't found.

Hopefully the style used is ok; let me know if anything needs changing.

mrbaseman:
The changes themselves look good to me. Two minor things:
Could you add a brief comment in configure.ac which explains that the second AC_CHECK_FUNCS line is not fatal when the function is not found?
Could you squash the commits and update the commit message accordingly?
@DimitriPapadopoulos, do you have anything to add?

src/tunnel.c
```diff
@@ -39,6 +39,7 @@
 #include <net/if.h>
 #include <arpa/inet.h>
 #include <openssl/err.h>
+#include <openssl/x509v3.h>
```
DimitriPapadopoulos: Just wondering: is this OpenSSL header **always** available?

Lobstros:
Sure thing--added the comment, but I can't figure out how to squash the commits and have GitHub not reject the push. Will I need to delete and re-make my fork?

Regarding `x509v3.h`, it's present in the earliest version on the OpenSSL website (0.9.6), and comments in the file itself seem to suggest it was introduced in 1999, so probably sometime in the first couple of releases (I *really* hope no-one is still running those in the wild ;-) ).

DimitriPapadopoulos:
See for example:
* [How to squash commits in a GitHub pull request](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request)
* [Squash All Commits Related to a Single Issue into a Single Commit](https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit)
* [Squashing Github pull requests into a single commit](https://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit)
* ...

PR #243, opened 2017-12-10
allow use of openfortivpn by unprivileged user (in group "dip")

gbon121:
After moving privileged pppd options from command line to a config file (owned by root) it is possible for any user in group dip to start pppd through openfortivpn.

Static routes and DNS settings are managed by /etc/ppp/ip-up.local and /etc/ppp/ip-down.local scripts, provided the following lines are added to /etc/openfortivpn/config:
```
set-routes = 0
set-dns = 0
pppd-ipparam = openfortivpn
pppd-call = openfortivpn
```

Tried this approach on Linux (debian stretch) and it works fine for my very limited setup, except for a few harmless error messages at tunnel shutdown when pppd is killed directly with SIGTERM.  I haven't macs or BSDs to test on.


mrbaseman:
Could you introduce a command line switch and/or a config file parameter that changes the behavior in the way you need it instead of breaking normal usage, please?

gbon121:
Introduced pppd-call option to change pppd invocation mode.

mrbaseman:
@gbon121 A big bonus for this PR is that it addresses the long-standing issue of the requirement to run under full root permissions **and** it also brings in a real-life example for a _pppd_ `ip-up` script which can guide other users who would like to adapt it to cover their use cases. Therefore, I'm very interested in merging this into our master branch.

However, the new scripts are not installed at all by the automake procedure. I think we need some extra hooks for them in `Makefile.am`. Probably we should install them as templates like the config file template to `$(DESTDIR)$(datadir)` and mention their existence in the man page. Admins who want to allow their dip-users to open vpn connections can pick them up from there and place the scripts to the locations where pppd is looking for them. A propos: The man page should mention the new command line option, too. 

I'm a bit hesitating about putting all the files directly in place because not all run-time requirements might be fulfilled on all platforms, and adaptions to the scripts by the admin of each machine might get lost by accident when they update _openfortivpn_.

Apart from that, I think it looks good, although I haven't tested it yet. 
@DimitriPapadopoulos and @adrienverge do you have any further comments?

PR #249, opened 2018-02-06
systemd script

uryossa:


examples/openfortivpn.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description = Forti VPN for default
+
+[Service]
+ExecStart = /usr/bin/openfortivpn --use-syslog -c /etc/openfortivpn/config
```
adrienverge: Can you remove `--use-syslog`? It is not the default and most systems use Systemd's journal.

examples/openfortivpn.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description = Forti VPN for default
```
adrienverge: It's not `Forti VPN`, it's `openfortivpn` ;-)

examples/openfortivpn.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
adrienverge:
```
Requires = network-online.target 
```
would be useful in this section, what do you think?

uryossa:
I'm sorry.
I don't have enough time for understanding english and learning github.

Issue #154, opened 2017-07-25
ERROR: read: Input/output error


mvernimmen-CG:
I was playing with openfortivpn on my raspberry pi3 (This 1.4.0 on Raspbian GNU/Linux 8 (jessie)) and noticed this error after connecting:
```
root@raspberrypi:~# openfortivpn xxxx:443 --username=xxxx --realm=xxxxx -v
WARN:   Bad port in config file: "0".
DEBUG:  Loaded config file "/etc/openfortivpn/config".
VPN account password:
DEBUG:  Config host = "xxxxx"
DEBUG:  Config realm = "xxxx"
DEBUG:  Config port = "443"
DEBUG:  Config username = "xxxxx"
DEBUG:  Config password = "********"
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Connected to gateway.
INFO:   Authenticated.
DEBUG:  Cookie: SVPNCOOKIE=xxxx
INFO:   Remote gateway has allocated a VPN.
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
DEBUG:  pppd_read_thread
DEBUG:  ssl_read_thread
DEBUG:  gateway ---> pppd (12 bytes)
DEBUG:  ssl_write_thread
DEBUG:  if_config thread
DEBUG:  pppd ---> gateway (16 bytes)
DEBUG:  pppd_write thread
DEBUG:  pppd ---> gateway (12 bytes)

DEBUG:  gateway ---> pppd (24 bytes)
INFO:   Got addresses: [172.19.64.129], ns [172.19.30.252, 172.19.30.5]
DEBUG:  pppd ---> gateway (944 bytes)
DEBUG:  pppd ---> gateway (10 bytes)
DEBUG:  pppd ---> gateway (25 bytes)
DEBUG:  pppd ---> gateway (25 bytes)
ERROR:  read: Input/output error
INFO:   Cancelling threads...
DEBUG:  Waiting for pppd to exit...
INFO:   Terminated pppd.
INFO:   Closed connection to gateway.
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Logged out.
```

When applying the workaround mentioned in #56 about setting a route before starting the vpn, the connection builds up fine. I do need to add routes manually after having the vpn up, to make sure I can reach the other side through the vpn.
```
[...]

PR #250, opened 2018-02-06
Update ipv4.c

jon2kx:
This should fix a ton of the issues that come up in https://github.com/adrienverge/openfortivpn/issues/154

Please look over my work, I'm new in C programming, I've been a shell and python and perl guy most of my career.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -138,7 +138,10 @@ static int ipv4_get_route(struct rtentry *route)
 	int len = sizeof(buffer) - 1;
 	char *saveptr3 = NULL;
 	
-	// Open the command for reading and properly account for the path on the local linux machine 
+	// Open the command for reading and
+	// properly account for the path on
+	// the local linux machine 
+	
 	wnst = popen("which netstat");
 	fp = popen("wnst -f inet -rn", "r");
 	if (fp == NULL)
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -138,9 +138,6 @@ static int ipv4_get_route(struct rtentry *route)
 	int len = sizeof(buffer) - 1;
 	char *saveptr3 = NULL;
 	
-	
-	
-
 	// Open the command for reading and properly account for the path on the local linux machine 
 	wnst = popen("which netstat");
 	fp = popen("wnst -f inet -rn", "r");
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -137,7 +137,7 @@ static int ipv4_get_route(struct rtentry *route)
 	FILE *wnst;
 	int len = sizeof(buffer) - 1;
 	char *saveptr3 = NULL;
-	
+
 	// Open the command for reading
 	wnst = popen("which netstat");
 	fp = popen("wnst -f inet -rn", "r");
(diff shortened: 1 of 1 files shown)


jon2kx:
0.02s$ make -C $HOME/astyle/build/gcc prefix=$HOME install
make: Entering directory `/home/travis/astyle/build/gcc'
install -o travis -g travis -m 755 -d /home/travis/bin
install -o travis -g travis -m 755 -d /home/travis/share/doc/astyle
make: Leaving directory `/home/travis/astyle/build/gcc'
0.33s$ ./tests/lint/run.sh
error: src/ipv4.c does not comply with coding style
diff --git a/src/ipv4.c b/tmp/tmp.BO1sP4rRcY
index 57d58ca..7e488ea 100644
--- a/src/ipv4.c
+++ b/tmp/tmp.BO1sP4rRcY
@@ -139 +139 @@ static int ipv4_get_route(struct rtentry *route)
-	char *saveptr3 = NULL;	
+	char *saveptr3 = NULL;

I 'm not sure where I am missing the coding style , I've checked a few times now, I'm new to C, any help in getting this would be appreciated.

adrienverge:
GitHub is showing the problem to you ;-)
problem

By the way you can run linting scripts locally to identify the problem.

DimitriPapadopoulos:
Which OS does this patch target exactly?

jon2kx:
@adrienverge  Thanks for that, the build passes now. 

So I saw your comments about just calling netstat, I assumed that it was called with system() at some point down the line here. 

@DimitriPapadopoulos non-macintosh unix/linux/bsd systems.

DimitriPapadopoulos:
This part of the code is not executed on non-macintosh unix/linux/bsd systems - hence there isn't anything to fix here.

The "Input/output errors" are unrelated to `netstat`.

jon2kx:
#ifndef __APPLE__
	int fd;
	// Cannot stat, mmap not lseek this special /proc file
	fd = open("/proc/net/route", O_RDONLY);
	if (fd == -1)
		return ERR_IPV4_SEE_ERRNO;

	size = read(fd, buffer, sizeof(buffer) - 1);
	if (size == -1) {
		close(fd);
		return ERR_IPV4_SEE_ERRNO;
	}
	close(fd);
#else
	FILE *fp;
	FILE *wnst;
	int len = sizeof(buffer) - 1;
	char *saveptr3 = NULL;

	// Open the command for reading
	wnst = popen("which netstat");
	fp = popen("wnst -f inet -rn", "r");
	if (fp == NULL)
return ERR_IPV4_SEE_ERRNO;

Please excuse me if I have this incorrect, but if the __APPLE__ token is set, this previous block of code executes, else the applicable block is executed? 

Do I have this incorrect, can you help me understand if I have this incorrect?

DimitriPapadopoulos:
These mutliple `ifdef`/`ifndef` occurrence are indeed confusing (notice the missing/extra **n**):
* some of them use `ifdef __APPLE__` (_if on Apple_)
* other use `ifndef __APPLE__` (_if **not** on Apple_)

Ideally the code should be rewritten using `ifdef __APPLE__` only for the sake of clarity.

jon2kx:
@DimitriPapadopoulos  Thank you for helping me understand. I appreciate your time.

Issue #253, opened 2018-02-19
avoid ifndef statements


mrbaseman:
in #250 it was noticed that probably the whole issue was opened based on a misunderstanding. 
The `#ifndef __APPLE__` statements are confusing and should be replaced by `#ifdef __APPLE__` accompanied by the appropriate code-reordering

I have opened this issue as a reminder for myself. I'm going to address this, probably after my next vacation...

DimitriPapadopoulos:
Hasn't this issue been solved by 84b92ac?

mrbaseman:
sure, I just forgot to close the issue

PR #254, opened 2018-02-20
change #ifndef __APPLE__ ... #else ... #endif

mrbaseman:
...to `#ifdef __APPLE__ ... #else ... #endif`
combined with appropriate code reordering, where appropriate. this addresses #253


lkundrak:
The commit message should ideally explain why you did the change (why are things better with the commits applied than without it), not literally explaining what the actual changes were.

mrbaseman:
good point. I have updated the commit message.

DimitriPapadopoulos:
@mrbaseman Thank you Martin, looks good to me now.

PR #261, opened 2018-03-16
 Linux kernel coding style 

DimitriPapadopoulos:


DimitriPapadopoulos:
The `line-length.sh` script contradicts the Linux kernel coding style.
Needs to be fixed to accept strings longer than 80 characters.

adrienverge:
Why? From https://www.kernel.org/doc/html/v4.10/process/coding-style.html:

> The limit on the length of lines is 80 columns and this is a strongly preferred limit.


I find it more readable with 80 chars max.

DimitriPapadopoulos:
This is from the Linux kernel coding style:
> Statements longer than 80 columns will be broken into sensible chunks, unless exceeding 80 columns significantly increases readability and does not hide information. Descendants are always substantially shorter than the parent and are placed substantially to the right. The same applies to function headers with a long argument list. **However, never break user-visible strings** such as printk messages, because that breaks the ability to grep for them.

I'm just the messenger: I don't agree nor disagree, with the guide.

adrienverge:
OK thanks! I didn't know.

I have a slight preference for "everything <= 80 chars" because of my text editor (Vim), but I am open to any new linting rules that are consistent and can be easily enforced.

DimitriPapadopoulos:
Indeed **easily enforced** is the issue at hand. Unfortunately `checkpatch.pl` is not written using `lint` – and to tell the truth I find this script too strict. The transition is probably not straightforward.

The rule about not breaking user-visible strings did make sense to me, because it helps grep'ping and googl'ing an error message. But then that's I don't have very strong feelings about it either. I was merely trying to squash some types of `checkpatch.pl` warnings and errors.

PR #262, opened 2018-03-17
Linux kernel coding style

DimitriPapadopoulos:
Very minor changes suggested by checkpatch.pl


adrienverge:
Looks good. If you'd like to replace current linting scripts with more up-to-date ones (like `checkpatch.pl`, or for example, another script that checks 80-char lines except for long strings), feel free!

DimitriPapadopoulos:
The problem with `checkpatch.pl` is that it generates many warnings, some of which I don't really feel like addressing at this point. So at the very least I would need to disable some errors (not possible as far as I can see) or get the script to check only the pull request (using `--patch`) but then how to run it from `tests/lint/run.sh`?

PR #264, opened 2018-03-19
Make the sure Coverity defect is a false positive

DimitriPapadopoulos:


DimitriPapadopoulos:
@mrbaseman Martin, does this change look OK to you? It's partly cosmetic but I feel this rewrite is needed to explain why the Coverity warning is a false positive.

src/tunnel.c
```diff
@@ -439,44 +443,87 @@ static int tcp_connect(struct tunnel *tunnel)
...
+				 * RFC2616 states in section 19.3 Tolerant Applications:
+				 * 	The line terminator for message-header fields
+				 * 	is the sequence CRLF. However, we recommend
+				 * 	that applications, when parsing such headers,
+				 * 	recognize a single LF as a line terminator
+				 * 	and ignore the leading CR.
+				 */
+				static const char *HTTP_EOL[] = {
+					"\r\n\r\n",
+					"\n\n"
```
mrbaseman: what about the "\n\r\n\r" case? The previous code was a bit more tolerant here, but probably this behavior was against specifications anyway...

src/tunnel.c
```diff
@@ -370,8 +370,12 @@ static int tcp_connect(struct tunnel *tunnel)
 		char *proxy_host, *proxy_port;
 		// protect the original environment from modifications
 		env_proxy = strdup(env_proxy);
+		if (env_proxy == NULL) {
```
mrbaseman: oh, I didn't check this. Good catch!

src/tunnel.c
```diff
@@ -370,8 +370,12 @@ static int tcp_connect(struct tunnel *tunnel)
...
+			log_error("strdup: %s\n", strerror(errno));
+			goto err_strdup;
+		}
 		// get rid of a trailing slash
-		if (env_proxy[strlen(env_proxy) - 1] == '/')
+		if (*env_proxy && env_proxy[strlen(env_proxy) - 1] == '/')
```
mrbaseman: but here it is really cosmetic then, but probably requested by coverity

src/tunnel.c
```diff
@@ -425,7 +429,7 @@ static int tcp_connect(struct tunnel *tunnel)
 	log_debug("gateway_port: %u\n", tunnel->config->gateway_port);
 
 	ret = connect(handle, (struct sockaddr *) &server, sizeof(server));
-	if (ret == -1) {
```
mrbaseman: I was checking explicitly for the [failure value -1](http://pubs.opengroup.org/onlinepubs/009695399/functions/connect.html), but your code is more tolerant about different implementations of the connect() function

src/tunnel.c
DimitriPapadopoulos: Coverity fears _getenv_() might return an empty string "" but I don't know whether this can happen. The documentation doesn't specify that.

src/tunnel.c
DimitriPapadopoulos: It's indeed against specifications. Therefore I would suggest starting with these two variants and adding `\n\r\n\r`if needed.Your call :-)

PR #268, opened 2018-03-22
Update readme for install with MacPorts

millerdev:
Submitted to MacPorts https://trac.macports.org/ticket/56121

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ this process.
 It is compatible with Fortinet VPNs.
 
 
-
 --------
 Examples
 --------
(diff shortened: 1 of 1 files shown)


adrienverge:
Hi, thanks for submitting to MacPorts.

But why do you want to store this `Portfile` here?

millerdev:
So it can be maintained and updated as new versions are released. Feel free to drop the Portfile from this PR if you don't want it.

adrienverge:
Yes I think it's better to drop it. Other package files (for Fedora, Debian, Ubuntu...) are also stored elsewhere.

DimitriPapadopoulos:
Here is an example of a re-organized README.md:
https://github.com/DimitriPapadopoulos/openfortivpn/blob/installing/README.md

Does it help?

millerdev:
Yes, I noticed that. Your version is much better. Updated.

DimitriPapadopoulos:
If you're OK with the above, I can squash into a single commit myself if it helps.

millerdev:
> Also you do realize that you suggest using only Homebrew and not MacPorts for building from sources, don't you?

Yes. The MacPorts build-from-source instructions were exactly the same as above install-with-MacPorts instructions, so it seemed redundant to me.

millerdev:
Squashed. Maybe wait for the [Portfile](https://trac.macports.org/ticket/56121) be added to MacPorts repos before merging?

DimitriPapadopoulos:
Sure, just add a comment when it's ready.

millerdev:
Portfile added to MacPorts: https://github.com/macports/macports-ports/pull/2026

mrbaseman:
@millerdev can you trigger an update of the Portfile? We have released _openfortivpn_ with some fixes in the routing code for OSX.

millerdev:
@mrbaseman https://github.com/macports/macports-ports/pull/3130

Issue #54, opened 2016-04-15
"--ppd-plugin" and "--ppd-log" parameters may be a security risk when running through sudo


mata-p:
When execution of the openfortivpn executable is granted to unprivileged users the way the example on the project frontpage shows, the `--ppp-plugin` and `--ppd-log` arguments are passed directly to pppd as `plugin`, respective `logfile`.

As pppd in this case will be executed as root, `plugin` allows to load a mailicious user controlled plugin from an arbitrary location as root. This could probably be mitigated by only allowing plugin names which don't contain a slash, therfore limiting loading plugins only from the `/usr/lib/pppd/[version]` directory.

`logfile` allows to open (or create) an aribitray file as root and append log output to it. `pppd` when not invoked as root opens the log file as the original user, but in this case it's already invoked as root thorugh sudo.

At the very least the instructions on the front page should be changed not to allow direct execution of the openfortivpn binary but of a wrapper script that doesn't take potentially insecure user arguments.

adrienverge:
@mata-p,

Thanks and sorry for the long delay.

I've just added a section to briefly warn about the problem in the README file. I'm open to any better solution ; pull requests are welcome.

mrbaseman:
I think we can close this, since the readme has been updated a long time ago.

PR #270, opened 2018-03-30
Add --pppd-call option.

gbon121:
On systems where pppd supports the "call" option --- eg. Debian derived
distros --- privileged options to pppd can be moved to a config file owned
by root, and any unprivileged user in group "dip" can invoke openfortivpn.

Static routes and DNS settings are managed by /etc/ppp/ip-up.local and
/etc/ppp/ip-down.local scripts, provided the following lines are added
to /etc/openfortivpn/config:

```
set-routes = 0
set-dns = 0
pppd-ipparam = openfortivpn
pppd-call = openfortivpn
```

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -515,12 +515,37 @@ err_destroy:
 	return ret;
 }
 
+static void add_text_route(struct tunnel *tunnel, const char *dest,
+                           const char *mask, const char *gw)
+{
+	size_t l0, l1;
+	const char fmt[] = ",%s/%s/%s";
+	const char trigger[] = "openfortivpn";
+	char **target = &tunnel->config->pppd_ipparam;
+	char *ptr;
+
+	if (*target == NULL || strncmp(*target, trigger, strlen(trigger)))
+		return;
+	if (!dest || !mask || !gw)
+		return;
+	log_info("Registering route %s/%s via %s\n", dest, mask, gw);
+	l0 = strlen(*target);
+	l1 = strlen(fmt) + strlen(dest) + strlen(mask) + strlen(gw) + 1;
+	if ((ptr = realloc(*target, l0 + l1))) {
+		*target = ptr;
+		snprintf(*target + l0, l1, fmt, dest, mask, gw);
+	} else {
+		log_error("realloc: %s\n", strerror(errno));
+	}
+}
+
 int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
                              char *gateway)
 {
 	struct rtentry *route;
 	char env_var[24];
 
+	add_text_route(tunnel, dest, mask, gateway);
 	if (tunnel->ipv4.split_routes == MAX_SPLIT_ROUTES)
 		return ERR_IPV4_NO_MEM;
 	if ((tunnel->ipv4.split_rt == NULL)
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -145,6 +145,16 @@ static int pppd_run(struct tunnel *tunnel)
 			NULL // terminal null pointer required by execvp()
 		};
 
+		if (tunnel->config->pppd_call) {
+			/* overwrite args[]: keep pppd_path, replace all
+			 * options with "call <name>" */
+			int j = 1;
+			args[j++] = "call";
+			args[j++] = tunnel->config->pppd_call;
+			while (j < ARRAY_SIZE(args))
+				args[j++] = NULL;
+		}
+
 		// Dynamically get first NULL pointer so that changes of
 		// args above don't need code changes here
 		int i = ARRAY_SIZE(args) - 1;
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -26,6 +26,7 @@ openfortivpn \- Client for PPP+SSL VPN tunnel services
 [\fB\-\-pppd-plugin=\fI<file>\fR]
 [\fB\-\-pppd-ipparam=\fI<string>\fR]
 [\fB\-\-pppd-ifname=\fI<string>\fR]
+[\fB\-\-pppd-call=\fI<name>\fR]
 [\fB\-\-persistent\fR]
 [\fB\-c\fR \fI<file>\fR]
 [\fB\-v|\-q\fR]
@@ -129,6 +130,13 @@ for further details
 \fB\-\-pppd-ifname=\fI<string>\fR
 Set the ppp interface name. Only if supported by pppd. Patched versions of pppd
 implement this option but may not be available on your platform.
+.TP
+\fB\-\-pppd-call=\fI<name>\fR
+Drop usual arguments from pppd command line and add `call <name>' instead.
+This can be useful on Debian and Ubuntu, where unprivileged users in
+group `dip' can invoke `pppd call <name>' to make pppd read and apply
+options from /etc/ppp/peers/<name> (including privileged ones).
+.TP
 \fB\-\-persistent\=\fI<interval>fR
 Runs the vpn persistently in an endless loop and tries to reconnect forever.
 The reconnect interval may be specified in seconds, where 0 means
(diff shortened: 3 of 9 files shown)


src/ipv4.c
```diff
@@ -515,12 +515,38 @@ int ipv4_protect_tunnel_route(struct tunnel *tunnel)
...
+		return;
+	log_info("Registering route %s/%s via %s\n", dest, mask, gw);
+	l0 = strlen(*target);
+	l1 = strlen(fmt) + strlen(dest) + strlen(mask) + strlen(gw) + 1;
+	if ((ptr = realloc(*target, l0 + l1))) {
+		*target = ptr;
+		snprintf(*target + l0, l1, fmt, dest, mask, gw);
+	} else {
+		int eno = errno;
+		log_error("Could not reallocate array: %s\n", strerror(eno));
```
DimitriPapadopoulos: Not that the message is any better this way, just mere consistency with the rest of the code:
```
		log_error("realloc: %s\n", strerror(eno));
```

src/ipv4.c
```diff
@@ -515,12 +515,38 @@ int ipv4_protect_tunnel_route(struct tunnel *tunnel)
...
+	if (!dest || !mask || !gw)
+		return;
+	log_info("Registering route %s/%s via %s\n", dest, mask, gw);
+	l0 = strlen(*target);
+	l1 = strlen(fmt) + strlen(dest) + strlen(mask) + strlen(gw) + 1;
+	if ((ptr = realloc(*target, l0 + l1))) {
+		*target = ptr;
+		snprintf(*target + l0, l1, fmt, dest, mask, gw);
+	} else {
+		int eno = errno;
```
DimitriPapadopoulos: While log_error may theoretically change the value of `errno`, for mere consistency with the rest of the code I suggest getting rid of `eno` and using `errno` directly.

DimitriPapadopoulos:
The man page should eventually be updated too: `doc/openfortivpn.1.in`

src/ipv4.c
DimitriPapadopoulos: Actually _log_error_() cannot change the value of `errno` before it is used as the `strerror(errno)` argument is evaluated before the function is called.

gbon121:
On Fri, Mar 30, 2018 at 08:17:27PM +0000, Dimitri Papadopoulos Orfanos wrote:
> DimitriPapadopoulos commented on this pull request.
> 
> What happens to this piece of code when _openfortivpn_ is called by an unprivileged user?
> https://github.com/adrienverge/openfortivpn/blob/25b2585/src/tunnel.c#L126-L146
> 
> > +	char **target = &tunnel->config->pppd_ipparam;
> +	char *ptr;
> +
> +	if (*target == NULL || strncmp(*target, trigger, strlen(trigger)))
> +		return;
> +	if (!dest || !mask || !gw)
> +		return;
> +	log_info("Registering route %s/%s via %s\n", dest, mask, gw);
> +	l0 = strlen(*target);
> +	l1 = strlen(fmt) + strlen(dest) + strlen(mask) + strlen(gw) + 1;
> +	if ((ptr = realloc(*target, l0 + l1))) {
> +		*target = ptr;
> +		snprintf(*target + l0, l1, fmt, dest, mask, gw);
> +	} else {
> +		int eno = errno;
> +		log_error("Could not reallocate array: %s\n", strerror(eno));
> 

Basically, (unprivileged) openfortivpn builds a list of routes to add
after the tunnel is up, and passes the list to (privileged) pppd via
the ipparam option, which is usually employed to pass auxiliary data available
at runtime.

DimitriPapadopoulos:
I must be missing something, but aren't all these options passed to _pppd_ when they should be read from `/etc/ppp/peers/openfortivpn` instead?
```
		static const char *args[] = {
			pppd_path,
			"38400", // speed
			":1.1.1.1", // <local_IP_address>:<remote_IP_address>
			"noipdefault",
			"noaccomp",
			"noauth",
			"default-asyncmap",
			"nopcomp",
			"receive-all",
			"nodefaultroute",
			"nodetach",
			"lcp-max-configure", "40",
			"mru", "1354",
			NULL, // "usepeerdns"
			NULL, NULL, NULL, // "debug", "logfile", pppd_log
			NULL, NULL, // "plugin", pppd_plugin
			NULL, NULL, // "ipparam", pppd_ipparam
			NULL, NULL, // "ifname", pppd_ifname
			NULL // terminal null pointer required by execvp()
		};
		[...]
		execv(args[0], (char *const *)args);
```

DimitriPapadopoulos:
Yes, I was definitely missing something! Perhaps some comments should be added to explain that `--pppd-call` overwrites the `args` array except for the _pppd_ path.

DimitriPapadopoulos:
I was thinking an `if` might be cleaner, something like this:
```
		if (tunnel->config->pppd_call) {
			...
		} else {
			...
		}
```
However I don't see how to initialize `args` and `i`in a more elegant way.

src/tunnel.c
```diff
@@ -145,6 +145,14 @@ static int pppd_run(struct tunnel *tunnel)
 			NULL // terminal null pointer required by execvp()
 		};
 
+		if (tunnel->config->pppd_call != NULL) {
```
DimitriPapadopoulos: Just write:
 ```
		if (tunnel->config->pppd_call) {
 ```
Again I do not care myself but but the [checkpatch.pl](https://github.com/torvalds/linux/blob/master/scripts/checkpatch.pl) script of the Linux kernel emits a warning and we try to follow the Linux kernel coding style.

DimitriPapadopoulos:
Looks good to me. Could you perhaps squash the commits so that I can merge your pull request?

DimitriPapadopoulos:
Does this mean _openfortivpn_ will not need to be run by _root_ anymore on Debian-based systems? If so this looks like a solution for issue #54. We should probably disable `--pppd-plugin` and `--pppd-log` when `--pppd-call` is used. Does this make sense?

No need to address the above issue in this pull request, we shall handle it in a different pull request.

gbon121:
@DimitriPapadopoulos I'm not sure I squashed my commits properly.  Should the PR result more difficult to merge than before, please let me know.

DimitriPapadopoulos:
Actually you've added on more commit instead of replacing all existing commits by a single one, but don't worry: I can squash these commits myself not that I have the definitive comment (the one in the last commit).

gbon121:
@DimitriPapadopoulos I'm not sure the following code is more elegant of the current static array, but it might silence some warnings from Coverity:

```
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>

struct v_arr {
        /* current capacity */
        unsigned cap;
        /* next slot to write to, always < max(cap - 1, 1) */
        unsigned off;
        /* NULL terminated */
        void **data;
};

void append_v_arr(struct v_arr *p, void *x) {
        if (p->off + 1 >= p->cap) {
                void **ndata;
                unsigned ncap = (p->off + 1) * 2;
                assert(p->off + 1 < ncap);
                ndata = realloc(p->data, ncap * sizeof(void *));
                if (ndata) {
                        p->data = ndata;
                        p->cap = ncap;
                } else {
                        abort();
                }
        }
        assert(p->off + 1 < p->cap);
        p->data[p->off] = x;
        p->data[++p->off] = NULL;
}       

void clear_v_arr(struct v_arr *p) {
        free(p->data);
}
        
void use_array_of_char(const char **arr) { 
        const char **p;
        printf("result:");
        for (p = arr; *p; ++p) 
                printf(" %s", *p);
        printf("\n"); 
}               
        
int main() {
        const char *frag[] = { "pppd", "option1", "option2", "option3" };
        struct v_arr v = { 0, 0, 0 };
        unsigned i; 
        for (i = 0; i < sizeof frag / sizeof *frag; i++)
                append_v_arr(&v, (void *)frag[i]);
        use_array_of_char((const char **)v.data);
        clear_v_arr(&v);
        return 0;
}       

```

PR #277, opened 2018-04-04
Linux kernel coding style

DimitriPapadopoulos:
* However, never break user-visible strings such as printk messages, because that breaks the ability to grep for them.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -195,16 +195,15 @@ int load_config(struct vpn_config *cfg, const char *filename)
 		} else if (strcmp(key, "half-internet-routes") == 0) {
 			int half_internet_routes = strtob(val);
 			if (half_internet_routes < 0) {
-				log_warn("Bad half-internet-routes in config"
-				         " file: \"%s\".\n", val);
+				log_warn("Bad half-internet-routes in config file: \"%s\".\n",
+				         val);
 				continue;
 			}
 			cfg->half_internet_routes = half_internet_routes;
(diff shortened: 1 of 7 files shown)


DimitriPapadopoulos:
@adrienverge I have rewritten the pull request to unfold user-visible strings. Tests have been updated accordingly. Would you consider this pull request?

adrienverge:
Thanks, you're right, it's probably better like that.

Can you put the `tests/lint/line-length.sh` change in a separated commit (for easier retrieval and understanding), and add a comment for `s/"[^"]*"/""/g` so that users can understand why we do this?
I would even break the `sed` command in two...

DimitriPapadopoulos:
Sure. How to break the _sed_ comment in two though? Using a second temporary file - I'd rather avoid that - or a simple pipe?

adrienverge:
I admit using a second temporary file sucks. Maybe a simple explanatory comment will be enough?

DimitriPapadopoulos:
I have modified the pull request. Comments welcome.

DimitriPapadopoulos:
@adrienverge Would you mind if I replaced `tests/lint/line-length.sh` with a Perl (or even Python) script? To properly handle strings, I need more elaborate code that I am unable to contribute in _awk_.

adrienverge:
OK for Python! I'd prefer not Perl because I wouldn't be able to review it.

DimitriPapadopoulos:
Good, I prefer Python too although Perl is handy for tiny scripts.

Issue #280, opened 2018-04-05
hard coded remote address in pppd parameters


mrbaseman:
In the parameters for calling pppd the address 1.1.1.1 [is hard coded](https://github.com/adrienverge/openfortivpn/blob/master/src/tunnel.c#L128) as remote address on the ppp link, see "Options" Section in [man 8 pppd](https://linux.die.net/man/8/pppd).
My understanding is that this is needed to bring up the connection until the remote side has assigned a proper IP. However, when starting `openfortivpn` a host route is added for this. It seems that this "virtual ip" has been widely used also by other vendors like [Cisco](https://supportforums.cisco.com/t5/getting-started-with-wireless/what-virtual-ip-should-i-use-on-a-cisco-wlc/td-p/3018060), but it has been assigned to [cloudflare](https://www.cloudflare.com/learning/dns/what-is-1.1.1.1/), now. Therefore, one should use a different IP from the range of
[RFC1918 private addresses](http://www.ietf.org/rfc/rfc1918.txt) or an [RFC3330 Test-Net address](https://tools.ietf.org/html/rfc3330) (such as 192.0.2.1/32) as explained for the use case of [DHCP servers](
http://www.revolutionwifi.net/revolutionwifi/2011/03/explaining-dhcp-server-1111.html). In the Cisco case also one of the [rfc5737 test nets](https://tools.ietf.org/rfc/rfc5737.txt) is recommended.
I would suggest to apply a similar change for `openfortivpn`. Does anybody know if this _can_ easily be changed on the client side, or do we have to wait for Fortigate to change it and adapt the client once the server has changed?

DimitriPapadopoulos:
[1.1.1.1 is used by Fortinet VPN as the virtual endpoint so doesn’t work when my VPN is running](https://twitter.com/chriswigginsnz/status/981315362015789056)

DimitriPapadopoulos:
Although I don't have a good understanding of the Fortinet VPN or network routing, it looks like 1.1.1.1 is just a default value and that it can be readily changed on the FortiGate device: [Howto Fortigate](https://www.shrew.net/support/Howto_Fortigate#Gateway_Configuration).

If that's the case, I'm curious how _FortiClient_ handles that. Can the gateway be queried instead of hard-coded in the client software or configuration file?

Have you tried using 1.1.1.1 as the default DNS server for your workstation together with _openfortivpn_? What happens then? Just DNS queries not working? _openfortivpn_ not working?

[...]

PR #281, opened 2018-04-05
replace hard-coded virtual ip address in pppd call parameters…

mrbaseman:
…by a rc3330 test-net address

1.1.1.1 was widely used for ppp connections in the past but it has been assigned by  IANA and is used by a public DNS server operated by Cloudflare now. Recommendation is to use a test-net address instead. This commit therefore changes it to 192.0.2.1 and solves #280

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -125,7 +125,7 @@ static int pppd_run(struct tunnel *tunnel)
 		static const char *args[] = {
 			pppd_path,
 			"38400", // speed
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
https://github.com/adrienverge/openfortivpn/blob/ce5ba40/etc/ppp/peers/openfortivpn#L2

mrbaseman:
oops, I have closed this somehow by accident

PR #282, opened 2018-04-05
replace hard-coded virtual ip address in pppd call parameters...

mrbaseman:
...by a rfc3330 test-net address

1.1.1.1 was widely used for ppp connections in the past but it has been assigned by IANA and is used by a public DNS server operated by Cloudflare now. Recommendation is to use a test-net address instead. This commit therefore changes it to 192.0.2.1 and solves #280 (this is a second attempt after #281 was closed by accident)

### DIFF
--- a/etc/ppp/peers/openfortivpn
+++ b/etc/ppp/peers/openfortivpn
@@ -1,5 +1,5 @@
 38400
-:1.1.1.1
+:192.0.2.1
 noipdefault
 noaccomp
 noauth
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -125,7 +125,7 @@ static int pppd_run(struct tunnel *tunnel)
 		static const char *args[] = {
 			pppd_path,
 			"38400", // speed
-			":1.1.1.1", // <local_IP_address>:<remote_IP_address>
+			":192.0.2.1", // <local_IP_address>:<remote_IP_address>
 			"noipdefault",
 			"noaccomp",
 			"noauth",


mrbaseman:
this is with updated `etc/ppp/peers/openfortivpn` now (I should have squashed during merge instead of messing up the pull request...). @DimitriPapadopoulos, perhaps we wait with the merge and test it from home again?

DimitriPapadopoulos:
Sure, I can test from home.

mrbaseman:
* I have tested this from work to home and from home to work. In both directions it works.
* At home I could configure 1.1.1.1 as nameserver. DNS resolution gets broken when _openfortivpn_ from current master opens a connection, but it continues to work when built with the current pull request. 
* just as a hint: I stumbled over a few formatting issues in the man page, which I would like to land as well. Please do not squash when merging, because the two commits address different things.

DimitriPapadopoulos:
I've tested from home to work, after setting the DNS server to `1.1.1.1` / `2606:4700:4700::1111`.Both with the current pull request or the previous version _openfortivpn_ starts as usual. I've added option `--no-dns` to make sure `1.1.1.1` is used once _openfortivpn_ is up and DNS still works in both cases. I'm running Ubuntu 18.04 and I'm not certain how NetworkManager configures DNS: `/etc/resolv.conf` points to my machine (`127.0.0.53`) and _dnsmasq_ handles DNS requests from there.

Anyway, your pull request doesn't break anything here.

Can I let you merge?

mrbaseman:
At home I have a `bind9` service running locally, which is normally queried and which caches the DNS responses. Therefore, I had to edit my `/etc/resolv.conf` manually for the tests, so that it contained  only `nameserver 1.1.1.1`. 
Since we didn't notice any problems in our tests, I have merged this PR now.

PR #283, opened 2018-04-06
HTTP end-of-line marker is CR LF

DimitriPapadopoulos:
HTTP is currently handled by different code implemented in src/http.c (for communication with FortiGate) and src/tunnel.c (for communication with an HTTP proxy).

The idea is to standardize so that some day we can use the same code base for all HTTP requests - or even use an external HTTP library if such a mainstream, portable and fast library exists.


DimitriPapadopoulos:
Examples of HTTP parsers:
[nodejs/http-parser](https://github.com/nodejs/http-parser) (C)
[h2o/picohttpparser](https://github.com/h2o/picohttpparser) (C)
[fukamachi/fast-http](https://github.com/fukamachi/fast-http) (Common Lisp)

DimitriPapadopoulos:
@adrienverge Would you agree with this one? I've checked it works and the benefit is two-fold:
1. standard HTTP,
2. use of standard HTTP facilitates refactoring of _openfortivpn_ so that HTTP is handled in a single place instead of two (`src/http.c` for communication with FortiGate and `src/tunnel.c` for communication with an HTTP proxy).

PR #284, opened 2018-04-06
Properly check vsnprintf() return value

DimitriPapadopoulos:
The manual page reads:
Thus, a return value of size or more means that the output was truncated.

Also there is no need to explictely add a NUL character '\0' at the
end of the formatting string. The C compiler automatically adds the
NUL character at the end of string literals, and vsnprintf() always
terminates its output with a NUL character, whether truncating or not.


DimitriPapadopoulos:
@mrbaseman Do you agree with this one? I have no doubt about the return value of _vsnprintf_(), but I'd like your feedback about removing the NUL character `'\0'`.

Issue #272, opened 2018-04-03
Wrong 'default route' added


josepmc:
The network config I have is the following (I'm randomizing the IPs):
- eth0 - main gateway, dhcp
- eth1 - has route to the VPN server

My route before connecting looks like this:
0.0.0.0    10.0.0.1 0.0.0.0 ... 100 ... eth0
**vpn address** 192.168.0.1 255.255.255.255 ... 10 ... eth1

Now, when I connect with openfortivpn (latest from master), the following route is added:
**vpn address** 10.0.0.1 255.255.255.255 ... 0 ... eth0

Which actually messes up the entire configuration and blocks any access to vpn-bound addresses until I manually delete the route. Obviously I can just create a small script to workaround this issue, but I actually though on using the persistence option that was added recently.

When killing openfortivpn, the following additional message will be shown:
`WARN:   Could not delete route to vpn server (No such process).`

In any case, wouldn't it be worthwile to add a flag to avoid this behavior?

DimitriPapadopoulos:
There are two ways to handle that:
* Try option `--half-internet-routes=1` (**not** the `--persistent`option which unless I am missing something is totally unrelated) and please report whether it works.
* Use options `--set-routes=0` and handle routes in a script launched when launching/stopping _pppd_.

I guess this is a configuration that hasn't been tested much. Does the _FortiClient_ proprietary client work any better here? What are the routes before/after?

josepmc:
I tried with half-internet-routes to no avail.

A pppd script is also an alternative, but I still believe the behavior is inherently wrong.

On 03 Apr 2018, at 10:50, Dimitri Papadopoulos Orfanos <notifications@github.com<mailto:notifications@github.com>> wrote:


There are two ways to handle that:

  *   Try option --half-internet-routes=1 (not the --persistentoption which unless I am missing something is totally unrelated) and please report whether it works.
  *   Use options --set-routes=0 and handle routes in a script launched when launching/stopping pppd.

I guess this is a configuration that hasn't been tested much. Does the FortiClient proprietary client work any better here? What are the routes before/after?

DimitriPapadopoulos:
Indeed the behaviour is probably wrong. Personally I cannot test that because I don't have a machine attached to two networks. Tight now:
* It would help to know how the proprietary _FortClient_ handles such a setup (routes before/after).
* I have no clue how easily this can be fixed because I don't have a good understanding of the problem.

mrbaseman:
[...]

PR #285, opened 2018-04-06
correctly set up route to vpn gateway

mrbaseman:
*  support vpn gateway being on a different interface than the default route (addresses #272)
*  correctly handle routes with same network address but different netmask


mrbaseman:
@DimitriPapadopoulos shall we squash and merge this one? apart from addressing the original issue the improvement of the `ipv4_get_route()` function also fixes issues when the gateway is on the same interface but reachable over a subnet route with a specific gateway, and I have elaborated a bit on the comments in the source code

DimitriPapadopoulos:
@mrbaseman I am not the best judge, this is the part of _openfortivpn_ I understand least. Yet this looks good as far as I can understand. I suggest you merge, and I'll test it over time...

PR #286, opened 2018-04-08
Improve script to find line length errors

DimitriPapadopoulos:
Move from shell/awk to Python to exclude lines that end with a string.


DimitriPapadopoulos:
@adrienverge Here is a Python line-legnth.py test to replace the Bash line-length.sh test.

Issue #46, opened 2016-01-28
Being able to connect with a --cookie parameter


fretn:
It would be nice if you are able to connect with a --cookie parameter.

I'm working on a wrapper gui for openfortivpn which does the fortinet http auth to catch the SVPNCOOKIE, it would be nice if openfortivpn could create a connection if it receives the cookie

then in ps -ef the password wouldn't show up, and the password wouldn't be exposed in a openfortivpn config file  

Ofcourse its possible to specify the password through an interactive prompt, but this is a bit complex to handle when you spawn openfortivpn in a background process

mrbaseman:
@fretn I'm just looking through old tickets. Is this still relevant? Wouldn't be an exported environment variable for the cookie be better than a --cookie command line option which again would be visible from ps -ef?

fretn:
yeah an environment variable would be better

but this is not really relevant anymore

mrbaseman:
with an environment variable it's not much effort to implement either. can you try [this branch](https://github.com/mrbaseman/openfortivpn/tree/supply_cookie)? You can set an environment variable called SVPNCOOKIE which then taken instead of performing the login procedure. I have tried to test, but with my two factor authentication and client certificate password I guess I'm simply too slow when typing in the passwords.... At least the debug output shows the correct cookie, but I get a "403 Forbidden"

fretn:
Okay, Somewhere next week I'll try this. I'll have to dig up what and how I was trying to do this, the memory is a bit fuzzy in my brain :)

DimitriPapadopoulos:
@fretn I shall close this issue for now, since it is now possible to supply a cookie using the SVPNCOOKIE environment variable. Do not hesitate to reopen if you still encounter problems.

PR #296, opened 2018-04-20
supply the SVPNCOOKIE via environment

mrbaseman:
if the SVPNCOOKIE is set in the enviroment authentication is skipped and
the cookie is taken from the environment instead. (addresses #46)

### DIFF
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -191,6 +191,14 @@ VPN_ROUTE_MASK_... the network mask for this route
 .br
 VPN_ROUTE_GATEWAY_... the gateway for the current route entry
 
+If an environment variable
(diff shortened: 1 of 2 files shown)


mrbaseman:
I have this branch hanging around for a while. Sorry for the noise, I had to push another commit to pass the lint check which I had forgotten to run locally and squashed it afterwards.

DimitriPapadopoulos:
@mrbaseman Martin, is this patch still of interest? Should we merge it?

mrbaseman:
Probably not (see [here](https://github.com/adrienverge/openfortivpn/issues/46#issuecomment-336418971)). It addresses a very old issue which is already closed.
We have added MacOSX support in the meantime and fixed many issues.

PR #298, opened 2018-04-21
Build with missing pthread_mutexattr_setrobust()

DimitriPapadopoulos:
Fixes #297

PR #299, opened 2018-04-21
Use autoconf macros in #ifdef

DimitriPapadopoulos:


mrbaseman:
@DimitriPapadopoulos I have similar changes in my [bsd_support](https://github.com/mrbaseman/openfortivpn/tree/bsd_support) branch. I have to rebase it manually because there were a couple of changes in the same code regions. I just didn't spend much time on the BSD support recently, because FreeBSD doesn't have pppd anymore, but a user space ppp utility instead, which doesn't take all parameters from the command line but from stdin... this was last missing bit, but I didn't find the time yet to solve that. Maybe I should split the branch up into several smaller bits...

mrbaseman:
Hi @DimitriPapadopoulos, I have rebased my [bsd_support](https://github.com/mrbaseman/openfortivpn/tree/bsd_support) branch on your [forkpty](https://github.com/DimitriPapadopoulos/openfortivpn/tree/forkpty) branch now. 

I have split up the changes into several smaller commits. I had to undo most of #254 in order to keep the logic simple, i.e. use `#ifdef` in preference of `#ifndef`. Some changes were already addressed by your branch or #298 already. 

For more details see my recent pull request #300

Issue #241, opened 2017-12-05
[Feature request] BSD support


mrbaseman:
In #222 we have agreed that looking at BSD would be interesting. I have moved BSD support into a  [separate branch](https://github.com/mrbaseman/openfortivpn/tree/bsd_support). I haven't yet made a pull request out of it because there are still a few things to fix, but I have decided to open this issue for discussion and documentation

mrbaseman:
with https://github.com/mrbaseman/openfortivpn/commit/91079d7397ef91a78ac38ddc130f7d8b7c559c11 it compiles on FreeBSD now, but it crashes with a core dump. So far I could determine that the `rtentry` data structures are not yet properly initialized. On FreeBSD [struct rtentry](https://www.freebsd.org/cgi/man.cgi?query=rtentry) starts with two `radix_node` structures that contain the destination address and mask and obviously it is not ok to just zero this memory area. These `rt_nodes` must be properly initialized.

mrbaseman:
one more thing I should link here: the current [README.md](https://github.com/adrienverge/openfortivpn/blob/master/README.md) on master by mistake already mentions FreeBSD support, but at least it states that this BSD port is not finished yet, see also
https://github.com/adrienverge/openfortivpn/issues/222#issuecomment-347519742

mrbaseman:
initialization of the `radix_node` structures should perhaps be done using [rtalloc1_fib](https://www.freebsd.org/cgi/man.cgi?query=rtalloc) or along the lines of the source of the [route](https://github.com/freebsd/freebsd/blob/master/sbin/route/route.c#L790) command. I'm still looking if this is feasible or if we should better use syscalls to the `route` command as a workaround, similar to the syscalls to the `netstat` command in the OSX part of the code (which I also use in the BSD port)

mrbaseman:
I have moved the line about FreeBSD in the README.md to the bsd_support branch and updated the path to netstat for BSD as well. The issue how to deal with the kernel routes on BSD is still open.

mrbaseman:
[...]

PR #300, opened 2018-04-23
FreeBSD support

mrbaseman:
I have started this work some time ago. At the current stage I have managed to configure, compile and run _openfortivpn_ on FreeBSD 11.
The last remaining issue is that FreeBSD doesn't have `pppd` anymore and uses a user-space tool called `ppp`. It doesn't take many options from the command line but rather from STDIN or from a config file. To my understanding of the [man page](https://www.freebsd.org/cgi/man.cgi?query=ppp&sektion=8) it should be possible to put all necessary parameters into `/etc/ppp/ppp.conf` and _openfortivpn_ should work with this. If this works, the only missing bit is to provide a working sample configuration and perhaps an update of the man page.
Many command line options of _openfortivpn_ concerning how `pppd` is called are currently not supported on BSD. At a later stage one can modify the way `ppp` is called and pipe setup commands into the client before sending it to the background. I just haven't figured out to do this correctly in combination with `forkpty`, but we can keep this as a separate task, once `ppp` has successfully opened a connection by feeding it with appropriate parameters from a config file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -244,18 +244,6 @@ AC_ARG_ENABLE([proc],
 			[Enable route manipulations directly via /proc/net/route \
 			when cross-compiling, use --disable-proc for the opposite]))
 
-
-# when neither ppp nor pppd are enabled, assume the previous behavior (for travis)
-AS_IF([test "x$with_ppp" = "xno" -a "x$with_pppd" = "xno"], [
-   with_pppd="yes"
-])
-
-# when both are enabled, give pppd the higher priority (we can only use one of them)
-AS_IF([test "x$with_ppp" = "xyes" -a "x$with_pppd" = "xyes"], [
-   with_ppp="no"
-])
-
-
 # Checks for existence of specific files.
 AC_CHECK_FILE([/proc/net/route],[
   AS_IF([test "x$enable_proc" = "x" ], [
@@ -291,6 +279,18 @@ AC_CHECK_FILE(/usr/sbin/pppd,[
 ],[])
 
 
+# when neither ppp nor pppd are enabled, assume the previous behavior (for travis)
+AS_IF([test "x$with_ppp" = "xno" -a "x$with_pppd" = "xno" ], [
+   with_pppd="yes"
+])
+AS_IF([test "x$with_ppp" = "x" -a "x$with_pppd" = "x" ], [
+   with_pppd="yes"
+])
+
+# when both are enabled, give pppd the higher priority (we can only use one of them)
+AS_IF([test "x$with_ppp" = "xyes" -a "x$with_pppd" = "xyes"], [
+   with_ppp="no"
+])
 
 AS_IF([test "x$with_ppp" = "xyes"], [
 	AC_DEFINE(HAVE_USR_SBIN_PPP, 1)
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 PKG_CHECK_MODULES(OPENSSL, [libcrypto >= 0.9.8 libssl >= 0.9.8], [], [AC_MSG_ERROR([Cannot find OpenSSL 0.9.8 or higher.])])
 AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthread.])])
 AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])
-PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [ ])
+PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [AC_MSG_RESULT([libsystemd not present])])
 
 # Checks for header files.
 AC_CHECK_HEADERS([arpa/inet.h assert.h ctype.h errno.h fcntl.h getopt.h ifaddrs.h limits.h netdb.h net/if.h netinet/in.h netinet/tcp.h pthread.h signal.h stdarg.h stdint.h stdio.h stdlib.h string.h sys/ioctl.h syslog.h sys/socket.h sys/stat.h sys/types.h sys/wait.h termios.h unistd.h], [], AC_MSG_ERROR([Required header not found]))
--- a/configure.ac
+++ b/configure.ac
@@ -23,14 +23,53 @@ AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthr
 AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])
 PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [AC_MSG_RESULT([libsystemd not present])])
 
-# Checks for header files.
-AC_CHECK_HEADERS([arpa/inet.h assert.h ctype.h errno.h fcntl.h getopt.h ifaddrs.h limits.h netdb.h net/if.h netinet/in.h netinet/tcp.h pthread.h signal.h stdarg.h stdint.h stdio.h stdlib.h string.h sys/ioctl.h syslog.h sys/socket.h sys/stat.h sys/types.h sys/wait.h termios.h unistd.h], [], AC_MSG_ERROR([Required header not found]))
+# we assume presence of the follwowing C standard headers
+# and ommit them in the follwoing header checks
+#
+# assert.h
+# ctype.h
+# errno.h
+# limits.h
+# signal.h
+# stdarg.h
+# stdint.h
+# stdio.h
+# stdlib.h
+# string.h
+
+# Checks for required header files.
+AC_CHECK_HEADERS([ \
+arpa/inet.h \
+fcntl.h \
+getopt.h \
+ifaddrs.h \
+netdb.h \
+net/if.h \
+netinet/in.h \
+netinet/tcp.h \
+pthread.h \
+sys/ioctl.h \
+syslog.h \
+sys/socket.h \
+sys/stat.h \
+sys/types.h \
+sys/wait.h \
+termios.h \
+unistd.h \
+], [], AC_MSG_ERROR([Required header not found]))
 
 # Checks for header files with prerequisites of other headers.
 AC_CHECK_HEADERS([net/route.h], [], AC_MSG_ERROR([Required header not found]), [#include <net/if.h>])
 
 # Checks for optional header files.
-AC_CHECK_HEADERS([libutil.h mach/mach.h pty.h semaphore.h sys/mutex.h util.h])
+AC_CHECK_HEADERS([ \
+libutil.h \
+mach/mach.h \
+pty.h \
+semaphore.h \
+sys/mutex.h \
+util.h \
+])
 
 # Checks for optional header files with prerequisites of other headers.
 AC_CHECK_HEADERS([net/if_var.h],[],[],[#include <net/if.h>])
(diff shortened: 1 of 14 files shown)


src/tunnel.c
```diff
@@ -46,7 +46,12 @@
 #elif HAVE_UTIL_H
 #include <util.h>
 #endif
+#if HAVE_LIBUTIL
+#include <libutil.h>
+#endif
+#if HAVE_TERMIOS_H
```
DimitriPapadopoulos: <termios.h> is POSIX and should be available on all supported platforms. While the #ifdef does't really hurt, perhaps we should remove it for simplicity.

DimitriPapadopoulos: Out of curiosity, what is <libutil.h> for?

mrbaseman:
I just see Travis build failed due to the missing runtime dependency of `pppd`. Maybe such dependencies should be implemented as configure options `--with-pppd=/usr/sbin/pppd` and `--with-ppp=...` etc.? Any hints from an autoconf-Guru how to do this better are welcome ;-)

src/tunnel.c
mrbaseman: since we didn't have to include termios.h explicitly on other platforms than FreeBSD so far, I have included it in the `#ifdef` section, honestly without further investigations if this header should always be there.

The reason for both headers is the [synopsis of forkpty](https://www.freebsd.org/cgi/man.cgi?query=forkpty)

DimitriPapadopoulos: Header <termios.h> is included implicitly on Linux, it just needs to be included explicitly on all platforms. Build should fail without <termios.h>. I suggest removing the `#ifdef` from the source file and have _autoconf_ fail if <termios.h> is missing.

mrbaseman:
@DimitriPapadopoulos our comments just crossed... Travis has given me the same feedback and as a quick fix I have changed the code back to default behavior when `/usr/sbin/ppp` is not available. 
But you are right, auto-detection at build-time is probably not what we want here. I'll check for better solutions, see [my comment above](https://github.com/adrienverge/openfortivpn/pull/300#issuecomment-383587953)

src/tunnel.c
DimitriPapadopoulos: Indeed it looks like FreeBSD is not compliant with recent versions of POSIX which specify _fokrpty_() is defined in `<termios.h>`.

I checked FreeBSD's [`<termios.h>`](https://github.com/freebsd/freebsd/blob/master/include/termios.h) does not define _forkpty_(), [`<libutil.h>`](https://github.com/freebsd/freebsd/blob/master/lib/libutil/libutil.h) does.

DimitriPapadopoulos:
Perhaps **build-time** is better though. No need to push cross-platform code in an executable specific to a single platform. Options `--with-pppd`/`--with-ppp` are they way to go as far as I can see:
[3. Adding Options](https://autotools.io/autoconf/arguments.html)

DimitriPapadopoulos:
While _pppd_run_() can switch between `pppd` and `ppp` _pppd_terminate_() is specific to `pppd` as far as I can see. Am I missing something?

mrbaseman:
* run-time detection of `pppd` vs. `ppp` was quite easy to implement because basically it was already there. 
* I'll check the build-time options tomorrow
* you are right, I forgot to handle this in `pppd_terminate()`. When I started to work on the BSD support, the `pppd` specific error messages weren't yet there. Probably the name of the program was already present in the error messages but just together with an exit code. I have to check the ppp man page again, for possible exit codes and what they mean
* About [required headers](https://github.com/adrienverge/openfortivpn/pull/300#discussion_r183413805), there is also [AC_CHECK_HEADER](https://github.com/adrienverge/openfortivpn/pull/300/commits/43418cc02f32f14aa3262ec60efcc47ea8ed538b), but if we want to make several ones required, we can also change the AC_CHECK_HEADERS line. However, not all of them are present on all platforms, so we need a second line with optional headers, just like we do for the functions. @DimitriPapadopoulos, in your opionion, which headers are required ones which should always be there?

mrbaseman:
I would assume that all includes that are not enclosed in an ifdef statement are required ones...

DimitriPapadopoulos:
Exactly, all headers not protected by `#ifdef HAVE_...` could be added to the list.

However do we really need to check for headers that are documented in the C standard (`<stdio.h>`, `<stdlib.h>`, ...) or have been in POSIX for years (`<unistd.h>`, `<sys/socket.h>`, ...)? That means more maintenance for no clear benefit. I would only check for headers/functions not available on some of the platforms we support (usually because they're a "recent" addition to the Single Unix Specification).

DimitriPapadopoulos:
Return values of _ppp_ are not documented.

The FreeBSD source code shows that _main_() returns or_exit_() is called inconsistently with numerical values or `EX_START`/`EX_NORMAL`/`EX_REDIAL`/`EX_RECONNECT`/`EX_...`:
* [`exit(EX_START);`](https://github.com/freebsd/freebsd/blob/f1d5ce4/usr.sbin/ppp/main.c#L201)
* [`exit(1);`](https://github.com/freebsd/freebsd/blob/f1d5ce4/usr.sbin/ppp/main.c#L298)
* [`return c;`](https://github.com/freebsd/freebsd/blob/f1d5ce4/usr.sbin/ppp/main.c#L497)
* [`return EX_NORMAL;`](https://github.com/freebsd/freebsd/blob/f1d5ce4/usr.sbin/ppp/main.c#L538)

Return values of _ppp_ are probably less informative than those of _pppd_.

DimitriPapadopoulos:
You're right, the list of header files is rather short.

The list of functions is much more difficult to get right.

mrbaseman:
concerning `ppp`'s exit codes: To a large extend [sysexits](https://www.freebsd.org/cgi/man.cgi?query=sysexits) are used and for them [err()](https://www.freebsd.org/cgi/man.cgi?query=err&sektion=3&apropos=0&manpath=FreeBSD+11.1-RELEASE+and+Ports) could be used. I'll have a look and update `pppd_terminate()` accordingly.

mrbaseman:
@DimitriPapadopoulos I think I have addressed all your comments now. Thanks for the input. 
The mechanism for detection is now:
* at configure time the existence of `/proc/net/route` is checked, if it is not available, `netstat` is searched at well-known locations. 
* Also, the existence of `ppp` is checked, and the path is stored in a preprocessor variable `PPP_PATH`. If `pppd` is found, it's location overwrites `PPP_PATH`.
* For cross-compiling there are configure-options that override the detected values. `--with-pppd` and `--with-ppp` override `PPP_PATH` and additionally set `HAVE_USR_SBIN_PPP` or `HAVE_USR_SBIN_PPPD` respectively. Depending on which one is set, it's assumed that `pppd` shall be used or `ppp` respectively.
* `--enable-proc` allows to override `HAVE_PROC_NET_ROUTE` if the first test resulted in a wrong result when cross-compiling.
* At run-time the existence of the file at `PPPD_PATH` specified at configure-time is double-checked

gbon121:
Hi Martin.
I don't have a FreeBSD installation to test on, but from the online manpage for ppp, it seems that

- trusted non-root users can invoke ppp, if an appropriate *allow users* option is added to the vpn-client stanza of */etc/ppp/ppp.conf*, or they belong to the *network* group;
- environment variables can be used to pass information to the ppp process (e.g. remote login credentials);
- if present, the */etc/ppp/ppp.linkup* script is executed when the ppp link comes up, and very likely it inherits the environment of the ppp process.

If the above conditions are met, then perhaps an easy way to handle additional routes and nameservers is to set environment variables in the (unprivileged) openfortivpn process, and use native commands in ppp.linkup, along the lines of https://github.com/adrienverge/openfortivpn/pull/270

mrbaseman:
I didn't have time to work on the BSD port in the last two weeks. I'm stuck with the problem that logging in and establishing the tunnel seems to work but the [read_thread always reads 0 bytes](https://github.com/mrbaseman/openfortivpn/blob/bsd_support/src/io.c#L208). My assumption is that the [FD_SET](https://github.com/mrbaseman/openfortivpn/blob/bsd_support/src/io.c#L185) didn't work as expected. 

With the configuration my current approach is to have a static config file that already contains the important parameters (`/etc/ppp/ppp.conf` seems to be the usual place). Once this works, the next step would be to pass parameters in a more elegant way to ppp, maybe using a [diagnostic port](https://www.freebsd.org/doc/handbook/userppp.html#idp76786760) and [pppctl](https://www.freebsd.org/cgi/man.cgi?query=pppctl&sektion=8&manpath=freebsd-release-ports). 

Maybe the problem with the file descriptor is that the [device line in the ppp config](https://github.com/mrbaseman/openfortivpn/blob/bsd_support/etc/ppp/ppp.conf.example#L6) is still wrong and ppp opens a separate tcp tunnel in parallel to the existing one. Probably I should pass the [file descriptor of the tunnel](https://github.com/mrbaseman/openfortivpn/blob/bsd_support/src/tunnel.c#L709) here. This doesn't work with a static config file, of course, but that's the current area I'm looking into. Maybe a first workaround could be a symlink in /dev with a known name which can be added to the config file... I have the feeling that I'm quite close to a working _openfortivpn_ on freebsd now.

DimitriPapadopoulos:
Perhaps you should change **pppd**, Linux-specific (part of) PPP implementation, to **ppp** in user-visible text, including `--pppd-...` options.

DimitriPapadopoulos:
You're right, in any case let's stabilize FreeBSD support and the DNS stuff first.

That said it's very easy to change `pppd` into `ppp` because there are no functional changes, just need to support both option styles for some time. By the way, do _pppd_ options on Linux match _ppp_ options on FreeBSD?

mrbaseman:
`ppp` on FreeBSD only supports a few options on the command line. Most of the connection details have to be supplied either in a config file [/etc/ppp/ppp.conf](https://www.freebsd.org/doc/handbook/userppp.html) with which I was experimenting a bit, or on stdin but I'm not sure how this works together with the pthreads that we are running at this stage. 
I was also trying to get pppd onto a current FreeBSD, just to see that everything else is working now, but I didn't succeed in that either. Probably I have to focus on debugging the communication of `ppp` with the remote side - first of all ensure it's talking on the right file descriptor and also figuring out all the config parameters needed to make it talk to a `pppd`-like service on the other end of the line.
I must admit that I didn't have much time to work on this in the past weeks.

mrbaseman:
I'm now struggling with the lint test. Somehow it behaves a bit different on my local desktop than on travis. 
Anyhow, this code works on Linux and on FreeBSD now (finally! big thanks to @nferch) 
It is a very long patch series that has started quite some time ago, but it boils down to a relatively small patch, once I squash everything. Probably, the most easy way to handle this is to start a fresh branch again and just copy the current sources there. @DimitriPapadopoulos, @adrienverge,  what do you think?

adrienverge:
@mrbaseman great job!! So it works on FreeBSD, without adding any risk that something breaks on Linux?

About the linting, your latest push seems the pass on Travis. In my opinion there is no need to start a fresh branch, if you just rebase on latest `master` and `push --force` it should do the job and tests should pass.

adrienverge:
I'm glad to hear that :)

Keep us informed on what it gives on Mac OS!

mrbaseman:
I have just compiled and tested my bsd_support branch on MacOS X. 
There it works smoothly, too. I've just fetched the latest [cleanup commits](https://github.com/adrienverge/openfortivpn/issues/241#issuecomment-432821012) by @nferch, too

DimitriPapadopoulos:
@mrbaseman Excellent work Martin!

adrienverge:
I confirm that compiling your branch (commit e66e1d3) on Fedora 28 works fine.

:+1:

mrbaseman:
Yes. I have tested the branch again on a few platforms: Ubuntu 16.04, CentOS 7.5,  SLES 12.3, and SL 6.10. On the latter I have found that configure fails due to the absence of libsystemd. 

Maybe it's worth cherry-picking 06d3241 into a separate PR? It's a fix to the autoconf system, but it's unrelated to FreeBSD and it fixes an issue which is already present on the current master. I'm not sure if "absence of libsystemd on linux" is the correct condition. Maybe also a particular version of autoconf is needed to reproduce this. (in this case it was autoconf (GNU Autoconf) 2.63)

configure.ac
```diff
@@ -17,16 +17,16 @@ AC_CANONICAL_HOST
...
 # Checks for libraries.
 PKG_CHECK_MODULES(OPENSSL, [libcrypto libssl])
 AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthread.])])
 AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])
 
 # Checks for header files.
-AC_CHECK_HEADERS([arpa/inet.h assert.h ctype.h errno.h fcntl.h getopt.h ifaddrs.h libutil.h limits.h mach/mach.h netdb.h net/if.h net/if_var.h netinet/in.h netinet/tcp.h net/route.h pty.h semaphore.h signal.h stdarg.h stddef.h stdint.h stdio.h stdlib.h string.h strings.h sys/ioctl.h sys/mutex.h syslog.h sys/socket.h sys/stat.h sys/time.h sys/types.h sys/wait.h unistd.h util.h])
-# fail without this one:
-AC_CHECK_HEADER([termios.h], [], AC_MSG_ERROR([Required header not present]))
+AC_CHECK_HEADERS([arpa/inet.h assert.h ctype.h errno.h fcntl.h getopt.h ifaddrs.h limits.h netdb.h net/if.h netinet/in.h netinet/tcp.h net/route.h pthread.h signal.h stdarg.h stdint.h stdio.h stdlib.h string.h sys/ioctl.h syslog.h sys/socket.h sys/stat.h sys/types.h sys/wait.h termios.h unistd.h], [], AC_MSG_ERROR([Required header not found]))
```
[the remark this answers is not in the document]
mrbaseman: I would prefer the complete list of headers. It's easier to check with a `grep -r '#include <' | awk -F\: '{print $2}' | sort -u`. They should appear either in the required ones or in the optional ones... BTW. `systemd/sd-daemon.h` should be added to the optional headers, then.

Another approach would be to reduce those lists and keep only the ones for which we actually have an `#if` statement, but I prefer having the HAVE_HEADER_H defined so that one can easily add the `#if` statement once it turns out to be necessary on some platform.

I must admit that those long lines in configure.ac are very difficult to review when changes happen there. Maybe we should write just one item on a line and mask the newlines with a backslash? Then it's easier to spot a single header that's inserted or removed, because it's always a line on its own.

DimitriPapadopoulos:
Indeed a separate PR would be better. Please pull it directly.

I seem to recall I tried on a Linux machine without _systemd_ and that it worked for me: build didn't fail, simply `HAVE_SYSTEMD` wasn't defined. I may be wrong though, so your patch is welcome.

configure.ac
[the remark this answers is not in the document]
DimitriPapadopoulos: I tend to avoid at least the headers that are part of the C standard, but I'm OK with your suggestion. Perhaps an item per line would be easier to maintain, but I have no strong opinion, do as you wish.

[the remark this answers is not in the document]
mrbaseman: I have taken the C standard headers out of the check and put each one on a separate line.

mrbaseman:
shall we merge this now?

adrienverge:
OK for me, I trust you and Dimitri's review!

mrbaseman:
There are overrides for almost all of them. If `/proc/net/route` is not available on the build system, it can be enabled via `--enable-proc`, the location of `netstat` can be controlled with `--with-netstat`, and the ppp program can be adjusted by `--with-pppd` or `--with-ppp` respectively. 
Ah, there is no `--disable-proc` for cross-compiling for FreeBSD on Linux.... Ok, I'll check that. 
With "tested at run time" you mean we should check if the program that we call really exists and is executable?  We did for `pppd` in the past and it's done for `ppp` also, but it wasn't done so far for `netstat` and for `route` for instance, but I can add such tests, too. 
So, the agenda is:
* implement a `--disable-proc` configure option
* add checks like `if (access(pppd_path, F_OK) != 0)` for the other external binaries
* repeat testing
* squash and merge :-)

DimitriPapadopoulos:
Yes, adding the run time tests you suggest could fix some edge cases (production machine very different from build machine).

I would even suggest getting rid of `AC_CHECK_FILE` in `configure.ac`. Finding the proper executables initially looked like a run time task to me, not a task that can be automated at build time based on characteristics of the build machine that may differ from the production machine. On the other hand testing multiple paths to find an executable file, instead of a single path, may be seen as a security issue (larger attack surface). So I would eventually stick to a single path for each executable on Linux, FreeBSD and macOS, and let users override the path at build time if needed. Which is almost what you do: I was just arguing that it's perhaps better to hard-wire the default executable path for each platform instead of relying on `AC_CHECK_FILE`. The override remains useful of course and should be added for all hard-wired paths.

I don't know if this makes sense. What do you think? Either way the current situation is OK with me if you'd rather stay with `AC_CHECK_FILE`. I have no strong opinion on this issue.

mrbaseman:
I would like to stay with the configure mechanism. We don't have a reliable mechanism to detect a specific platform. At compile time we used to have some `#ifdef __APPLE__` but large parts of this code are re-used on BSD, and there we don't have this macro definition. With every new platform the decisions would become more complex, and apart from that it's not at runtime anyway. Also, the configure mechanism might work out of the box on other platforms that are similar to the ones that we already support.

I have re-worked the override options now, because they were not working in both directions. I have added comments in `configure.ac` and to the commit message, where I have explained the relevant details about autoconf. Basically, the order in which the statements are put into the  file is not necessarily the order in which they are executed, at least that's my interpretation of what I have observed in various tests of the override options.

I have also added a few runtime checks for the compiled in paths. I have also hard-coded the absolute path to the `route` command on OSX and BSD, while being at it. I believe we can't do such a check on the special file `/proc/net/route` but there is already a check to handle failure of the `open()` command.

mrbaseman:
I have re-tested with the current state on Ubuntu 16.04, CentOS 7, SL 6, SLES 12, FreeBSD 11, OSX 10.11. Building works fine on all of them, as client I could test Ubuntu, OSX, and FreeBSD. I'd merge it now. I it turns out that it needs further adjustment, then we can improve the code in separate pull requests.

DimitriPapadopoulos:
Agreed.

adrienverge:
:+1:

PR #303, opened 2018-05-11
Rework OpenSSL library detection

tadeboro:
Currently, X509_check_host function is always marked as missing, since
AC_CHECK_FUNCS is missing linker flags for libssl and libcrypto.

Culprit for this is the PKG_CHECK_MODULES call that does not add
libraries to the link arguments that configure uses to create function
tests, which in turn makes function check fail.

To remedy the situation, this commit replaces PKG_CHECK_MODULES with
two AC_CHECK_LIB calls (one for libcrypto and one for libssl) that
properly populate configure linker flags.


DimitriPapadopoulos:
I had to revert to keep taking into account `PKG_CONFIG_PATH` on macOS.

PR #306, opened 2018-05-13
Require OpenSSL 1.0.2

DimitriPapadopoulos:
OpenSSL 1.0.2 provides X509_check_host so we don't have to explictly
check its existence with AC_CHECK_FUNCS, which in turn simplifies the
build infrstructure and the code base.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,7 +18,7 @@ AC_CANONICAL_HOST
 AM_SILENT_RULES([yes])
 
 # Checks for libraries.
-PKG_CHECK_MODULES(OPENSSL, [libcrypto libssl])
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
Fails despite:
```
dist: xenial
```
Strange, _xenial_ [ships with OpenSSL 1.0.2](https://packages.ubuntu.com/source/xenial/openssl).

DimitriPapadopoulos:
While OpenSSL 0.9.8 < 1.0.2 is not maintained any more, some Linux distributions based on OpenSSL 0.9.8 have still not reached EOL:
* CentOS 6 ([EOL in on 30 November 2020](https://wiki.centos.org/About/Product))
* SLES 11 SP4 ([EOL on 31 March 2022](https://www.suse.com/lifecycle/))

Let's not depend on OpenSSL >= 1.0.2 for now.

PR #307, opened 2018-05-13
Use PKG_CHECK_MODULES compiler/linker flags

### LINKED ISSUE #304: X509_check_host detection always fails
Configuration process fails to detect X509_check_host function even on systems that have it defined.

## Steps to reproduce

 1. Ensure we have proper verison of OpenSSL using `pkg-config --modversion libssl` (should return `1.1.0` or `1.0.2`).
 2. Clone the repository.
 3. Run `./autogen.sh`.
 4. Run `./configure.sh`.
 5. One of the last lines in the output is `checking for X509_check_host... no`.

This was tested on Fedora 27 (OpenSSL 1.1.0h) and unstable Gentoo (OpenSSL 1.0.2o).

## What is causing the issue

Main reason for this issue is the fact that PKG_CHECK_MODULES does not add proper linker flags to the set of flags that configure is using to check for function presence. This can be seen in the config.log file:

    configure:5102: checking for X509_check_host
    configure:5102: gcc -o conftest -g -O2   conftest.c -lutil -lpthread  >&5
    /tmp/cckiBJJz.o: In function `main':
    /home/tadej/devel/openfortivpn/conftest.c:184: undefined reference to `X509_check_host'
    collect2: error: ld returned 1 exit status

In the second line, `-lssl` and `-lcrypto` linker flags are missing.

## How to fix this

Initial stab at fixing this issue in in pull request #303

DimitriPapadopoulos:
The OPENSSL_CFLAGS and OPENSSL_LIBS flags were taken into account in
Makefile.am, but not in configure.ac when testing for X509_check_host.

Fixes #304.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,7 +18,7 @@ AC_CANONICAL_HOST
 AM_SILENT_RULES([yes])
 
 # Checks for libraries.
-PKG_CHECK_MODULES(OPENSSL, [libcrypto libssl])
+PKG_CHECK_MODULES(OPENSSL, [libcrypto >= 0.9.8 libssl >= 0.9.8], [], [AC_MSG_ERROR([Cannot find OpenSSL 0.9.8 or higher.])])
 AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthread.])])
 AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])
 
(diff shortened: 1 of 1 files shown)


tadeboro:
I think it would be better to use #306 instead of this pull request, since OpenSSL < 1.0.2 is not maintained anymore.

configure.ac
```diff
@@ -18,7 +18,7 @@ AC_CANONICAL_HOST
 AM_SILENT_RULES([yes])
 
 # Checks for libraries.
-PKG_CHECK_MODULES(OPENSSL, [libcrypto libssl])
+PKG_CHECK_MODULES(OPENSSL, [libcrypto >= 0.9.8 libssl >= 0.9.8], [], [AC_MSG_ERROR([Cannot find OpenSSL 0.9.8 or better.])])
```
mrbaseman: maybe "0.9.8 or higher" sounds better? apart from that it looks good to me. I didn't test if it actually solves #304 but I assume you have already verified that

Issue #230, opened 2017-11-08
bug: --no-dns and set-dns=0 not working


skuti-is:
/etc/resolv.conf file is updated when starting vpn.
With --no-dns on command line and set-dns=0 in config file.

Ubuntu 16.04.3 xenial
Kernel 4.4.0-96-generic x86_64
INFO:   Got addresses: [10.250.255.1], ns [192.168.170.5, 192.168.1.99]

openfortivpn v1.5.0 3af718e

DimitriPapadopoulos:
It''s probably not _openfortivpn_ modifying this file, see [tunnel.c](https://github.com/adrienverge/openfortivpn/blob/master/src/tunnel.c#L69):
```
	if (tunnel->config->set_dns) {
		log_info("Adding VPN nameservers...\n");
		ipv4_add_nameservers_to_resolv_conf(tunnel);
	}
```
Do you see the "_Adding VPN nameservers..._" message?

Rather it's probably _pppd_ modifying the file, see option `--pppd-no-peerdns` in [main.c](https://github.com/adrienverge/openfortivpn/blob/master/src/main.c#L87).

What happens if you add option `--pppd-no-peerdns`?

Not sure if and why we need two distinct options at this point, and I do understand this can be confusing. I suggest we first focus on understanding what happens, then we can discuss/modify that if needed.

jeremyj:
I agree with you since I do not see "_Adding VPN nameservers..._"  and with `--pppd-no-peerdns` it works.
Thanks

DimitriPapadopoulos:
OK, so we need to:
* understand why DNS is handled in two places (_openfortivpn_ and _pppd_) and why there are two associated options,
* somehow merge these options/features, or not,
* document the relation between `--no-dns` and `--pppd-no-peerdns`.

skuti-is:
updated vpn config file and this works like a charm
set-dns = 0
pppd-use-peerdns = 0

DimitriPapadopoulos:
@adrienverge and @mrbaseman Do you know why `/etc/resolv.conf` can be modified from two different subsystems (_openfortivpn_ itself and _pppd_)?

adrienverge:
Sorry, I have no idea...

mrbaseman:
[...]

PR #326, opened 2018-05-27
Let pppd handle DNS servers

DimitriPapadopoulos:
From now on _openfortivpn_ will not attempt to modify DNS settings and `/etc/resov.conf` by default. Instead leave it to _pppd_ to handle DNS (or not if `--pppd-no-peerdns`).

* Option `set-dns` is now disabled by default.
* Remove `--no-dns` which is now useless.

This is a first step towards entirely getting rid of `--set-dns` as suggested in #230 and #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Examples
 
 * Don't set IP routes and don't add VPN nameservers to `/etc/resolv.conf`:
   ```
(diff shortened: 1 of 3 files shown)


DimitriPapadopoulos:
@mrbaseman Martin, what about attempting to set `set-dns` to 0 in the next version? Works well for me and is a first step towards letting _pppd_, _openresolv_ or user-provided scripts handle DNS.

mrbaseman:
sure, let's follow this path

PR #330, opened 2018-06-09
Rationalize DNS options

### LINKED ISSUE #328: plasma-nm and not obtaining DNS servers
Hi,

I'm using NetworkManager with plasma-nm and openfortivpn.

Have configured openfortivpn through plasma-nm. It connects nicely (might want to add a note that the port is required in the URL), but so far I've been unable to ignore the DNS servers provided by DHCP.

On CLI (--pppd-no-peerdns) or via config file (pppd-use-peerdns = 0) the servers aren't replaced in /etc/resolv.conf. When using plasma-nm I don't see such options, but under IPv4 tab you can set it to Automatic or to 'Automatic (addresses only)'. The latter is normally used to not configure DNS.

It does configure the DNS unfortunately.

Would also be nice if the CLI parameters match the config ones (pppd-no-peerdns vs pppd-use-peerdns).

Thanks in advance and huge thanks for the client :).

DimitriPapadopoulos:
* The `--set-dns`/`set-dns` option enables either of _pppd_/_openfortivpn_ to get DNS settings from peer _ppp_/FortiGate server and rewrite `/etc/resolv.conf`.
* The `--pppd-no-peerdns`/`pppd-use-peerdns` option switches the above functionality from _pppd_ to _openfortivpn_. This option is not documented any more.

The difference with the previous behaviour is that:
* `--set-dns=0`/`set-dns=0` disables DNS handling in both _pppd_ and _openfortivpn_ instead of _openfortivpn_ only. A single option is now sufficient to avoid messing with `/etc/resolv.conf`.
* `--pppd-no-peerdns`/`pppd-use-peerdns=0` switches DNS handling from _pppd_ to _openfortivpn_. It didn't make sense to have both programs compete to modify `/etc/resolv.conf`, did it? Feels much better this way.

Fixes #328.

This is also a first step towards entirely getting rid of one of the DNS options as suggested in #230, #101 and #326. Future steps might include:
* Remove code that handles DNS changes and modifies `/etc/resolv.conf` directly in _openfortivpn_.
* Instead always delegate DNS handling to _pppd_.
* Remove `--pppd-no-peerdns`/`pppd-use-peerdns` options.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -94,7 +94,7 @@ static int on_ppp_if_up(struct tunnel *tunnel)
 		}
 	}
 
-	if (tunnel->config->set_dns) {
+	if (tunnel->config->set_dns && !tunnel->config->pppd_use_peerdns) {
 		log_info("Adding VPN nameservers...\n");
 		ipv4_add_nameservers_to_resolv_conf(tunnel);
 	}
@@ -113,7 +113,7 @@ static int on_ppp_if_down(struct tunnel *tunnel)
 		ipv4_restore_routes(tunnel);
 	}
 
-	if (tunnel->config->set_dns) {
+	if (tunnel->config->set_dns && !tunnel->config->pppd_use_peerdns) {
 		log_info("Removing VPN nameservers...\n");
 		ipv4_del_nameservers_from_resolv_conf(tunnel);
 	}
@@ -176,7 +176,7 @@ static int pppd_run(struct tunnel *tunnel)
 				ofv_append_varr(&pppd_args, v[i]);
 		}
 
-		if (tunnel->config->pppd_use_peerdns)
+		if (tunnel->config->set_dns && tunnel->config->pppd_use_peerdns)
 			ofv_append_varr(&pppd_args, "usepeerdns");
 		if (tunnel->config->pppd_log) {
 			ofv_append_varr(&pppd_args, "debug");
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Examples
 
 * Don't set IP routes and don't add VPN nameservers to `/etc/resolv.conf`:
   ```
-  openfortivpn vpn-gateway:8443 -u foo -p bar --no-routes --no-dns --pppd-no-peerdns
+  openfortivpn vpn-gateway:8443 -u foo -p bar --no-routes --no-dns
   ```
 * Using a config file:
   ```
@@ -39,7 +39,6 @@ Examples
   password = bar
   set-routes = 0
   set-dns = 0
-  pppd-use-peerdns = 0
   # X509 certificate sha256 sum, trust only this one!
   trusted-cert = e46d4aff08ba6914e64daa85bc6112a422fa7ce16631bff0b592a28556f993db
   ```
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -115,8 +115,8 @@ $ openssl s_client -connect \fI<host:port>\fR
 (default: HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4)
 .TP
 \fB\-\-pppd-no-peerdns\fR
-Do not ask peer ppp server for DNS server addresses and do not make pppd
-rewrite /etc/resolv.conf.
+Deprecated; do not ask peer ppp server for DNS server addresses and do not make
+pppd rewrite /etc/resolv.conf.
 .TP
 \fB\-\-pppd-log=\fI<file>\fR
 Set pppd in debug mode and save its logs into \fI<file>\fR.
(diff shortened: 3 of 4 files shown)


DimitriPapadopoulos:
This pull request breaks option compatibility, in the sense my current settings with _openfortivpn_ 1.7.0:
`--set-dns=0`/`--no-dns`/`set-dns=0`
`pppd-use-peerdns=1`
will need to be changed to:
`--set-dns=1`/`set-dns=1`
`pppd-use-peerdns=1`

That's very unfortunate, but then the previous situation was a mess and I feel it cannot be fixed without breaking compatibility.

DimitriPapadopoulos:
Indeed the case of the `set-dns=0`/`pppd-use-peerdns=1` combination is a clear change. But there's also the default case of `set-dns=1`/`pppd-use-peerdns=1` where _openfortivpn_ and _pppd_ somehow used to compete to change `/etc/resolv.conf` - not sure how it was supposed to work in practice.

Also there's indeed a discrepancy between the `--pppd-no-peerdns` command line option and the `pppd-use-peerdns` configuration file parameter, as reported in #328 for example.

Note that I haven't removed `--pppd-no-peerdns`/`pppd-use-peerdns` for now - just their documentation. This looks like your 2nd suggestion. Your 1st suggestion would be to change the `--pppd-no-peerdns`/`pppd-use-peerdns` pair to a `--pppd-use-peerdns`/`pppd-use-peerdns` pair, with the benefit that the command line option changes and explicitly breaks command line compatibility - however the configuration file option does not change. A 3rd suggestion would be to change to `--pppd-set-peerdns`/`pppd-set-peerdns` (_set_ instead of _use_) so that compatibility breaks both for command line and in the configuration file.

The main problem I have is that I'm not 100% sure we can rely solely on _pppd_. Of course _pdpd_ does work for me, but what if there are corner cases where it doesn't? Should we keep the possibility of falling back to _openfortivpn_? And if so, shouldn't we have multiple strategies? For example:
* have _openfortivpn_ directly modify `/etc/resolv.conf`,
* let _openresolv_ handle DNS as suggested in #101.

If so we need an alternate option, for example `set-dns=HANDLER` where `HANDLER` could be `pppd`, `openresolv`, `openfortivpn`, etc.

mrbaseman:
I just wanted to point out once more that I appreciate the cleanup. I just wouldn't remove documentation of an existing feature in the first step. I would:

* apply the changes to [src/tunnel.c](https://github.com/adrienverge/openfortivpn/pull/330/files#diff-5df244ea721353ca452473c7213d3f27) and mark `--pppd-no-peerdns/pppd-use-peerdns` as deprecated
* in a next step maybe inplement `set-dns=HANDLER` where `HANDLER=0` means don't change dns resolution and  `HANDLER=1` means the default handler, which in most cases is `pppd`, on FreeBSD it might be `ppp` hoping that I find enough time to work on that project
* in a third step remove all `--pppd-no-peerdns/pppd-use-peerdns` options

DimitriPapadopoulos:
Totally agree with your proposal. I have been trying to find the less disruptive path and your proposal sounds good. I'll put back the documentation of `--pppd-no-peerdns`/`pppd-use-peerdns` but modify it to explain the new behaviour and mark it as **deprecated**.

DimitriPapadopoulos:
@mrbaseman I've refactored the patch as suggested. Would you mind having a look at the pull request?

src/tunnel.c
```diff
@@ -176,7 +176,7 @@ static int pppd_run(struct tunnel *tunnel)
 				ofv_append_varr(&pppd_args, v[i]);
 		}
 
-		if (tunnel->config->pppd_use_peerdns)
+		if (tunnel->config->set_dns && tunnel->config->pppd_use_peerdns)
```
mrbaseman: maybe we should just change this to `if (tunnel->config->set_dns)` to address #349

mrbaseman: correcting myself: this would fix the current problem but leave us with a behavior that is even more confusing than it was before (it would send `usepeerdns` always, even when `pppd-use-peerdns=0` is configured),  [reverting](https://github.com/adrienverge/openfortivpn/issues/349#issuecomment-444470569) commit f383084 is probably the best solution

mrbaseman:
I think I have found what went wrong last year: We don't parse the DNS server addresses from the xml config, see [my comment here](https://github.com/adrienverge/openfortivpn/pull/483#discussion_r338410444). 
Also, not sure if [parse_config](https://github.com/adrienverge/openfortivpn/blob/1bea0850958bb4b5fb50e19eb4d040db154db0df/src/http.c#L737) is still of any relevance (I guess it has been superseded by the xml variant), but neither dns-suffix, nor the assigned address nor dns servers are retrieved in this fallback function.

PR #333, opened 2018-06-13
update README.md for MacOS X

mrbaseman:


mrbaseman:
@adrienverge @DimitriPapadopoulos Finally I have a Mac available to test and build on OS X by myself

README.md
```diff
@@ -76,7 +76,12 @@ For other distros, you'll need to build and install from source:
     /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 
     # Install Dependencies
-    brew install automake autoconf openssl@1.0
+    brew install automake autoconf openssl@1.0 pkg-config
+
+    # You may neet to make this openssl available to compilers
+    echo 'export LDFLAGS="-L/usr/local/opt/openssl/lib $LDFLAGS"' >> ~/.bash_profile
+    echo 'export CPPLAGS="-I/usr/local/opt/openssl/include $CPPFLAGS"' >> ~/.bash_profile
+
```
adrienverge: This empty line is not strictly needed I think

DimitriPapadopoulos: Also I wouldn't refer to `~/.bash_profile` when setting environment variables. Just show how to set the variables from a shell:
```
export LDFLAGS="-L/usr/local/opt/openssl/lib $LDFLAGS
export CPPLAGS="-I/usr/local/opt/openssl/include $CPPFLAGS
```
There are many way to set environment variables permanently. One of the best ways on Linux at least is in `~/.profile` so that it works with Ubuntu's DASH shell as well. Other prefer `~/.bashrc` which is an heresy of course!

adrienverge: Agree with `export LDFLAGS` instead of `.bash_profile`.

Is `CPPLAGS` needed on Mac OS to compile C (not C++)?

mrbaseman: It is needed by the preprocessor to find `openssl/ssl.h`, note that it should read `CPPFLAGS`.  Agreed to just mention the exports and leave it up to the user if and how to make this permanent. I'll push a commit that addresses all these points.

Issue #129, opened 2017-05-24
Works with CLI arg for host, not in config


rask:
When I define a gateway on the CLI as such:

    $ openfortivpn hostname.tld:443

The client connects properly.

But if I use the config file and define the same in there

    host = hostname.tld
    port = 443

The client timeouts on any connection attempts.

mrbaseman:
strange, I use the config file and it works for me. Can you run openfortivpn with option -v to see if it tries to read the config in the right place and what it loads from there? 

I also have a line for the username in my config. I think you need this also.

rask:
Yes, I have username, password, host, port, and the trusted-cert. Verbose output shows the config is loaded correctly with all of those, with the addition of the default empty `realm`.

    $ openfortivpn --version
    > 1.3.1

Checked out the latest stable(?) tag from source repo and bult from that. Let me know if a dev branch has a fix for this.

mrbaseman:
1.3.1 was tagged the day before yesterday, so it is nearly up to date with master (except for some changes to the README.md file)

To come back to your problem: openfortivpn takes the host:port tuple from the command line and overwrites the one which it has read from the config file before. Later it operates on the same data structures. So, if it works from the command line and not out of the file, it might possibly be a strange white space character heading or trailing the hostname or port. openfortivpn removes blanks and tab characters and treats newlines as line separator. If you for some reason have dos line endings, openfortivpn might take the carriage return as part of the FQHN.  
You could inspect your config file with a hex editor or rewrite a clean text file (maybe by echoing line by line and appending them). 
I hope this helps, otherwise I have to try if I can reproduce this somehow. If I can reproduce it, I can try to make the parser more robust.

rask:
I'll have to check the file more thoroughly. Thanks!

mrbaseman:
@rask did you find out what was causing this issue?

rask:
Sorry, been less reliant on VPN lately so have not had any time or chance to look into this any further. I'll add a reminder in case I manage to find a moment to investigate.

jlchatha:
I was able to reproduce and correct (white space issue).

mrbaseman:
@jlchatha So my suspicion was in the right direction. Could you give more details, or, since you write that you could correct the white space issue, prepare a pull request? That would be great, thanks in advance

DimitriPapadopoulos:
[...]

PR #334, opened 2018-06-13
treat carriage returns as white space (might solve #129)

mrbaseman:
carriage returns (that might have come in from copy&paste out of dos/win files) can lead to a damaged config file

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -135,19 +135,21 @@ int load_config(struct vpn_config *cfg, const char *filename)
 		val = equals + 1;
 
 		// Remove heading spaces
-		while (key[0] != '\0' && (key[0] == ' ' || key[0] == '\t'))
+		while (key[0] != '\0'
+		       && (key[0] == ' ' || key[0] == '\t' || key[0] == '\r'))
 			key++;
-		while (val[0] != '\0' && (val[0] == ' ' || val[0] == '\t'))
+		while (val[0] != '\0'
+		       && (val[0] == ' ' || val[0] == '\t' || val[0] == '\r'))
 			val++;
 		// Remove trailing spaces
 		for (i = strlen(key) - 1; i > 0; i--) {
-			if (key[i] == ' ' || key[i] == '\t')
+			if (key[i] == ' ' || key[i] == '\t' || key[i] == '\r')
 				key[i] = '\0';
 			else
 				break;
 		}
 		for (i = strlen(val) - 1; i > 0; i--) {
-			if (val[i] == ' ' || val[i] == '\t')
+			if (val[i] == ' ' || val[i] == '\t' || val[i] == '\r')
 				val[i] = '\0';
 			else
 				break;
--- a/src/config.c
+++ b/src/config.c
@@ -135,21 +135,19 @@ int load_config(struct vpn_config *cfg, const char *filename)
 		val = equals + 1;
 
 		// Remove heading spaces
-		while (key[0] != '\0'
-		       && (key[0] == ' ' || key[0] == '\t' || key[0] == '\r'))
+		while (iswhitespace_like(key[0]))
 			key++;
-		while (val[0] != '\0'
-		       && (val[0] == ' ' || val[0] == '\t' || val[0] == '\r'))
+		while (iswhitespace_like(val[0]))
 			val++;
 		// Remove trailing spaces
 		for (i = strlen(key) - 1; i > 0; i--) {
-			if (key[i] == ' ' || key[i] == '\t' || key[i] == '\r')
+			if (iswhitespace_like(key[i]))
 				key[i] = '\0';
 			else
 				break;
 		}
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
Ah... Good catch!

Just a short question though. The DOS/Windows new line is `\r\n` compared to `\n` on Unix. Therefore you do need to discard a trailing `\r`but why would you want to discard heading ` \r`? This should never happen in any normal situation - meaning the probability of having a heading \r` is not any larger than having other non-printable ASCII characters (NUL, EOT, BEL, VT, FF, FS, GS, ...) elsewhere in the configuration file.

mrbaseman:
if you cat the config file together from several pieces, e.g. a config template which happens to have dos line endings and you append some lines with `echo "something" >> config` you end up with a mixture of line endings. If you then join two lines, the `\r` may end anywhere in the file. The result probably also depends on what the editor thinks about the format of the file. If we treat the carriage returns as white space I think it makes sense to do so everywhere, just like tabs and space characters.

mrbaseman:
@DimitriPapadopoulos shall we check for other non-printable ASCII characters, too?

DimitriPapadopoulos:
I wouldn't check for other characters. Thinking about it, we've already gone beyond the call of duty: I don't think other software do that, except perhaps specialized software such as some text editors. For example _awk_ fails (very) miserably on simple text files with DOS/Windows newlines.

DimitriPapadopoulos:
Actually I would rather see more low-level functions remove `\r`when reading the configuration file. But that would be a much more involved change!

DimitriPapadopoulos:
Perhaps an _iswhitespace_-like static function or macro would help clarify the existing code.

mrbaseman:
you mean like [this](https://github.com/adrienverge/openfortivpn/pull/334/commits/20fc7e8587a0ac3a6e1659304826a1352220c19a)?

DimitriPapadopoulos:
Looks good, just remove the `#endif` though. Perhaps you should use an inline function, in modern C compilers it should be as fast a macro and actually recommended.
[6.43 An Inline Function is As Fast As a Macro](https://gcc.gnu.org/onlinedocs/gcc/Inline.html)
[PRE00-C. Prefer inline or static functions to function-like macros](https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152416)

adrienverge:
It could worth adding other unicode whitespace characters as well?
http://jkorpela.fi/chars/spaces.html

DimitriPapadopoulos:
It's as you wish, but at I said I feel we've already gone beyond the call of duty. I don't know of any software that would go to such lengths to discard unexpected non-printable characters when reading a configuration file.

Also we should limit ourselves to ASCII. The encoding of the configuration is not defined, so it is expected to be ASCII, not UTF-8 or anything else. Do we need non-ASCII characters?

adrienverge:
Agreed that this is already much better as is! So it's fine to if you merge like that :+1: 

I know that strange whitespace characters can be copied-pasted in text files, sadly it happens sometimes. Should we support unicode and and detect it? Maybe not, you're right.

mrbaseman:
I was thinking of an array, too, or `if(character > 0) && (character <= 32))`, but I agree with you @DimitriPapadopoulos, we have already covered the most probable cases.

@adrienverge, with UTF-8 we would have to change really [a lot of things](https://www.cprogramming.com/tutorial/unicode.html) and when handling UTF-8 one should avoid implementing this on the low level, and rather use [iwwspace](http://pubs.opengroup.org/onlinepubs/9699919799/functions/iswspace.html) for instance

adrienverge:
OK!

DimitriPapadopoulos:
Not sure `const` is useful here by the way, `char` is OK (this is not a pointer). Also it just occurred to me that most _is_-like functions expect and return `int`:
https://linux.die.net/man/3/isspace

DimitriPapadopoulos:
Perhaps _isspace()_ is actually enough here, except it doesn't handle `\0`.

DimitriPapadopoulos:
Perhaps the _isspace()_ and _iscntrl()_ combination is what we need.

mrbaseman:
@DimitriPapadopoulos yes, you are right, the `const` is obsolete here. When calling by reference or with a pointer it would give a hint to the compiler that we may not change the value inside of the inline function. I already had merged this branch, but I can create a new PR.

Actually, `isspace()` does what we need. The additional check `character != '\0'` is obsolete anyway. If `character` is one of the space characters, it is already not `'\0'`. Note, that I have changed the code a little bit when I have introduced this function. The additional check was not there when removing the trailing white spaces, but going backwards through a null terminated string one should never hit a `'\0'` character. But also when removing the heading white spaces, as long as we find one of `' ', '\t', '\r'` it's anyway not the terminating `'\0'`. 

After this analysis, I'm going to open another pull request for switching to `isspace()`.
I'll open another pull request

PR #340, opened 2018-07-03
Fix detection of HTTP chunked encoding

stesie:
In `src/http.c:154` the `find_header` function is called with payload `"Transfer-Encoding: chunked"` to detect whether chunked encoding is used.  If the server sends this exact string in the result header, then `find_header` fails to find the header as `line_len` equals `head_len` in that case.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -101,7 +101,7 @@ static const char *find_header(const char *res, const char *header)
 		int line_len = (char *) memmem(line, BUFSZ, "\r\n", 2) - line;
 		int head_len = strlen(header);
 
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Ah indeed can reproduced with this piece of code:
```
int main() {
    const char *result = find_header("Transfer-Encoding: chunked\r\n\r\n",
                                     "Transfer-Encoding: chunked");
    if (result)
        printf("[%s]\n", result);
    else
        printf("NULL\n");
}
```
I'm wondering: how to start add unit tests for such small functions that are usually static?

PR #342, opened 2018-07-04
Increase buffer size  in ipv4_get_route to 64k, add sanity check for overflows in MacOS

DirtyHairy:
On MacOS, `ipv4_get_route` gets stuck in an infinite loop if the buffer is too small. This patch increases buffer size to 64k and adds a sanity check that raises an error instead of locking up.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -93,12 +93,18 @@ int http_send(struct tunnel *tunnel, const char *request, ...)
 	return 1;
 }
 
-static const char *find_header(const char *res, const char *header, uint32_t response_size)
+static const char *find_header(
+        const char *res,
+        const char *header,
+        uint32_t response_size
+)
 {
 	const char *line = res;
 
 	while (memcmp(line, "\r\n", 2)) {
-		int line_len = (char *) memmem(line, response_size - (line - res), "\r\n", 2) - line;
+		int line_len = (char *) memmem(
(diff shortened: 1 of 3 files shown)


DirtyHairy:
Hi @DimitriPapadopoulos !

Yes, definitely, the same problem applies to Linux, too. I didn't touch the code for Linux as I have no quick way of testing it, though. An even better solution would be to grow the buffer with `realloc` as needed. In particular on MacOS, the routing table can be pretty big, as it contains (among other things) ARP lookup results. I can implement this if you like, but I cannot test on linux :smirk:

Cheers
    -Christian

mrbaseman:
a mechanism with `realloc` would be nice. One could allocate more space in chunks of `0x1000` when needed. 
We can test on Linux, and at least the travis checks to compile `openfortivpn` for each pull request run on linux as well.
Testing on Mac OS X is more a problem, but since recently, I also have an El Capitan available for testing.

DirtyHairy:
I am on High Sierra, so I can test there. The change is pretty much agnostic of the precise MacOS version though, so I think this is less of an issue.

DirtyHairy:
I have changed the PR to realloc the buffer in chunks of 4k if it is too small. It builds on Linux, but some testing is still required.

While building the code with asan enabled (in order to check for memory corruption in my code), I found a heap buffer overflow in `find_header` that I took the liberty to fix.

PR #344, opened 2018-07-09
gcc 8: fix -Wstring-truncation warnings

martinetd:
New warnings introduced with recent of version of GCC need ignoring
or fixing.
This is an attempt to fix them.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -485,7 +485,7 @@ int auth_log_in(struct tunnel *tunnel)
 	char reqid[32] = { '\0' };
 	char polid[32] = { '\0' };
 	char group[128] = { '\0' };
-	char data[256], token[128], tokenresponse[256];
+	char data[1024], token[128], tokenresponse[256];
 	char *res = NULL;
 
 	url_encode(username, tunnel->config->username);
@@ -494,7 +494,7 @@ int auth_log_in(struct tunnel *tunnel)
 
 	tunnel->config->cookie[0] = '\0';
 
-	snprintf(data, 256, "username=%s&credential=%s&realm=%s&ajax=1"
+	snprintf(data, sizeof(data), "username=%s&credential=%s&realm=%s&ajax=1"
 	         "&redir=%%2Fremote%%2Findex&just_logged_in=1",
 	         username, password, realm);
 
@@ -550,8 +550,9 @@ int auth_log_in(struct tunnel *tunnel)
 		}
 
 		url_encode(tokenresponse, cfg->otp);
-		snprintf(data, 256, "username=%s&realm=%s&reqid=%s&polid=%s&grp=%s"
-		         "&code=%s&code2=&redir=%%2Fremote%%2Findex&just_logged_in=1",
+		snprintf(data, sizeof(data), "username=%s&realm=%s&reqid=%s"
+		         "&polid=%s&grp=%s&code=%s&code2="
+		         "&redir=%%2Fremote%%2Findex&just_logged_in=1",
 		         username, realm, reqid, polid, group, tokenresponse);
 
 		ret = http_request(tunnel, "POST", "/remote/logincheck", data, &res);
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
It would be great to see the actual warnings to understand the rationale behind this change.

I had already installed GCC 7.2.0 but option `-Wstring-truncation` is unrecognized and I don't see any warnings by default. Which GCC version do we need to test with? Perhaps GCC 8? I have found an article about [Detecting String Truncation with GCC 8](https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/).

martinetd:
> Perhaps we should switch to asprintf in addition to using strdup, although asprintf is not available in alternative C libraries - so we still need to check for asprintf and maintain two versions of the code. Does this make sense?

I didn't even know about that note when I switched to `strdup`, I guess that's good :)

I really don't like the idea of maintaining two versions even if there aren't many snprintf calls.
Although we could just redefine it easily, just `snprintf(NULL, 0, ...)` then malloc the proper size... I guess we could just always use our own definition of the function instead of relying on the glibc-specific `asprintf`, they're probably more efficient though, I don't know.

martinetd:
Sure; it kind of feels like sweeping the issue under the rug but with the current static string sizes it should be safe to just extend the data buffer, I'm not fussy.
Will force push an update later today as well.

DimitriPapadopoulos:
I believe we are not really sweeping the issue under the rug since _GCC_ calculates the maximal possible size of the _snprintf()_ output, based on the maximum possible size of each individual string involved. Does this make sense or have I misunderstood the compiler warnings?

DimitriPapadopoulos:
@martinetd I'm not necessarily against dynamic allocation, especially if it avoids **silent** truncation errors. The alternative is static arrays together with systematic truncation detection - but then errors have to be really exceptional. I don't want to dictate a solution, just thinking about alternatives. If you feel dynamic allocation is better in the long term, so be it. But then shouldn't we do the same with _snprintf()_ ?

martinetd:
I'm not sure what to say there. I think this (using bigger static strings) is fine for now, but I don't think we should be afraid of adding errors (e.g. "defensive programming" - making sure things that should work actually work)


Regarding dynamic allocations in general and `snprintf`, I agree in general, I just don't - personally - like relying on `_GNU_SOURCE`-constrained functions.
Now that I'm looking though, we already define `_GNU_SOURCE` for `memmem`, and I see no alternate definition of `memmem` for non-glibc build so _openfortivpn_ will fail to build on alternative libcs already. If that choice has been made consciously, then yes, we probably should make the jump and switch to `asprintf`.

It doesn't have to be done in this PR though.

DimitriPapadopoulos:
Indeed _asprinf()_ is available [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=asprintf&manpath=FreeBSD+2.1.6.1-RELEASE) and apparently on macOS too.

Let's stick with the bigger static strings then. We can handle dynamic allocation in a future PR.

mrbaseman:
Looking over this discussion I would vote for an own implementation of a function along the lines of asprintf as suggested already before

DimitriPapadopoulos:
@mrbaseman Do you agree with increasing sizes for now, adding error checks, and leaving dynamic allocation to a later PR?

Also _asprintf()_ is available on all supported platforms, why write our own implementation? If really needed, implementations are already available out there:
* [Unix Incompatibility Notes: I/O Functions](http://unixpapa.com/incnote/stdio.html)
* [A short asprintf implementation](https://modelingwithdata.org/pdfs/174-asprintf.pdf)

martinetd:
Yeah, after checking, both musl and uClibc define both `memmem` and `asprintf`, so most systems will have that. It looks like whatever libc MacOS uses defines these as well... (that leaves windows native builds, but I don't think we care about that)

So they're not really standard and need `#define _GNU_SOURCE` but I guess they're "standard enough"?...

Anyway, if we already use `memmem` we're already in that boat -- both will need redefining if someone complains about this :)

martinetd:
Works for me.
I just rebased the patch so it's up to date with master (already had bigger buffers done)

PR #345, opened 2018-07-09
Handle backspace/^U for password prompt

martinetd:
The way we're setting terminal options when reading password right now means that we're filling the password buffer with non-printable bytes when someone hits backspace or ^U instead of clearing it appropriately.

I've done two versions of this patch, that I left in the commit history:
 - the first version naively catches such backspace/^H/^U bytes and tries to handle it; this "works" mostly but won't handle utf-8 characters in the input or weird terminals (although the two bytes I'm comparing should work for most)
 - the second version simply does not unset ICANON, I'm not sure why it was unset in the first place (been since the first version of this function) -- not unsetting it means the terminal will handle these as usual (like in a bash 'read' prompt) and will just work™

Please see the individual commits for details.

### DIFF
--- a/src/userinput.c
+++ b/src/userinput.c
@@ -32,7 +32,7 @@ void read_password(const char *prompt, char *pass, size_t len)
 	// Try to hide user input
 	if (tcgetattr(STDIN_FILENO, &oldt) == 0) {
 		newt = oldt;
-		newt.c_lflag &= ~(ICANON | ECHO);
+		newt.c_lflag &= ~ECHO;
 		tcsetattr(STDIN_FILENO, TCSANOW, &newt);
 		masked = 1;
 	}


DimitriPapadopoulos:
I don't know why `ICANON` had been unset either. For example [A Brief Introduction to termios: termios(3) and stty](https://blog.nelhage.com/2009/12/a-brief-introduction-to-termios-termios3-and-stty/) states:

### local modes

* `ICANON` - Perhaps the most important bit in `c_lflag` is the `ICANON` bit. Enabling it enables “canonical” mode – also known as “line editing” mode. When `ICANON` is set, the terminal buffers a line at a time, and enables line editing. Without `ICANON`, input is made available to programs immediately (this is also known as “cbreak” mode).
* `ECHO` in `c_lflag` controls whether input is immediately re-echoed as output. It is independent of `ICANON`, although they are often turned on and off together. **When `passwd` prompts for your password, your terminal is in canonical mode, but `ECHO` is disabled.**

So I'm all for this fix.

DimitriPapadopoulos:
According to [17.3 Two Styles of Input: Canonical or Not](https://www.gnu.org/software/libc/manual/html_node/Canonical-or-Not.html):
> POSIX systems support two basic modes of input: canonical and noncanonical.
> 
> In _canonical input processing_ mode, terminal input is processed in lines terminated by newline (`'\n'`), EOF, or EOL characters. No input can be read until an entire line has been typed by the user, and the `read` function (see [I/O Primitives](https://www.gnu.org/software/libc/manual/html_node/I_002fO-Primitives.html#I_002fO-Primitives)) returns at most a single line of input, no matter how many bytes are requested.
> 
> In canonical input mode, the operating system provides input editing facilities: some characters are interpreted specially to perform editing operations within the current line of text, such as ERASE and KILL. See [Editing Characters](https://www.gnu.org/software/libc/manual/html_node/Editing-Characters.html#Editing-Characters).

> In _noncanonical input processing mode_, characters are not grouped into lines, and ERASE and KILL processing is not performed. The granularity with which bytes are read in noncanonical input mode is controlled by the MIN and TIME settings. See [Noncanonical Input](https://www.gnu.org/software/libc/manual/html_node/Noncanonical-Input.html#Noncanonical-Input).
> 
> Most programs use canonical input mode, because this gives the user a way to edit input line by line. The usual reason to use noncanonical mode is when the program accepts single-character commands or provides its own editing facilities.

So I understand we need to chose between _canonical mode_ (the kernel does the editing using a limited set of cryptic editing characters known only to Unix command-line aficionados) and _noncanonical mode_ (editing is performed in user space preferably by re-using existing libraries such as _GNU Readline_).

So I agree, let's use _canonical mode_ even though its editing facilities are probably useless to most casual users, to avoid control characters. We could of course use _noncanonical mode_ to improve editing facilities, but then we should use a library such as _GUN Readline_ to take into account control characters.

DimitriPapadopoulos:
I suggest we squash these two commits. Instead provide an explanation of the choice between canonical and noncanonical input processing mode.

martinetd:
> I suggest we squash these two commits. Instead provide an explanation of the choice between canonical and noncanonical input processing mode.

Sure, the two commits were mostly here to discuss if we have any reason to keep `ICANON` toggled out.

I'll let this sink for a day to give others a chance to look and do that tomorrow, unless there's any hurry :)

DimitriPapadopoulos:
@adrienverge Any special reason why you had set _noncanonical mode_?
```
newt.c_lflag &= ~(ICANON | ECHO);
```

DimitriPapadopoulos:
A last comment about the commit message. I believe it's not the shell but the kernel handling the control sequences for us. I'd add something like this:

In canonical mode the kernel maintains a line buffer and handles control characters directly.
In noncanonical mode control characters are passed directly to the software running in user space, to be handled by the software itself, possibly with help from a 3rd party library such as _GNU Readline_.
Canonical mode is good enough for our purposes and avoids the burden of handling control characters in _openfortivpn_.

martinetd:
Right, the commit message was still a bit confusing as I didn't reword part of the second commit message that was written as a serie of two; I've added some more explanation.

I was wrong to say shell there but it definitely isn't the kernel either (unless you're writing in a tty), your terminal emulator is handling the ioctl and providing this feature

PR #346, opened 2018-07-09
config: prioritize cli options to config file

### LINKED ISSUE #258: Command-line switches should have priority over config file
Hello,

I finally bothered testing half-internet-routes, but my config file has set-routes=0 so I figured I could just pass --set-routes=1 to restore the default value except that.. It does not work like that, and config file won.

The way we currently fill our cfg struct is:
 * set default values
 * parse options
 * parse config file

This order cannot be changed completely, because options can change where the config file is, so I see two solutions:
 * Split options parsing in two, and loop over arguments twice - the first time only looks for -c/--config while the second time properly sets all the rest.

or
 * Instead of setting defaults at start, set all values to something invalid (e.g. -1), then when parsing config only set fields if they are currently invalid (means they weren't touched by command line options), then finally restore defaults.
We could do that without changing anything for most string arguments which are set to NULL and handled in code, but we really need to separate setting default values for int arguments e.g. `set-routes` to differentiate user asking for 1 and the default value.

The second one is more invasive in the code (adds many `if (field)` checks), but feels somewhat cleaner, I don't have much preference.
Happy to submit a PR if there is agreement on either way to do this (or a 3rd one comes out)

martinetd:
Parse config file and command line arguments in two different variables
and merge them appropriately.

Future work could make the merge function behave differently depending
on environmental circumstances (e.g. openfortivpn run as a setuid
program probably would want to refuse some cli options)

Fixes #258.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -24,6 +24,35 @@
 #include <ctype.h>
 #include <sys/stat.h>
 
+
+const struct vpn_config invalid_cfg = {
+	.gateway_host = {'\0'},
+	.gateway_port = 0,
+	.username = {'\0'},
+	.password = {'\0'},
+	.otp = {'\0'},
+	.cookie = {'\0'},
+	.realm = {'\0'},
+	.set_routes = -1,
+	.set_dns = -1,
+	.pppd_use_peerdns = -1,
+	.use_syslog = -1,
+	.half_internet_routes = -1,
+	.persistent = -1,
+	.pppd_log = NULL,
+	.pppd_plugin = NULL,
+	.pppd_ipparam = NULL,
+	.pppd_ifname = NULL,
+	.pppd_call = NULL,
+	.ca_file = NULL,
+	.user_cert = NULL,
+	.user_key = NULL,
+	.verify_cert = -1,
+	.insecure_ssl = -1,
+	.cipher_list = NULL,
+	.cert_whitelist = NULL
+};
+
 /*
  * Adds a sha256 digest to the list of trusted certificates.
  */
@@ -217,14 +246,19 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			}
 			cfg->pppd_use_peerdns = pppd_use_peerdns;
 		} else if (strcmp(key, "pppd-log") == 0) {
+			free(cfg->pppd_log);
 			cfg->pppd_log = strdup(val);
 		} else if (strcmp(key, "pppd-plugin") == 0) {
+			free(cfg->pppd_plugin);
 			cfg->pppd_plugin = strdup(val);
 		} else if (strcmp(key, "pppd-ipparam") == 0) {
+			free(cfg->pppd_ipparam);
 			cfg->pppd_ipparam = strdup(val);
 		} else if (strcmp(key, "pppd-ifname") == 0) {
+			free(cfg->pppd_ifname);
 			cfg->pppd_ifname = strdup(val);
 		} else if (strcmp(key, "pppd-call") == 0) {
+			free(cfg->pppd_call);
 			cfg->pppd_call = strdup(val);
 		} else if (strcmp(key, "use-syslog") == 0) {
 			int use_syslog = strtob(val);
@@ -244,10 +278,13 @@ int load_config(struct vpn_config *cfg, const char *filename)
 				log_warn("Could not add certificate digest to whitelist.\n");
 
 		} else if (strcmp(key, "ca-file") == 0) {
+			free(cfg->ca_file);
 			cfg->ca_file = strdup(val);
 		} else if (strcmp(key, "user-cert") == 0) {
+			free(cfg->user_cert);
 			cfg->user_cert = strdup(val);
 		} else if (strcmp(key, "user-key") == 0) {
+			free(cfg->user_key);
 			cfg->user_key = strdup(val);
 		} else if (strcmp(key, "insecure-ssl") == 0) {
 			int insecure_ssl = strtob(val);
@@ -258,6 +295,7 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			}
 			cfg->insecure_ssl = insecure_ssl;
 		} else if (strcmp(key, "cipher-list") == 0) {
+			free(cfg->cipher_list);
 			cfg->cipher_list = strdup(val);
 		} else {
 			log_warn("Bad key in config file: \"%s\".\n", key);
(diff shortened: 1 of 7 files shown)


mrbaseman:
a first comment (without having tested and reviewed that in detail yet):
The code change looks quite elegant at first glance, thanks! 
(I hope to find some time to have a closer look soon)

src/config.c
```diff
@@ -274,3 +310,95 @@ int load_config(struct vpn_config *cfg, const char *filename)
...
+	if (src->gateway_host[0])
+		strcpy(dst->gateway_host, src->gateway_host);
+	if (src->gateway_port != invalid_cfg.gateway_port)
+		dst->gateway_port = src->gateway_port;
+	if (src->username[0])
+		strcpy(dst->username, src->username);
+	if (src->password[0])
+		strcpy(dst->password, src->password);
+	if (src->otp[0])
+		strcpy(dst->otp, src->otp);
```
mrbaseman: cookie is in the structure here in between, but doesn't need to be merged (it isn't obtained from cli nor from config file). Maybe we should insert a comment about this?

src/config.c
```diff
@@ -24,6 +24,33 @@
...
+	.realm = {'\0'},
+	.set_routes = -1,
+	.set_dns = -1,
+	.pppd_use_peerdns = -1,
+	.use_syslog = -1,
+	.half_internet_routes = -1,
+	.pppd_log = NULL,
+	.pppd_plugin = NULL,
+	.pppd_ipparam = NULL,
+	.pppd_call = NULL,
```
mrbaseman: isn't `.pppd_ifname = NULL,` missing here?

src/config.c
```diff
@@ -274,3 +310,95 @@ int load_config(struct vpn_config *cfg, const char *filename)
...
 	return ret;
 }
+
+void destroy_vpn_config(struct vpn_config *cfg)
+{
+	free(cfg->pppd_log);
+	free(cfg->pppd_plugin);
+	free(cfg->pppd_ipparam);
+	free(cfg->pppd_ifname);
+	free(cfg->pppd_call);
```
mrbaseman: good catch, `pppd_call` was missing in the [original code](https://github.com/adrienverge/openfortivpn/blob/master/src/main.c#L129)

mrbaseman:
apart from the changes in this PR I have noticed that we use statically allocated strings for some fields in [vpn_config](https://github.com/adrienverge/openfortivpn/blob/master/src/config.h#L56) and dynamically allocated ones for other fields. 

Sure, the essential ones like `gateway_host` must be specified somehow, either on cli or in the config file, whereas `pppd_log` for instance is optional (I think that's the main difference), but anyhow, the code would look more consistent if dynamic strings were used for all config parameters. 

I just would like to discuss if we should we switch to dynamically allocated strings in general. What do you think? However, if we decide to do so, this would be a separate pull request

src/config.c
martinetd: Ah, it wasn't in the default struct when I copied it, and I only fixed that over there; good catch.
I was actually thinking of removing all the 0/NULL setting here though as that is the default values; being explicit has some advantages but it's also more maintenance work

src/config.c
martinetd: I'd actually rather move cookie into the `tunnel` struct, I can't think of any good reason for it to be in the config.
Any reason not to?

src/config.c
```diff
@@ -274,3 +310,95 @@ int load_config(struct vpn_config *cfg, const char *filename)
...
+	if (src->set_routes != invalid_cfg.set_routes)
+		dst->set_routes = src->set_routes;
+	if (src->set_dns != invalid_cfg.set_dns)
+		dst->set_dns = src->set_dns;
+	if (src->pppd_use_peerdns != invalid_cfg.pppd_use_peerdns)
+		dst->pppd_use_peerdns = src->pppd_use_peerdns;
+	if (src->use_syslog != invalid_cfg.use_syslog)
+		dst->use_syslog = src->use_syslog;
+	if (src->half_internet_routes != invalid_cfg.half_internet_routes)
+		dst->half_internet_routes = src->half_internet_routes;
```
martinetd: I noticed I'm missing `persistent` here for some reason (I probably started this on an old version), will fix

src/config.c
```diff
@@ -24,6 +24,33 @@
 #include <ctype.h>
 #include <sys/stat.h>
 
+
+const struct vpn_config invalid_cfg = {
+	.gateway_host = {'\0'},
+	.gateway_port = 0,
+	.username = {'\0'},
+	.password = {'\0'},
+	.otp = {'\0'},
```
martinetd: Writing it there to not forget, I noticed `--otp` isn't in the man page, but it is in `--help`.

src/main.c
```diff
@@ -369,21 +354,22 @@ int main(int argc, char **argv)
...
 		else
 			log_warn("Could not load config file \"%s\" (%s).\n",
 			         config_file, err_cfg_str(ret));
 	}
+	// Then apply CLI config
+	merge_config(&cfg, &cli_cfg);
+	set_syslog(cfg.use_syslog);
 
 	// Read host and port from the command line
 	if (optind == argc - 1) {
```
martinetd: !$#@ github for not being able to comment on arbitrary lines.. 20 or so lines below this I noticed some option parsings are redundant with the merge (e.g. reset username if it was given on command line etc) -- I'll update these to write in cli_cfg directly and let the merge function do the job

src/config.c
martinetd: (Note that persistent is set as an unsigned int in the config, read as a long int (boundaries only checked in cli option not in config file), but used as a bool?! Should this be changed to int like our other "bools" instead?)

src/config.c
```diff
@@ -24,6 +24,33 @@
...
+	.use_syslog = -1,
+	.half_internet_routes = -1,
+	.pppd_log = NULL,
+	.pppd_plugin = NULL,
+	.pppd_ipparam = NULL,
+	.pppd_call = NULL,
+	.ca_file = NULL,
+	.user_cert = NULL,
+	.user_key = NULL,
+	.verify_cert = -1,
```
martinetd: Now I'm checking stuff, it looks like `verify_cert` is used but never set -- nor command line nor config, so default value of 1 is always used.

src/config.c
mrbaseman: `persistent` is used as delay in seconds between connection attempts. persistent=300 would mean that _openfortivpn_ tries to re-establish a connection every 5 minutes after dropping out, and if I recall right it was on cli only at the beginning. Negative int values would overflow to very large uint numbers, which effectively would be a very long delay between connection attempts, in other words _openfortivpn_ would basically not reconnect within a reasonable time (which is the default anyway).

martinetd: Oh, I had missed it being used as `sleep` argument, right. I'll just add the overflow check to config.c

martinetd:
Done most of my remarks, except moving `gateway_ip` out, removing redondant initializations and reordering fields of the `config` struct. Some were merged in initial commit (had to rebase anyway), unrelated changes were kept separate

martinetd:
It's been a while so I don't really remember, but each commit should compile and work individually (just checked they compile at least) -- I'm not fussy either way though if you prefer squashing.

Issue #229, opened 2017-11-07
Feature request, fork in background


skuti-is:
Add command line and config file option to fork in background and specify log file.

dweggemans:
Any suggestions on how to do accomplish this with the current version?

DimitriPapadopoulos:
You may start _openfortivpn_ as a _systemd_ service:
* `StandardOutput=` should redirect the output of _openfortivpn_ to a file.
* With `Type=simple`, I think the PID of _openfortivpn_ is known to _systemd_.
* `Restart=` can be used instead of `--persistent`.
* Is forking really required?

As was the case with `--persistent`, I would rather see daemon-like options implemented in _systemd_. Perhaps we could create a wiki to explain how to start _openfortivpn_ as a service. Alternatively, if really needed inside _openfortivpn_, the requested functionality could perhaps be implemented in a _daemonize()_ function that could include the --persistent` functionality.

@skuti-is @dweggemans Can you expand on why it would be better to implement daemon-like functionality in _openfortivpn_ rather than use _systemd_? I am not against the suggestion by the way, just trying to get the big picture.

jandersonbt:
I second the _systemd_ options.  Using existing technologies to accomplish tasks is always best.  I would also like to see the ability to use `Type=notify` in the service file.  Currently, I am working on a project where other services need to start after openfortivpn has the tunnel up and running.  However, when using `Type=simple` in my _systemd_ service file the service shows as complete before the tunnel is up and running. I can see this behavior when using `tail -f` to view the syslog file while restarting the openfortivpn service.  I did attempt to use the `Type=notify` option, however openfortivpn doesn't notify _systemd_ when the tunnel is up and running, and this leaves the systemctl command hanging.  I've been reviewing systemd documentation, but I have yet to find a way around the notify option to accomplish what I need.  Currently, I'm using a `sleep` command to pause while the vpn connects.  This works as long as the vpn connects quickly.

DimitriPapadopoulos:
@jandersonbt Any clue/link how to implement the _systemd_ notification? Additional headers needed and libraries to link with?

DimitriPapadopoulos:
Any clue about _systemd_-like functionality on FreeBSD and macOS?

DimitriPapadopoulos:
[...]

PR #370, opened 2018-09-11
Notify tunnel up to systemd with sd_notify

fsateler:
Allows usage of Type=notify systemd service types.

Fixes: #229


DimitriPapadopoulos:
Perhaps this functionality would be best left to a _pppd_ startup script (in `/etc/ppp/ip-up`). However this is a small patch which helps delegate functionality to _systemd and therefore possibly avoid future options and duplication of functionality in _openfortivpn_. So I'm happy to add it.

But keep in mind notifying the _pppd_ tunnel is up is best left to _pppd_. Unfortunately _pppd_ doesn't provide much more than scripts in `/etc/ppp/ip-up` or `/etc/ppp/ip-down` (and `/etc/ppp/ipv6-up` and `/etc/ppp/ipv6-down` the day IPv6 is supported). Perhaps the proper way to handle this would be to make _pppd_ more friendly to users and accept more options from users.

PR #376, opened 2018-10-11
Add --otp-delay option

### LINKED ISSUE #348: Include a delay when sending the OTP
Hi all. 
The server I currently have to connect to seems to reject OTP with a simple EOF if the two authentication requests are too fast.

I solved it by adding a `sleep(2);` before the call to `try_otp_auth` in `http.c`.

Can you please include this in future release?

ShogunPanda:
This fixes #348.

I added a --otp-delay option, which takes a integer representing the number of seconds to wait before sending the OTP to the server.

### DIFF
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -9,6 +9,7 @@ openfortivpn \- Client for PPP+SSL VPN tunnel services
 [\fB\-u\fR \fI<user>\fR]
 [\fB\-p\fR \fI<pass>\fR]
 [\fB\-\-otp=\fI<otp>\fR]
+[\fB\-\-otp-delay=\fI<delay>\fR]
 [\fB\-\-realm=\fI<realm>\fR]
 [\fB\-\-set-routes=<bool>\fR]
 [\fB\-\-no-routes\fR]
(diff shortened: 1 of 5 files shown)


ShogunPanda:
Up - Any plan to merge this?

mrbaseman:
We have decided to focus on finding the root cause of the issue, rather than adding another command line option. The improved debugging output of the recently released version 1.8.0 might help here. Does the http response code differ between the two cases "wrong password used" and "http requests sent too quickly after each other"?

ShogunPanda:
Hi, as specified in #348, the problem is not happening anymore, so I think this is not needed anymore. Closing it.

ShogunPanda:
As mentioned on #348, the issue is still happening for me so I think this PR is still needeed. Reopening.

ShogunPanda:
PR updated to match upstream master.

DimitriPapadopoulos:
@ShogunPanda Have you been able to upgrade to 1.8.1 and investigate the root cause of this issue?

ShogunPanda:
@DimitriPapadopoulos Yup, I did. If you look in #348 I posted the output but there is nothing interesting. Apparently that server doesn't like the OTP to being sent too quickly.

mrbaseman:
I have merged the change via #433. It will be in the next release.

PR #380, opened 2018-10-26
fix regression when no systemd is present on linux

mrbaseman:
on Scientific Linux release 6.10 (Carbon), a RHEL clone maintained at CERN
I have noticed that configure fails with the following error:

checking for LIBSYSTEMD... ./configure: line 4877: syntax error near unexpected token `else'
./configure: line 4877: `else'

if libsystemd is not found. The empty square bracket in configure.ac:
  PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [ ])
suppresses a human readable configure error but produces invalid shell code.
Printing out a message that we haven't found libsystemd solves the problem.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,7 +21,7 @@ AM_SILENT_RULES([yes])
 PKG_CHECK_MODULES(OPENSSL, [libcrypto >= 0.9.8 libssl >= 0.9.8], [], [AC_MSG_ERROR([Cannot find OpenSSL 0.9.8 or higher.])])
 AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthread.])])
 AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])
-PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [ ])
+PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [AC_MSG_RESULT([libsystemd not present])])
 
 # Checks for header files.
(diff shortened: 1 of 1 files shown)


adrienverge:
I see `[]` (without spaces) at other places.

Does the following fix the problem?
```diff
-PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [ ])
+PKG_CHECK_MODULES(LIBSYSTEMD, [libsystemd], [AC_DEFINE(HAVE_SYSTEMD)], [])
```

If not, OK for your fix! :+1:

DimitriPapadopoulos:
Ah indeed, for the sake of consistency perhaps it is better to use `[ ]`.

mrbaseman:
No, I have tried that first. An empty `[]` without space (or an omitted square brace pair) makes this a hard requirement and prints out an error message:
```checking for LIBSYSTEMD... configure: error: Package requirements (libsystemd) were not met:

No package 'libsystemd' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables LIBSYSTEMD_CFLAGS
and LIBSYSTEMD_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
```

DimitriPapadopoulos:
I'm not suggesting an empty `[]` but one with a space `[ ]` like Adrien.

DimitriPapadopoulos:
My wrong, sorry;

Actually most other tests do print a message:
`AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthread.])])`
`AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])`
The first `[]` is indeed empty but is followed by a second one that is not empty. I guess _autoconf_ needs at least one occurrence of an non-empty message.

mrbaseman:
I think the `PKG_CHECK_MODULES` macro is somewhat special, citing from [here](https://autotools.io/pkgconfig/pkg_check_modules.html): "In contrast with almost all of the original macros, though, the default action-if-not-found will end the execution with an error for not having found the dependency."
So, in the other places the `[]` is an empty argument which causes the default action to be taken. Here, it may not be empty, and `[ ]` (a space enclosed in square brackets) indeed is not an empty string, but inserted into the shell code of the configure script it's "almost nothing" between a `then`and an `else`. 
Probably we could have used  `[:]` (at least for bash, but I'm not sure how common the noop command is in other shells). Anyhow, a message about what's happening behind the scenes is nice, so nobody can complain that the systemd dependency has been ignored silently ;-)

Issue #378, opened 2018-10-16
Unable to set correct route for OSX high sierra 10.13.6


demonlj:
mac osx: high sierra 10.13.6
openfortivpn: 1.7.1 via homebrew installation
It seems like delete my default static route, cause the network shutdown, then app terminated
config

```host = vpn.suda.edu.cn
port = 443
username = user
password = pass
set-routes = 1
set-dns = 0
trusted-cert = c57ed276aab8b6fe7b8bbbe1e9c5f677eaf6916485bd7db9dfd66fa276137d0d
```
debug
```INFO:   Cancelling threads...
INFO:   Setting ppp interface down.
INFO:   Restoring routes...
DEBUG:  route -n delete 221.224.34.184 -netmask 255.255.255.255
delete net 221.224.34.184
DEBUG:  route -n delete 0.0.0.0 -netmask 0.0.0.0
delete net 0.0.0.0
DEBUG:  route -n add -net 0.0.0.0 -netmask 0.0.0.0 -interface ppp0
add net 0.0.0.0: gateway ppp0
DEBUG:  Waiting for pppd to exit...
DEBUG:  waitpid: pppd exit status code 16
INFO:   pppd: The link was terminated by the modem hanging up.
INFO:   Terminated pppd.
INFO:   Closed connection to gateway.
DEBUG:  server_addr: 221.224.34.184
DEBUG:  server_port: 443
DEBUG:  gateway_addr: 221.224.34.184
DEBUG:  gateway_port: 443
ERROR:  connect: Network is unreachable
INFO:   Could not log out.
```

Correct route via original forticlient
```$ netstat -nrt
Routing tables

Internet:
Destination        Gateway            Flags        Refs      Use   Netif Expire
default            link#15            UCS            68        0    ppp0
default            192.168.1.1        UGScI           8        0     en1
1.1.1.1            42.244.60.70       UH              0        0    ppp0
10.9.0.110         link#15            UHWIi           3       18    ppp0
17.248.161.8       link#15            UHWIi           1       11    ppp0
17.248.161.46      link#15            UHWIi           2       29    ppp0
17.252.156.35      link#15            UHWIi           1       28    ppp0
23.198.140.135     link#15            UHW3I           0       34    ppp0      7
74.125.203.188     link#15            UHWIi           2       13    ppp0
117.18.237.29      link#15            UHW3I           0        6    ppp0      6
127                127.0.0.1          UCS             0        0     lo0
[...]

PR #385, opened 2018-11-06
improve routing code for Mac OSX and FreeBSD

mrbaseman:
this solves #378:

On OSX and FreeBSD there is no /proc file system where we could look up the routes currently configured.  We have to parse the output of the netstat tool. In this output there are also routes which are directly connected to specific interface, e.g. the tunnel interface automatically has a default route attached. When determining the route to the vpn gateway for instance, we need a route which does not go through the tunnel. Therefore we need a way to specify that routes over a particular interface are taken (or explicitly not taken, in this particular case).

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -622,6 +622,12 @@ static int ipv4_del_route(struct rtentry *route)
 	strcat(cmd, " -netmask ");
 	strncat(cmd, inet_ntoa(route_mask(route)), 15);
 
+	if (!(route->rt_flags & RTF_GATEWAY)) {
+		strcat(cmd, " -interface ");
+		strncat(cmd, route_iface(route),
+		        SHOW_ROUTE_BUFFER_SIZE - strlen(cmd) - 1);
+	}
+
 	log_debug("%s\n", cmd);
 
 	int res = system(cmd);
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -667,15 +667,6 @@ int ipv4_protect_tunnel_route(struct tunnel *tunnel)
 			goto err_destroy;
 	}
 
-	ret = ipv4_get_route(def_rt);
-	if (ret != 0) {
-		log_warn("Could not get current default route (%s).\n",
-		         err_ipv4_str(ret));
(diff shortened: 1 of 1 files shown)


mrbaseman:
the string handling is a bit ugly, but in #378 we have the confirmation that it works. I'd squash the commits during merge and clean up the commit messages in this turn

src/ipv4.c
```diff
@@ -487,7 +487,13 @@ static int ipv4_get_route(struct rtentry *route)
 
 		if (((dest & mask) == (rtdest & rtmask & mask))
 		    && (mask >= route_mask(route).s_addr)
-		    && (mask <= rtmask)) {
+		    && (mask <= rtmask)
+		    && ((route_iface(route) == NULL)
+		        || (strcmp(iface, route_iface(route)) == 0)
+		        || ((strlen(route_iface(route))>0)
```
DimitriPapadopoulos: Add spaces around binary operator `>` and perhaps remove some parenthesis, changing:
```
		        || ((strlen(route_iface(route))>0)
		            && (route_iface(route)[0] == '!')
		            && (strcmp(iface, &route_iface(route)[1]) != 0))
```
to:
```
		        || (strlen(route_iface(route)) > 0
		            && route_iface(route)[0] == '!'
		            && strcmp(iface, &route_iface(route)[1]) != 0)
```

DimitriPapadopoulos:
Just a side note:
In the future perhaps we should call the same C API as _netstat_ instead of running _netstat_. See [_netstat_ source code](https://github.com/freebsd/freebsd/tree/master/usr.bin/netstat) and the [route(4) man page](https://www.freebsd.org/cgi/man.cgi?query=route&sektion=4).

mrbaseman:
About using the C API: Actually I have thought about this quite some time ago, but to be honest, I didn't understand how it works internally. Maybe I just have to study the source of the tools with my knowledge of today, but at that time I felt it's so much easier to parse the output than diving into the internal mechanisms. But well, you are right, we can keep this in mind for future improvement.

src/ipv4.c
mrbaseman: I have implemented this as suggested before merging

PR #387, opened 2018-11-07
Bail out if not run as root

### LINKED ISSUE #373: Bail out if not run as root?
See latest comments in #362.

DimitriPapadopoulos:
Actual root privileges are required in many places, depending on the OS:
* On Linux, the 'noauth' option of pppd seems to require actual root privileges. Being a meber of the 'dip' group on Ubuntu is not sufficient.
* On Linux, setting routes requires root privileges unless user has CAP_NET_ADMIN capability.
* On FreeBSD, setting routes using 'route' requires root privileges.

Fixes #373.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -458,8 +458,11 @@ int main(int argc, char **argv)
 	if (cfg.otp[0] != '\0')
 		log_debug("One-time password = \"%s\"\n", cfg.otp);
 
-	if (geteuid() != 0)
-		log_warn("This process was not spawned with root privileges, this will probably not work.\n");
+	if (geteuid() != 0) {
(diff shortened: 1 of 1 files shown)


mrbaseman:
Hmm... I would merge this, but travis-ci is stuck somehow, although the commit itself seems to have passed the required check

mrbaseman:
ok, after updating the branch travis-ci ran through smoothly

DimitriPapadopoulos:
Not the first time travis is stuck. This seems to be happening often these days.

mrbaseman:
I have seen it on my pull requests occasionally and there it helped to refresh the travis-ci web interface a couple of times. But I think this can do only the owner of each pull request.

PR #389, opened 2018-11-08
Helpful error messages

DimitriPapadopoulos:
Printing "Bad HTTP response" for any HTTP response different from 200 is imprecise and eventually not helpful.

Better error messages might be a compelling reason to use a lightweight 3rd party HTTP library.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -552,7 +552,8 @@ int auth_log_in(struct tunnel *tunnel)
 
 	if (strncmp(res, "HTTP/1.1 200 OK\r\n", 17)) {
 		char word[17];
-		if (sscanf(res, "%s %d", word, &ret) < 2)
+
+		if (sscanf(res, "%16s %d", word, &ret) < 2)
 			ret = ERR_HTTP_BAD_RES_CODE;
 		goto end;
 	}
@@ -606,7 +607,8 @@ int auth_log_in(struct tunnel *tunnel)
 
 		if (strncmp(res, "HTTP/1.1 200 OK\r\n", 17)) {
 			char word[17];
-			if (sscanf(res, "%s %d", word, &ret) < 2)
+
+			if (sscanf(res, "%16s %d", word, &ret) < 2)
 				ret = ERR_HTTP_BAD_RES_CODE;
 			goto end;
 		}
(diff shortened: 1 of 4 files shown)


mrbaseman:
Sure, this answer that http status was different from 200 OK is often not very helpful, but [this strcmp](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c#L553)  is not the only place that returns an error code in [auth_log_in](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c#L518-L616). 

There is also [try_otp_auth](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c#L548),[get_auth_cookie](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c#L610), [http_request](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c#L599) and [here](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c#L541), too, [get_value_from_response](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c#L573). `ERR_HTTP_BAD_RES_CODE` is just one out of [7 possible return values](https://github.com/adrienverge/openfortivpn/blob/master/src/http.h#L23-L29). These return codes are all negative.

What about returning the http response code (a positive value) for a more detailed diagnostics instead of  `ERR_HTTP_BAD_RES_CODE`? Or we could expand the list of possible error codes and return a specific one for each error case that we have (`ERR_HTTP_BAD_RES_CODE_IN_PARSE_CONFIG`, `ERR_HTTP_BAD_RES_CODE_DURING_LOGIN`, `ERR_HTTP_BAD_RES_CODE_AFTER_LOGIN_CHECK`, ... and maybe more), and of course more meaningful text messages assigned in [err_http_str](https://github.com/adrienverge/openfortivpn/blob/master/src/http.h#L31-L48), too.

mrbaseman:
I have proposed a mechanism to show the HTTP status code in the debug output (and possible other reasons from `err_http_str()`, too). @DimitriPapadopoulos could you have a look, please? I hope that coding style is ok this time :-)

If we want we can add more different error values and/or more meaningful messages in the debug output.

DimitriPapadopoulos:
Returning a positive HTTP status code (except 200 for success which should translate to 0?) *or* a negative code for other non-HTTP errors sounds good.

DimitriPapadopoulos:
@mrbaseman Looks good to me with you changes. Thanks!

PR #392, opened 2018-11-13
use ppp interface name instead of ip for routing

mrbaseman:
on FreeBSD (and OSX), namely when adding routes via route command

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -81,6 +81,8 @@ int http_send(struct tunnel *tunnel, const char *request, ...)
 	else if (length >= BUFSZ)
 		return ERR_HTTP_TOO_LONG;
 
+	log_debug_details("http_send : \n%s\n", buffer);
+
 	while (n == 0)
 		n = safe_ssl_write(tunnel->ssl_handle, (uint8_t *) buffer,
 		                   length);
@@ -147,6 +149,7 @@ int http_receive(
 		                  (uint8_t *) buffer + bytes_read,
 		                  BUFSZ - 1 - bytes_read);
 		if (n > 0) {
+			log_debug_details("http_receive : \n%s\n", buffer);
 			const char *eoh;
 
 			bytes_read += n;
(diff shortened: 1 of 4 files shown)


mrbaseman:
With 5875f03 I have tested this successfully on Mac OSX. 
The problem was that on OSX a second _default_ route is added, which is needed for the ppp link. `ipv4_get_route()` has found that one and removed it by mistake. 
For completeness we should also check the netmask. I'll push another commit.

mrbaseman:
I have successfully tested on Ubuntu, OSX and FreeBSD, now.

mrbaseman:
I have squashed the commits on this branch

mrbaseman:
I have added another detail here: On Mac OSX and FreeBSD the debug output was really huge because the whole routing table and all parsing results for each line were included. Now it's only printed when the `-v` option is used twice.

@DimitriPapadopoulos @adrienverge Ok to merge this one? 

I would not squash these commits, because it's two separate things, but both about OSX and BSD, therefore I have chosen to push c79879e into the same branch.

adrienverge:
About the change, I trust you, so I let you and Dimitri decide what's best.

About the commits, you're right to avoid squashing them if they are two different changes. However if you want you can drop the third commit (merge commit) by clicking "Rebase and merge" instead of "Merge", or in command-line by rebasing on `origin/master` and `push -f`.

mrbaseman:
I have rebased this on the current master. The third commit is new now. I have noticed that I abuse the "packet logging level" in 725cb18 for the netstat parsing output, so a more general name for this constant seems appropriate.
@DimitriPapadopoulos do you have any remarks on this? Otherwise I would like to apply this one and tag the new release in #393

mrbaseman:
I have successfully re-tested on Ubuntu 12.04, 16.04, 18.04, MacOS X 10.11, and FreeBSD 11.2.

Issue #396, opened 2018-11-20
cant connect with 1.8.0 (1.7.1 works)


cheese1:
with 1.8.0 i cannot connect:

INFO:   Connected to gateway.
Two-factor authentication token: 
INFO:   Authenticated.
INFO:   Remote gateway has allocated a VPN.
ERROR:  : No such file or directory.
INFO:   Closed connection to gateway.
INFO:   Logged out.

with 1.7.1 it works as usual.
would e great, if the error-message would show the file/directory requested.

cheese1:
since https://github.com/adrienverge/openfortivpn/pull/300 pppd is checked against at build-time. as i did not install ppp in my buildsystem, that part broke.

DimitriPapadopoulos:
The file/directory requested really is an empty string(should be `/usr/sbin/pppd` or similar instead). That's a build time issue, but again _pppd_ should **not** be a requirement at build time, only run time. My suggestions:
* We should detect that the path to _pppd_ is an empty string at **build time**.
* We should choose a default value if pppd is not to be found on the build machine, such as `/usr/sbin/pppd` which should actually fit most Linux distributions.

DimitriPapadopoulos:
I'm closing #396 and re_opening #397 instead.

mrbaseman:
The assumption was that `pppd` is installed on the build machine as well, and if not, one should specify the path using the configure option. Since we already have a fallback to use `pppd` when nothing is specified and nothing is found, it makes sense to set up the PPP_PATH accordingly. That's what I have implemented in #399

DimitriPapadopoulos:
Does it sound possible to use different fallbacks, `/usr/sbin/pppd` on Linux and Mac OS and `/usr/sbin/ppp` on FreeBSD? I think this can be achieved with `AM_COND_IF([FREEBSD], ...`.

mrbaseman:
hmm... that would be a fallback for a very specific case:
* compiling on a different machine than the one that is supposed to run the code
* **both of them** must be FreeBSD
* `ppp` client **not** installed on the build machine (at run-time it is needed anyway).

This autoconf code is already quite questionable. We currently check for only one place (`/usr/sbin/pppd`) and use it if it is present. 
With the fallback now we use it even when it is not found _and_ `/usr/sbin/ppp` is not found either _and_ the user has not specified a different location via configure argument. 
[...]

PR #397, opened 2018-11-20
add ppp to build requirements on fedora

cheese1:
one needs to install ppp before building on fedora
https://github.com/adrienverge/openfortivpn/issues/396


DimitriPapadopoulos:
Please show us the error messages, _pppd_ should **not** be needed to build _openfortivpn_, only to run it.

Which version of _openfortivpn_ is this?

PR #398, opened 2018-11-20
cipher-list option was always ignored

mrbaseman:
the cipher list option has always been considered only inside a branch
which handles the case when it is not set. So, the default setting
did always win. When the option was set in the config or on the command line
there was no code to handle it, so the default from openssl has been
taken in that case

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -788,9 +788,12 @@ int ssl_connect(struct tunnel *tunnel)
 	if (!tunnel->config->insecure_ssl && !tunnel->config->cipher_list) {
 		const char *cipher_list = "HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4";
 
-		if (tunnel->config->cipher_list)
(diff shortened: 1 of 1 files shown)


mrbaseman:
codacy has spotted some dead code and upon a closer look I have spotted this issue. 
Easy test: specify an empty string as a cipher list and it should fail

mrbaseman:
hmm... now Codacy is listed, but it is stuck

DimitriPapadopoulos:
@mrbaseman I have set up Coverity which runs on the `coverity_scan` branch but I haven't set up Codacy which should be triggered by updates on the `master` branch. My only guess is that they might blacklist us if too many commits in a short time (that's why Coverity runs on the `coverity_scan` branch).

mrbaseman:
@DimitriPapadopoulos the checks have returned results in the meantime. There is nothing special apart from the fact that the free account for open source projects has a lower priority than payed accounts. So just waiting for some hours should be sufficient to obtain the results. 
Anyhow, the dead-code issue has been solved by this pull request.

PR #399, opened 2018-11-20
add a fallback for PPP_PATH

### LINKED ISSUE #396: cant connect with 1.8.0 (1.7.1 works)

mrbaseman:
this should fix #396:
if neither ppp nor pppd are found at configure time, the assumption
is that pppd shall be used on the target system, but PPP_PATH was
not properly set in this case.
easy test: run configure --with-pppd=no --with-ppp=no and check
the config.log and the resulting Makefile


mrbaseman:
hmm... automated code review is stuck here as well...

cheese1:
works for my problem. thanks

Issue #401, opened 2018-11-21
url_encode() doesn't work for utf-8


darddan:
in `src/http.c` in function `url_encode` the `isalnum()` works with ASCII characters only. So if a username contains special characters from other languages that aren't in ASCII then it can't authenticate them.

DimitriPapadopoulos:
Thinking about it again, are you certain the point of failure is here? `isalnum()` is used precisely to distinguish ASCII characters (which don't need further encoding) and non-ASCII characters which **are** handled as far as I can see [later on](https://github.com/adrienverge/openfortivpn/blob/01898c3/src/http.c#L49-L55).

Have you experienced actual failures? If so are you certain the problem is in this part of the code or even in _openfortivpn_?

darddan:
Yes I have experienced this problem today. I was trying to use [ł](https://en.wikipedia.org/wiki/%C5%81) on the username and it wasn't working. 
I did [this](https://github.com/darddan/openfortivpn/commit/f05ffdee4eaebecb7fd13954d14e4c70416beea2) as a workaround and now it works. I tried to implement it in a better way but it appears that checking if an utf-8 character in c is a little more complicated than I thought.

darddan:
Found the issue, fixed it and explained it in the pull request comment.
I tested it by adding: 
```
	printf("%s\n", username);
	exit(0);
```

after 

```
	url_encode(username, tunnel->config->username);
```

in lines `535` and `536`

DimitriPapadopoulos:
@mrbaseman I think we get UTF-8 encoded strings from all sources:
* I expect configuration files to be UTF-8 encoded.
* I believe the arguments passed to `main()` are also UTF-8 encoded under most locales on Linux/macOS/BSD.
* Also I believe _getchar_() and other functions we use for user input return UTF-8 encoded strings as well.

On my own Linux computer, the following test program:
```C
#include <stdio.h>

int main(int argc, char **argv)
{
    for (int i = 1; i < argc; i++) {
        printf("argv[%d]:", i);
        for (unsigned char *c = (unsigned char *)argv[i]; *c; c++)
            printf(" 0x%2x", *c);
        printf("\n");
    }

    printf("Enter a character: ");
    int c = getchar();
    while (c != EOF && c != '\n') {
        printf("0x%2x ", c);
        c = getchar();
    }
    printf("\n");
}
```
shows UTF-8 encoded strings:
```
$ ./test 1 ł
argv[1]: 0x31
[...]

PR #402, opened 2018-11-21
Fix url_encode() for non alphanumericals (utf-8)#401

darddan:
This fixes the issue for me. I suspect this has to do with casting a `signed char` to `int` or something similar. The value of `*str >> 4` was returning a negative integer and then it would get that value from `hex[]` array, which in my case was a `0`.
This `0` was being considered an end of string in `dest` then.


DimitriPapadopoulos:
My wrong, URL encoding uses UTF-8 as well . `ł` translates to `%C5%82` so it does work.

PR #403, opened 2018-11-22
HTML URL Encoding Reference uses upper case

DimitriPapadopoulos:
All examples in this reference use upper case for hexadecimals:
	https://www.w3schools.com/tags/ref_urlencode.asp

While URL decoding is (hopefully!) case insensitive, let's follow the reference blindly.


mrbaseman:
I have tested successfully on Ubuntu, Mac OSX and FreeBSD and successfully connected to our vpn gateway with this.

PR #405, opened 2018-11-25
Changelog

mrbaseman:
* improve markdown coding style
* add changes made in 1.6.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,72 +4,214 @@ Openfortivpn Changelog
 Legend
 ------
 
-*  [+] new feature, improvement
-*  [-] bug fix
-*  [~] change in behavior
+* [+] new feature or improvement
+* [-] bug fix
+* [~] change in behavior
 
 Releases
 --------
 
 ### Upcoming release (current master)
-Note: this section still has to be reviewed. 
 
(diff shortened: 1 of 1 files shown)


mrbaseman:
Hi Adrien,

the background for the format changes here is that Codacy has complained a lot about wrong markdown format in the changelog (add two spaces here, remove one there), but it seems that I have fixed all the 50 issues now . The most severe problem was the usage of the square brackets which can be understood as reference links in some context (obviously Github doesn't treat them as such, at least in the way I have used them). Probably moving the legend down and adding colons there has made the checks happy. The background behind these checks is that the markdown code should be portable between different online platforms.

I'll go through the rest of the commit history, and add it to the high level changelog as well. 
Now that I believe that I have found out the correct format, I think it is more a laborious task.

I think I'll keep the changelog updates in this separate branch and in the end I can squash the commits. The further we go back in history the less important the high level summary, but anyhow, users who are running the first releases that went into distributions probably are interested in such a write-up.

adrienverge:
OK Martin, understood. Once again thanks for all that good work!

Suggestion: if you don't want to deal with Codacy, you can also transform `CHANGELOG.md` to a plain-text ` CHANGELOG` (while keeping the `###`, bullet points and other formattings).

Issue #366, opened 2018-09-05
Could not authenticate to gateway (Bad HTTP response code)


HtwoO:
The other day I was trying to connect to my SSLVPN gateway on [Huawei USG6620](https://e.huawei.com/en/products/enterprise-networking/security/firewall-gateway/usg6600/Components/usg6620). I got an error saying "ERROR:  Could not authenticate to gateway (Bad HTTP response code)." Below are more detailed log and the process I used to setup the VPN.

On client side, I built openfortivpn-1.7.1, to be used on Debian Stretch.

I put the self-signed CA swan.crt in /usr/local/share/ca-certificates/swan/
and then ran update-ca-certificates

Then I ran openfortivpn to connect to the gateway.

`sudo openfortivpn vpn-gateway:443 --username=cat1 --cipher-list=AES256-SHA256 --trusted-cert (censored-fingerprint) -v`

WARN:   Bad port in config file: "0".
DEBUG:  Loaded config file "/etc/openfortivpn/config".
VPN account password: 
DEBUG:  Config host = "censored.gateway.net"
DEBUG:  Config realm = ""
DEBUG:  Config port = "443"
DEBUG:  Config username = "cat1"
DEBUG:  Config password = "********"
DEBUG:  server_addr: censored.gateway.ip
DEBUG:  server_port: 443
DEBUG:  gateway_addr: censored.gateway.ip
DEBUG:  gateway_port: 443
DEBUG:  Gateway certificate validation succeeded.
INFO:   Connected to gateway.
ERROR:  Could not authenticate to gateway (Bad HTTP response code).
INFO:   Closed connection to gateway.
DEBUG:  server_addr: censored.gateway.ip
DEBUG:  server_port: 443
DEBUG:  gateway_addr: censored.gateway.ip
DEBUG:  gateway_port: 443
DEBUG:  Gateway certificate validation succeeded.
INFO:   Logged out.

Gateway setting on Huawei USG6620

 v-gateway public ssl version tlsv11 tlsv12
 v-gateway public ssl ciphersuit custom aes256-sha non-des-cbc3-sha aes128-sha
 v-gateway public certificate-server gw_local.cer enable
 v-gateway ssl interface GigabitEthernet1/0/3 private censored.gateway.net
 v-gateway ssl alias ssl

`#****BEGIN***ssl**1****#` (Just prevent being enterpreted wrongly)

v-gateway ssl
 basic
 ...
[...]

PR #408, opened 2018-12-07
Fix value of Accept-Encoding request header

mss:
While debugging some case of #366 I noticed that the value of the `Accept-Encoding` header had a typo.

Fixing the value doesn't make the issue go away but is one less thing which might go wrong.

I also added the fall-back value `*/*` to the `Accept` request header, just in case this might cause some trouble somewhere.


DimitriPapadopoulos:
I understand `identify` → `identity`.

Not sure about `*/*;q=0.1` though. Is it required? What sort of trouble do you expect when it is missing?

PR #410, opened 2018-12-07
Use default HTTPS port 443 if none is specified

mss:
I use the NetworkManager plugin and since errors are hidden in syslog, it took me a while to find out what went wrong when I just entered the gateway name without any port information into the config UI. At least according to the documentation I got will the default FortiClient also use port 443 if you don't enter anything special so this patch will make it use that port as well.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -56,7 +56,7 @@
 #endif
 
 #define usage \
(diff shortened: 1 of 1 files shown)


mrbaseman:
looks good to me. I have tested successfully with various combinations (host and port taken from config file, with host and port from command line, with host only from command line, with port from config file being different from 443).

Issue #411, opened 2018-12-07
try_top_auth to send realm info


tcavalca:
First, thanks for the excellent work

I have implemented 2FA using Radius+PrivacyIdea and is working quite well.
For openfortivpn, I have found that "try_otp_auth()" does not send 'realm' information, so connection is dropped even if authentication was successful.

I've made a small patch to fix this - I have not coded for ages so I will appreciate a lot if you guys could come with a more elegant solution.

Let me know if you more information on this

Again, thanks for all the work
Tenorio

PS: adding the patch here for the curious, sorry if I was not supposed to do that
--------- patch begin -----------------
$ cat  http.patch 
--- src/http.c	2018-12-07 13:18:47.516683443 -0400
+++ src/http.c-realm	2018-12-07 16:27:44.840616228 -0400
@@ -500,6 +500,16 @@
 				return -1;
 			url_encode(d, cfg->otp);
 			d += strlen(d);
+			/*  realm workaround */
+			if (cfg->realm[0] != '\0') {
+				l = strlen(cfg->realm);
+				if (!SPACE_AVAILABLE(3 * l + 1))
+					return -1;
+				strcat(d, "&realm=");
+				d += strlen(d);
+				url_encode(d, cfg->realm);
+				d += strlen(d);
+			}
 		} else if (strncmp(t, "submit", 6) == 0) {
 			/* avoid adding another '&' */
 			n = v = e = NULL;
--------- patch end ---------------

tcavalca:
Ops, forgot of markup lang ...

```
$ cat  http.patch 
--- src/http.c	2018-12-07 13:18:47.516683443 -0400
+++ src/http.c-realm	2018-12-07 16:27:44.840616228 -0400
@@ -500,6 +500,16 @@
 				return -1;
 			url_encode(d, cfg->otp);
 			d += strlen(d);
+			/*  realm workaround */
+			if (cfg->realm[0] != '\0') {
+				l = strlen(cfg->realm);
+				if (!SPACE_AVAILABLE(3 * l + 1))
+					return -1;
+				strcat(d, "&realm=");
+				d += strlen(d);
+				url_encode(d, cfg->realm);
+				d += strlen(d);
+			}
 		} else if (strncmp(t, "submit", 6) == 0) {
 			/* avoid adding another '&' */
 			n = v = e = NULL;
```

mrbaseman:
posting the patch here is totally ok. 
[...]

PR #412, opened 2018-12-11
support realm together with 2FA

mrbaseman:
Citing from issue #411:
I have implemented 2FA using Radius+PrivacyIdea and is working quite well.
For openfortivpn, I have found that "try_otp_auth()" does not send 'realm'
information, so connection is dropped even if authentication was successful.

Additional correction of the check for LENGTH_AVAILABLE: Take into account
the length of the additional "&realm="

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -500,6 +500,16 @@ int try_otp_auth(
 				return -1;
 			url_encode(d, cfg->otp);
 			d += strlen(d);
+			/*  realm workaround */
+			if (cfg->realm[0] != '\0') {
+				l = strlen(cfg->realm);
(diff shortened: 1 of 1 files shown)


mrbaseman:
hmm... Codacy warns about usage of `strlen()` because it doesn't handle strings that are not '\0' terminated. Given the fact that the surrounding code relies on `strlen()` as well, I'd postpone that "new" issue to a later code cleanup. Up to now we assume that strings are '\0' terminated and we try to ensure this at various places, for `realm` in [main.c](https://github.com/adrienverge/openfortivpn/blob/master/src/main.c#L303) and [config.c](https://github.com/adrienverge/openfortivpn/blob/master/src/config.c#L211) for instance.

Issue #406, opened 2018-11-30
After openfortivpn 1.8.0 update: problem with networkmanager-fortisslvpn automatic DNS config


SerGIlles:
System: Manjaro x64 (Arch-based Linux).
Everything worked correctly until the update to openfortivpn 1.8.0. Now, it seems that automatic DNS  configuration method changed and it is not compatible with networkmanager-fortisslvpn plugin.
I had to enter my DNS IP manually in order for DNS to work again.
When I use openfortivpn with my config file, everything still works as before, but this is not as handy as using the network manager.

BlueCase:
+1 
Same here
```
02b5e6380,c5248abe-f70d-495c-a6ce-9727608f6a83,"*******",0]: Started the VPN service, PID 16451
02b5e6380,c5248abe-f70d-495c-a6ce-9727608f6a83,"*******",0]: Saw the service appear; activating connection
0x55b02b4e2a60,c5248abe-f70d-495c-a6ce-9727608f6a83]: write: successfully updated (keyfile: update /etc/NetworkManager/system-connections/******* (c5248abe-f70d-495c-a6ce-9727608f6a83,"M>
02b5e6380,c5248abe-f70d-495c-a6ce-9727608f6a83,"*******",0]: Saw the service appear; activating connection
0x55b02b4e2a60,c5248abe-f70d-495c-a6ce-9727608f6a83]: write: successfully updated (keyfile: update /etc/NetworkManager/system-connections/******* (c5248abe-f70d-495c-a6ce-9727608f6a83,"M>
02b5e6380,c5248abe-f70d-495c-a6ce-9727608f6a83,"*******",0]: VPN connection: (ConnectInteractive) reply received
02b5e6380,c5248abe-f70d-495c-a6ce-9727608f6a83,"*******",0]: VPN plugin: state changed: starting (3)



so loaded.



 Ppp device (/org/freedesktop/NetworkManager/Devices/6)
.0.0, 0.0.0.0]


 Ppp device (/org/freedesktop/NetworkManager/Devices/6)
.0.0, 0.0.0.0]





 change: unmanaged -> unavailable (reason 'connection-assumed', sys-iface-state: 'external')
 change: unavailable -> disconnected (reason 'none', sys-iface-state: 'external')
02b5e6380,c5248abe-f70d-495c-a6ce-9727608f6a83,"*******",0]: VPN connection: (IP4 Config Get) reply received from old-style plugin
[...]

Issue #349, opened 2018-07-10
No /etc/resolv.conf entries made by openfortivpn on Opensuse Leap 42.3


mirceadr:
I can ping LAN IPs, so VPN tunnel is ok, I can ping external IPs , so routing is fine.
Any ping on domain names fails . Browser also gets stuck, even Midnight Commander.
/etc/resolv.conf  remains unmodified.

DimitriPapadopoulos:
Which version of _openfortivpn_? How was _openfortivpn_ started? Which options? A log (option `-v`) would help.

DimitriPapadopoulos:
Also could you provide the output of `ls -l /etc/resolv.conf` before/after running _openfortivpn_ ?

DimitriPapadopoulos:
Searching the web, I have also found a few ancient issues with `/etc/resolv.conf` on older versions of _openSUSE_. Perhaps it's unrelated though. Just for the sake of completeness:
* [Fix DNS problem on openSUSE](https://flanaras.wordpress.com/2016/09/26/fix-dns-problem-on-opensuse/)
* [How to fix the DNS in OpenSuSE 13.1 ?](https://goinggnu.wordpress.com/2013/10/14/how-to-fix-the-dns-in-opensuse-13-1/)

mirceadr:
openfortivpn --version  1.7.1 ( rpm based ) 

###

host = 89.238.206.106
port = 10443
username = mircea.dragota
password = XXXXXX
trusted-cert = 87aaa219b54cc00c7765d257bcfc94d326cdb23f8fdf84edcda747fcedea3330 
#set-dns = 0
#pppd-use-peerdns = 0

I tried with set-dns = 1 and pppd-use-peerdns = 1. same results

I ll get back with the log

mirceadr:
DEBUG:  Loaded config file "/etc/openfortivpn/config".
DEBUG:  Config host = "89.238.206.XXX"
DEBUG:  Config realm = ""
DEBUG:  Config port = "104xx"
DEBUG:  Config username = "mircea.dragota"
DEBUG:  Config password = "********"
DEBUG:  server_addr: 89.238.206.XXX
DEBUG:  server_port: 104xx
DEBUG:  gateway_addr: 89.238.206.XXX
DEBUG:  gateway_port: 104xx
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Connected to gateway.
[...]

PR #416, opened 2018-12-13
Revert "Rationalize DNS options"

mrbaseman:
This reverts commit f3830844090d9107112fb83dd93089da4961add1.

See #349 and #406.

With f383084 it is not possible anymore to use _openfortivpn_'s internal mechanism to update  `/etc/resolv.conf`, because it ***also*** needs to pass the `usepeerdns` option to `pppd` in order to receive this information from the vpn server. Here a symmary of the combinations of the options and the result:

New behavior without specific ip-up scripts for the ppp interface or _openresolv_ service:

| set-dns | pppd-use-peerdns | `/var/run/ppp/resolv.conf` | `/etc/resolv.conf` |
| --- | --- | --- | --- | 
| true | true |  created/updated | **not** updated (default case) |
| true | false |  not created/updated | not updated (**broken**) |
| false | true |  **not** created/updated | not updated |
| false | false |  not created/updated | not updated |

old behavior (without specific ip-up scripts for the ppp interface or _openresolv_):

| set-dns | pppd-use-peerdns | `/var/run/ppp/resolv.conf` | `/etc/resolv.conf` |
| --- | --- | --- | --- | 
| true | true |  created/updated | **updated** (default case) |
| true | false |  not created/updated | not updated (**expected but not very intuitive**) |
| false | true |  **created/updated** | not updated (intended for _openresolv_ or `ip-up` script) |
| false | false |  not created/updated | not updated |

With `ip-up` script or _openresolv_, the new default behavior is the required one, with the old behavior `set-dns=0` would be needed. 
Without `ip-up` script or _openresolv_, the old default behavior is the required one, whereas there is **no** combination that can update `/etc/resolv.conf` out of the box.

Edit: restore formatting which was lost during copy&paste

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Examples
 
 * Don't set IP routes and don't add VPN nameservers to `/etc/resolv.conf`:
   ```
-  openfortivpn vpn-gateway:8443 -u foo -p bar --no-routes --no-dns
+  openfortivpn vpn-gateway:8443 -u foo -p bar --no-routes --no-dns --pppd-no-peerdns
   ```
 * Using a config file:
   ```
@@ -39,6 +39,7 @@ Examples
   password = bar
   set-routes = 0
   set-dns = 0
+  pppd-use-peerdns = 0
   # X509 certificate sha256 sum, trust only this one!
   trusted-cert = e46d4aff08ba6914e64daa85bc6112a422fa7ce16631bff0b592a28556f993db
   ```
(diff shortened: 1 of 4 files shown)


DimitriPapadopoulos:
Thank you Martin. I think it's best to release a version 1.8.1 including this commit.

Then we have time to review the options, what they do, and maybe document both these options and the code.

mrbaseman:
for the record, I have created #419 to improve the documentation of the options related to DNS server handling and other documentation issues that I have spotted in the course of looking through the man page and command line options

Issue #415, opened 2018-12-13
Cant connect with openfortivpn-1.8.0-1.fc29.x86_64


mrkarpinski:
Hello,
After upgrading my Fedora from 26 to 29 my pppd connections based on openfortivpn stop working.
I found 1.8.0 testing rpm and upgraded from my 1.6.3
Now when I try to connect from command line or from Network Manager I got this:

DEBUG:  openfortivpn 1.8.0
DEBUG:  Loaded config file "/etc/openfortivpn/config".
DEBUG:  Config host = "xxxxxxxxx"
DEBUG:  Config realm = ""
DEBUG:  Config port = "xxxx"
DEBUG:  Config username = "xxxxxxxxxxxxxxx"
DEBUG:  Config password = "********"
DEBUG:  Resolving gateway host ip
DEBUG:  Establishing ssl connection
DEBUG:  server_addr: xxxxxxxxxxxxxx
DEBUG:  server_port: xxxx
DEBUG:  gateway_addr: xxxxxxxxxxxx
DEBUG:  gateway_port: xxxxxxxxxx
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Connected to gateway.
DEBUG:  Cookie: SVPNCOOKIE=7o0bDj6RhFLGwuknyY+jGvJpiIwq556PAg+aJBKALN/B8aXssmpU4finabySCNj4XZ7us/Mn0OlW87DNYe54/ev3M6I4HtEwXL63G5f+J/Ffx4JOJEH5LaloBG07Gv0i7soGotnuKzdmApT/erR3Wz+M6MY5s1vNicSKFDT2wR7PT4OE4HOCkGluShm+ELypxb3bOYE8oUWRK1JKr/WWsg==
INFO:   Authenticated.
DEBUG:  Cookie: SVPNCOOKIE=7o0bDj6RhFLGwuknyY+jGvJpiIwq556PAg+aJBKALN/B8aXssmpU4finabySCNj4XZ7us/Mn0OlW87DNYe54/ev3M6I4HtEwXL63G5f+J/Ffx4JOJEH5LaloBG07Gv0i7soGotnuKzdmApT/erR3Wz+M6MY5s1vNicSKFDT2wR7PT4OE4HOCkGluShm+ELypxb3bOYE8oUWRK1JKr/WWsg==
INFO:   Remote gateway has allocated a VPN.
[...]

Issue #414, opened 2018-12-11
provide a spec file for CentOS 7


mrbaseman:
I was asked to provide a spec file for a current _openfortivpn_ version. Currently we don't maintain distribution specific files in the upstream repository though. Anyhow, we have example files for ppp configuration in the etc subdirectory. Maybe we can put example spec files in a similar directory, or we could add a wiki page. Here is a proposal for the content of the spec file:
```
Name:           openfortivpn
Version:        1.8.0
Release:        1
Summary:        Client for PPP+SSL VPN tunnel services

Group:          Applications/Internet
License:        GPLv3+
URL:            https://github.com/adrienverge/openfortivpn
Source0:        openfortivpn-%{version}.tar.gz

BuildRequires:  autoconf
BuildRequires:  automake
BuildRequires:  openssl-devel
Requires:       ppp

%description
openfortivpn is a client for PPP+SSL VPN tunnel services. It spawns a pppd
process and operates the communication between the gateway and this process.

It is compatible with Fortinet VPNs.


%prep
%setup -q -n %{name}-%{version}


%build
autoreconf -fi
%configure
make %{?_smp_mflags} V=1


%install
make install DESTDIR=%{buildroot}


%files
%{_bindir}/openfortivpn
%{_mandir}/man1/openfortivpn.1*
%dir %{_sysconfdir}/openfortivpn
%config(noreplace) %{_sysconfdir}/openfortivpn/config
%{_datadir}/openfortivpn/config.template
%doc README.md
%license LICENSE
```
I'm sure there are other suggestions, as for example Fedora and OpenSUSE are rpm based distributions as well, and they provide packages for _openfortivpn_, too. 
Maybe we can provide a minimal template which mostly works on all rpm based platforms? 
However, I haven't looked at the details yet

adrienverge:
I already packaged it for Fedora: https://bodhi.fedoraproject.org/updates/?packages=openfortivpn
The spec file can be copied-pasted from the SRPM :)

> Maybe we can provide a minimal template which mostly works on all rpm based platforms?

I'd rather let distribution files outside the repo, it's a lot of work to maintain. If we start with Fedora/CentOS, we'll have to do Debian-based distributions too, etc.

mrbaseman:
[...]

PR #417, opened 2018-12-13
openfortivpn version 1.8.1

mrbaseman:
We have two important fixes on master now (#416 and #399).

I would also like to land #361 just because it is not a big change, hanging around for such a long time (which causes trouble when the branches have diverged too much as we have seen). 
I have that description already in the updated CHANGELOG.md, so before merging this one we should either merge the fix for long passwords or update the high level changelog once again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,18 @@ Legend
 Releases
 --------
 
-### Upcoming release (current master)
+This high level changelog is usually updated when a release is tagged.
+On the master branch there may be changes that are not (yet) described here.
 
-Note: This list is going to change and perhaps needs some review before the next release is tagged.
+### 1.8.1
 
+* [~] Support longer passowrds by allocation of a larger buffer
+* [-] With version 1.8.0 /etc/resolv.conf was not updated anymore in some situations.
+  To avoid this regression the change "Rationalize DNS options" has been reverted again
+  to restore the behavior of versions up to 1.7.1.
+* [-] Correctly use realm together with two factor authentication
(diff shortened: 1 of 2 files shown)


mrbaseman:
Ok, Codacy complains again about blank separator lines in the changelog... Maybe I switch that check off, too.

mrbaseman:
@adrienverge from my point of view we can tag 1.8.1 now. 

Maybe you have a look at #415 first, if my analysis is correct. 

With the latest patches on master, the rpm builds mentioned in #414 should work again without an additional configure parameter. The ones currently in testing seem to have the problem that in the configure run `pppd` wasn't found and a proper fallback wasn't included yet in 1.8.0.

adrienverge:
@mrbaseman thanks for preparing this release!

I haven't much time these days, so I trust your analysis of #415.

Let's release 1.8.1!

mrbaseman:
thanks @adrienverge for tagging the release. Could you trigger another build of the rpms for Fedora and CentOS, please?

adrienverge:
@mrbaseman done!
https://bodhi.fedoraproject.org/updates/FEDORA-2018-872446dea9
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2018-bb2a09060c

Sorry for the delay, it was a bit cumbersome.

mrbaseman:
oh, is there anything we can improve to make this an easier task?

adrienverge:
Unfortunately no... But you're already doing an amazing job at preparing the releases :+1:

PR #419, opened 2018-12-18
update documentation

mrbaseman:
update help message and man page:
    - indicate the fact that the tcp port may be omitted now (defaults to https)
    - better explanation of --set-dns and --pppd-no-peerdns options
    - mention --ppp-system in man page
    - mention in man page that pppd/ppp options are available only if compiled in
    - update example config file in the man page
    - minor fixes of typos etc.


DimitriPapadopoulos:
@mrbaseman Thank you Martin for documenting the DNS stuff. It really helps.

PR #420, opened 2018-12-19
improve configure output

mrbaseman:
In the rpms for 1.8.1 we did not notice that PPP_PATH was not set properly.
This might have been noticed if there was an output of the value actually used,
so adding those results to the configure output may help avoiding similar
issues in the future.


mrbaseman:
I have just noticed the typo in the commit message. 1.8.1 should have read 1.8.0. I'll push it again with a corrected commit message

Issue #426, opened 2019-02-02
Cannot set route on Raspbian with Rpi


jbremond:
Hi,

I have installed the latest tagged version 1.8.1 on a RPI 3, distribution Raspbian 9.6.
When connecting to my Fortinet SSL Gateway, it seems that openfortivpn cannot set the route.

```
INFO:   negotiation complete
DEBUG:  pppd ---> gateway (6 bytes)
DEBUG:  Got Address: 10.212.134.200
DEBUG:  Interface Name: ppp0
DEBUG:  Interface Addr: 10.212.134.200
INFO:   Interface ppp0 is UP.
INFO:   Setting new routes...
DEBUG:  ip route show to 0.0.0.0/0.0.0.0 dev !ppp0
DEBUG:  ip route show to 192.168.10.14/255.255.255.255 dev ppp0
DEBUG:  Route not found.
DEBUG:  ip route show to 192.168.10.14/255.255.255.255 dev !ppp0
DEBUG:  Setting route to vpn server...
DEBUG:  ip route show to 192.168.10.14/255.255.255.255 via 255.255.255.127 dev enxb827ebdb08e2
DEBUG:  ip route add to 192.168.10.14/255.255.255.255 via 255.255.255.127 dev enxb827ebdb08e2
WARN:   Could not set route to vpn server (Network is unreachable).
DEBUG:  Deleting the current default route...
DEBUG:  ip route del to 0.0.0.0/0.0.0.0 via 255.255.255.127 dev enxb827ebdb08e2
WARN:   Could not delete the current default route (No such process).
DEBUG:  Setting new default route...
DEBUG:  ip route add to 0.0.0.0/0.0.0.0 dev ppp0

```

What is strange is the gateway address that openfortivpn want to add:
`ip route show to 192.168.10.14/255.255.255.255 via 255.255.255.127 dev enxb827ebdb08e2`

255.255.255.127 is not a gateway.

I don't know from where this gateway is calculated, but the problems seems coming from there. Any ideas ?

Thanks,
Joris

DimitriPapadopoulos:
Note sure why it happens, but the relevant messages are printed in _ipv4_protect_tunnel_route_():
https://github.com/adrienverge/openfortivpn/blob/7af5394/src/ipv4.c#L687-L715
[...]

PR #431, opened 2019-03-08
use strtoul to correctly convert to unsigned long

mrbaseman:
this solves #426

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -191,6 +191,7 @@ strstr \
 strtok \
 strtok_r \
 strtol \
+strtoul \
 syslog \
 system \
 tcgetattr \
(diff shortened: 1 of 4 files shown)


DimitriPapadopoulos:
On Linux `sockaddr_in.sin_port` is of type [`in_port_t`](http://man7.org/linux/man-pages/man7/ip.7.html) which is equivalent to [`uint16_t`](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/netinet/in.h.html).

Therefore for signed/unsigned consistency, you might want to change _strtol()_ to _strtoul()_ here as well:
https://github.com/adrienverge/openfortivpn/blob/7af5394/src/tunnel.c#L455

DimitriPapadopoulos:
Other than that, great work!

Issue #427, opened 2019-02-25
One-time password prompt problem


tcrespog:
Hi, I'm using openfortivpn on Linux Mint 19 (Ubuntu 18.04).

I want to connect to a VPN that asks for a code contained in a code card after the user's password is given.
Using the FortiClient SSL VPN client to connect the following output is shown:
> $ /opt/forticlient-sslvpn/64bit/forticlientsslvpn_cli --server ... --vpnuser ...
Password for VPN:
STATUS::Setting up the tunnel
STATUS::Connecting...
...
Enter code 7:

As you can see, the last line asks for a code corresponding to the 7th position in the code card. When I enter the code the connection is established.

However, when I try to connect using openfortivpn, a "one-time password" is asked for, but the code request output is not shown and I'm not able to see what code I must enter:
> openfortivpn ...:443 --username=... --trusted-cert ...
WARN:   Bad port in config file: "0".
VPN account password: 
WARN:   This process was not spawned with root privileges, this will probably not work.
INFO:   Connected to gateway.
Please enter one-time password:

What can I do in order to bypass this problem?

mrbaseman:
When started with '-v -v' all traffic of the authentication is shown in latest versions of openfortivpn. 
Could you capture the http code around the question for 'code 7', please (be sure to remove sensitive information like passwords etc.)?
Then we can include this feature in the future.

tcrespog:
I've realized that the openfortivpn available for the Linux Mint 19 distro repositories is 1.6.0. I've been trying to install 1.8.1 and 1.7.1 from source, but I get the following output after running `./configure`:

> checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking for gcc... gcc
checking whether the C compiler works... no
configure: error: in `/home/user/Programs/openfortivpn-1.8.1':
configure: error: C compiler cannot create executables
See `config.log' for more details

Maybe this is a different issue, but I'm not able to install a recent version due to this. Any ideas?

DimitriPapadopoulos:
You're lacking some “dev” packages that are required to build executables. I'm not familiar with Mint but since it is similar to Debian and Ubuntu you could try:
```
sudo apt install build-essential
```

tcrespog:
[...]

PR #433, opened 2019-03-08
new OTP related options

mrbaseman:
* print out OTP prompt starting with 'Enter' (which may contain the index of the code in a list) see #427
* add a config file option (and command line parameter) for that
* rebased the --otp-delay option which is still open in #376

### DIFF
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -10,6 +10,7 @@ openfortivpn \- Client for PPP+SSL VPN tunnel services
 [\fB\-p\fR \fI<pass>\fR]
 [\fB\-\-otp=\fI<otp>\fR]
 [\fB\-\-otp\-prompt=\fI<prompt>\fR]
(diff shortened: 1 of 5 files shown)


mrbaseman:
We could drop f3b820e again, now that 2fde0a5 offers a more flexible solution, but for testing I have left this commit in my branch
The code changes are at the same places as the ones in #376 so they cause conflicts, so I have cherry-picked the commit from there and merged it manually. 
For #427 I feel we should provide a solution and when I had the solution ready and looked at #376 I have noticed that the changes are not only in the same code regions but in fact very similar. So it would be quite unfair to decline the latter one and include a solution for the other one.

DimitriPapadopoulos:
:+1:

mrbaseman:
sorry for the noise of all those commits. I have squashed the changes into two commits, one for the `--otp-prompt` and one for `--otp-delay` and later on squashed a few more updates for the man page into the first one.

PR #439, opened 2019-04-01
userinput: add capability of spawning a helper to get a secret

lkundrak:
This adds the --pinentry option that allows passing a name of a program
that can supply a password. This is going to be useful in order to relay
the password requestes to the NetworkManager plugin.

The protocol used is compatible with what GnuPG's pinentry program uses
(documented in pinentry GNU Info manual), so that this could be tested
or used without the NetworkManager plugin.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -61,6 +61,7 @@
 
 #define usage \
 "Usage: openfortivpn [<host>[:<port>]] [-u <user>] [-p <pass>]\n" \
+"                    [--pinentry=<program>]\n" \
 "                    [--realm=<realm>] [--otp=<otp>] [--otp-delay=<delay>]\n" \
 "                    [--otp-prompt=<prompt>] [--set-routes=<0|1>]\n" \
 "                    [--half-internet-routes=<0|1>] [--set-dns=<0|1>]\n" \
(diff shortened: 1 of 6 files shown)


src/main.c
```diff
@@ -61,6 +61,7 @@
 
 #define usage \
 "Usage: openfortivpn [<host>[:<port>]] [-u <user>] [-p <pass>]\n" \
+"                    [--pinentry=<progran>]\n" \
```
madscientist: s/progran/program/

lkundrak: Thanks, fixed.

adrienverge:
@lkundrak thanks for your contribution.
Is `--pinentry` standard? I mean is it general enough to work with other interfaces than a NetworkManager plugin?

lkundrak:
> @lkundrak thanks for your contribution.
> Is `--pinentry` standard?

Sort of. It's somewhat documented. As I try to make it clear in the commit message, I've used this to avoid inventing a new protocol.

> I mean is it general enough to work with other interfaces than a NetworkManager plugin?

Yes.

> I tried this but I was not able to move forward. I tried --pinentry pinentry-gnome3 but no widow for entering the password is shown. After some debugging I found that pinentry returns
> S ERROR gnome3.isatty 83918950 \nERR 83918950 Inappropriate ioctl
> 
> But I obviously have a DISPLAY set.

My guess is that you're running the helper as root, which is why it can't talk to the Gcr secrets service.

Try pinentry-gtk or pinentry-qt.

I've improved the diagnostics.

emelenas:
Yes, now it asks for password and 2FA sequentially

lkundrak:
Anyone minds if I merge this?

adrienverge:
@DimitriPapadopoulos @mrbaseman no objection?

mrbaseman:
@lkundrak it's fine for me

mrbaseman:

PR #441, opened 2019-04-08
Connect to old TLSv1.0 software - override new openssl defaults.

ebikt:
Openssl 1.1 sets min tls version to 1.2. This prevents me to connect to fortivpn at my work.
I do not want to fiddle with system defaults, thus I added min_tls option. And then I added
seclevel-1 option, because I realised, that compiled-in list of ciphers is sufficient for me.

Inspired by similar wpa_supplicant bug, triggered by libssl1.1
  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=907518#95
  https://w1.fi/cgit/hostap/commit/?id=cc9c4feccc5588137f66c40a4a6729476556853e

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -25,6 +25,7 @@
 #include <sys/stat.h>
 #include <limits.h>
 #include <openssl/tls1.h>
+#include <openssl/x509.h.h>  /* work around OpenSSL bug, missing STACK_OF definition */
 
 const struct vpn_config invalid_cfg = {
 	.gateway_host = {'\0'},
--- a/src/config.c
+++ b/src/config.c
@@ -133,8 +133,8 @@ int parse_min_tls(const char *str)
 	case '2':
 		return TLS1_2_VERSION;
 #if OPENSSL_VERSION_NUMBER < 0x020000000L
-	/* 
-	 * libressl uses version numbers starting with major version 2 
+	/*
+	 * libressl uses version numbers starting with major version 2
 	 * but does not yet support TLS 1.3
 	 */
 	case '3':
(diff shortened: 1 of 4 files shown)


mrbaseman:
hmm... the build fails, at least with the openssl version that travis uses. Perhaps an explicit 
`#include <openssl/ssl.h>`
is missing before including `tls1.h`

DimitriPapadopoulos:
About this error:
```
/home/travis/.openfortivpn-deps/include/openssl/tls1.h:331:52: error: unknown type name ‘STACK_OF’
 int SSL_check_chain(SSL *s, X509 *x, EVP_PKEY *pk, STACK_OF(X509) *chain);
                                                    ^
```

This looks like an OpenSSL bug on whatever platform _travis_ use to build. Identical error in other software:
* [Issue #7314](https://freeswitch.org/jira/si/jira.issueviews:issue-html/FS-7314/) in _FreeSWICTH_.
* [Build error](https://karkhaz.github.io/tuscan/vanilla/sofia-sip.html) in _sofia_sip_.

This seems to have been fixed in _OpenSSL_ by this patch:
https://github.com/openssl/openssl/commit/e30dff9e
The best workaround is probably to precisely add the missing include directly in _config.c_:
`# include <openssl/buffer.h>`

mrbaseman:
I have noticed that this PR has broken building against libressl, e.g. on FreeBSD. I have pushed a fix for that. From my point of view we can merge this PR now. @DimitriPapadopoulos do you have anything to add?

DimitriPapadopoulos:
@mrbaseman Nothing to add, the part I understand looks OK to me. I suggest you merge.

mrbaseman:
we have introduced a problem when linking against openssl 1.1.0 with this. 
TLS 1.3 is available not before 1.1.1. Fix is in #452 together with another openssl fix.

mrbaseman:

PR #443, opened 2019-05-10
suppress cleartext password in debug detail output

mrbaseman:
...and introduce another log level "debug all" to even show the password
for debugging encoding stuff. Currently only the password is cleared,
otp, cookie and perhaps other information during the authentication
process is considered less critical.


mrbaseman:

PR #444, opened 2019-05-10
increase speed setting for pppd

mrbaseman:
this might be helpful for #428


DimitriPapadopoulos:
Does _pppd_ use the _speed_ parameter when there is no modem? In any case it shouldn't hurt to change to 115200. Still, it would be great to have feedback from Mac users.

I suggest you change the commit description to explain the reason for the change is to hopefully fix speed issues on Mac.

mrbaseman:

PR #445, opened 2019-05-18
work around EAGAIN issue on FreeBSD

dhke:
FreeBSD seems to have the issue that `write()` can return `EAGAIN` even when an immediately preceding `select()` indicated a writable socket.

Work around this issue by retrying.

Note that this creates a potential busy loop around failing `write()`, which this change does nothing to avoid.


mrbaseman:

PR #447, opened 2019-05-31
Redirect all traffic to vpn and don't honour given split rules if half_internet_routes is set to zero

Achal-Aggarwal:
…

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -912,9 +912,10 @@ static int ipv4_set_default_routes(struct tunnel *tunnel)
 
 int ipv4_set_tunnel_routes(struct tunnel *tunnel)
 {
-	int ret = ipv4_protect_tunnel_route(tunnel);
+	struct vpn_config *cfg = tunnel->config;
 
-	if (tunnel->ipv4.split_routes)
+	int ret = ipv4_protect_tunnel_route(tunnel);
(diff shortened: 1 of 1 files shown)


mrbaseman:
there are two modes for the ssl vpn on a Fortigate: Tunnel mode and split route mode. The mode can be configured on the server-side and the client redirects all traffic through the tunnel if tunnel mode is configured or receives specific routes from the server side in the case of split route mode.

If you want to override the routing on the client side, you can do so by calling _openfortivpn_ with the `--no-routes` flag and manually add a default route through the ppp interface. Don't forget to add a host-route to the vpn gateway before you cut off the branch you are sitting on ;)

If you want to automate the rounting setup there are examples for `pppd` ip-up scripts [on our wiki page](https://github.com/adrienverge/openfortivpn/wiki#how-to-add-specific-routes-using-pppd)

The half-interent-routes mode is just another way to set up the tunnel routing mode and avoid deleting the original default route. This option does not apply to the split routes mode. With the current pull request you seem to abuse that switch to enforce the tunnel mode if it is not advertised by the vpn. The problem with this is that the default value for this setting is 0 which means that your patch by default disables the split mode even if it is configured on the server side, unless the user explicitly changes the setting that you abuse here for a different purpose.

I'd suggest that we close this, because the change would enforce unexpected behavior and the desired feature can be achieved by already existing mechanisms.

Issue #448, opened 2019-06-11
openssl 1.1.1 support


mrbaseman:
when openssl is not build to include deprecated APIs _openfortivpn_ can not be linked against openssl 1.1.1

see https://github.com/adrienverge/openfortivpn/issues/446#issuecomment-500248050

mrbaseman:
this was  in fact a minor issue: starting with openssl 1.1.0 it is not necessary anymore to initialize the library and the error strings explicitly.

mrbaseman:
fix is landed on master

mrbaseman:
this went into the [1.10.0 Release](https://github.com/adrienverge/openfortivpn/releases/tag/v1.10.0).

Just as an explanation: Openssl 1.1.0 onwards defines OPENSSL_API_COMPAT as OPENSSL_MIN_API in `<openssl/opensslconf.h>` if it is not set prior to the inclusion of the ssl headers. 
We pick up this value after inclusion of the ssl headers and base our decision if the library needs an explicit initialization on this value.

Users can enforce a different API compatibility level e.g. at configure time:
`CPPFLAGS=-DOPENSSL_API_COMPAT=0x10101000L ./configure ...`

PR #452, opened 2019-06-14
fix openssl 1.1.x compatibility issues (#448)

mrbaseman:
starting with openssl 1.1.0 the library and error strings are
initialized automatically and no explicit initialization is needed anymore

in openssl 1.1.0 tls 1.3 is not yet present, so instead of comparing
version numbers it's better just to check if TLS1_3_VERSION is defined


mrbaseman:
successfully tested on CentOS 7.6, SL 6.10, SLES 12sp3, Ubuntu 16.04, FreeBSD 11, MacOS X 10.11

Issue #351, opened 2018-07-17
Could not authenticate to gateway


jprrezende:
Hi, I am using version 1.7.1 on Debian 9.
I am receiving authentication error, but, I tested using FortiClient SSLVPN 4.0.2333 for Linux and it is working.

I debugged:

```
joaopaulo@pc-jp:~/Downloads/openfortivpn$ sudo openfortivpn -v -c /etc/openfortivpn/config
DEBUG:  Loaded config file "/etc/openfortivpn/config".
VPN account password: 
DEBUG:  Config host = "remote.stategrid.com.br"
DEBUG:  Config realm = ""
DEBUG:  Config port = "10443"
DEBUG:  Config username = "JOAO PAULO RAMOS REZENDE"
DEBUG:  Config password = "********"
DEBUG:  server_addr: xxx.xxx.xxx.xxx
DEBUG:  server_port: 10443
DEBUG:  gateway_addr: xxx.xxx.xxx.xxx
DEBUG:  gateway_port: 10443
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Connected to gateway.
ERROR:  Could not authenticate to gateway (No cookie given).
INFO:   Closed connection to gateway.
DEBUG:  server_addr: xxx.xxx.xxx.xxx
DEBUG:  server_port: 10443
DEBUG:  gateway_addr: xxx.xxx.xxx.xxx
DEBUG:  gateway_port: 10443
DEBUG:  Gateway certificate validation failed.
DEBUG:  Gateway certificate digest found in white list.
INFO:   Logged out.
```
So, is it a bug or I am doing something wrong?
Thanks in advance.

mrbaseman:
Hi,

typical reasons for this error are 
* typo in user name or password
* wrong configuration on the fortigate you are trying to connect to (there seem to be some problems when groups are configured, but clients usually specify users, and in my experience each user should be configured separately in the ssl vpn portal)
* expired certificate if user certificates are in use
* restriction to specific IPs while you are coming from elsewhere
the ssl vpn logs on the fortigate may give a hint, but the logged errors sometimes are also misleading ( at least in fortios 5.2 I have seen "unknown user" in the logs, when the user existed, but tried to login with wrong credentials. 
see also #350, #267 and #49

DimitriPapadopoulos:
[...]

PR #457, opened 2019-07-24
fix (http request): pass single space as-is

amanjeev:
#351 : a very dumb fix.

I am sure I am not the first one to think about this but I was facing the same issue where my username at work has a space in it. I tried by passing space as-is and it actually worked for me. I could create the tunnel!

Please feel free to close etc. if need be.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -43,10 +43,8 @@ static void url_encode(char *dest, const char *str)
 {
 	while (*str != '\0') {
 		if (isalnum(*str) || *str == '-' || *str == '_'
-		    || *str == '.' || *str == '~')
+		    || *str == '.' || *str == '~' || *str == ' ')
 			*dest++ = *str;
-		else if (*str == ' ')
-			*dest++ = '+';
 		else {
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
We may choose not to encode spaces if it works. Another possibility: use an alternative encoding, `%20` instead of `+`. Could you try this alternative encoding and report back?

DimitriPapadopoulos:
All HTTP requests to the FortiGate device seem to be using the `application/x-www-form-urlencoded` media type. As far as I can tell such `GET`/`POST` requets should encode spaces using the obsolete `+` form. It could be that FortiGate devices expect the alternative `%20` form. Or maybe plain ` ` spaces.

amanjeev:
Please see #458

mrbaseman:
from the current perspective I believe #458 is the better solution

amanjeev:
Please feel free to close this.

mrbaseman:
the solution from #458 went into the [1.11.0 Release](https://github.com/adrienverge/openfortivpn/releases/tag/v1.11.0).

PR #458, opened 2019-07-24
fix (http request): pass single space with encoding

amanjeev:
Part 2 with encoding. This works too for me.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -45,8 +45,11 @@ static void url_encode(char *dest, const char *str)
 		if (isalnum(*str) || *str == '-' || *str == '_'
 		    || *str == '.' || *str == '~')
 			*dest++ = *str;
-		else if (*str == ' ')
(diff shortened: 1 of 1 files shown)


cfhammill:
This fixed things for me, thanks so much!

DimitriPapadopoulos:
@mrbaseman Martin, what is your take on this pull request? I think we should merge. Not exactly HTTP compliant but if I understand you correctly that's what FortiGate appliances are expecting.

mrbaseman:
Yes, we have some positive feedback at least, so let's merge it. Hopefully the encoding hasn't changed between the different FortiOS versions, and if so, we have successfully tested with the recent ones.

mrbaseman:
this went into the [1.11.0 Release](https://github.com/adrienverge/openfortivpn/releases/tag/v1.11.0).

PR #466, opened 2019-09-04
one-byte fix to build with lcc compiler

mrbaseman:
I have received this from Mike: --pedantic should correctly be -pedantic


mrbaseman:

Issue #470, opened 2019-09-16
Issues connecting from Mojave


madrover:
Hi.
I am using version 1.10.0 from Macos Mojave and, even though the VPN seems to connect, I can not access any service exposed on the VPN and the VPN stops in 20-30 seconds approximately.

The command output is:

> WARN:   You should not pass the password on the command line. Type it interactively or use a config file instead.
WARN:   Bad port in config file: "0".
INFO:   Connected to gateway.
INFO:   Authenticated.
INFO:   Remote gateway has allocated a VPN.
INFO:   No split VPN route
Mon Sep 16 15:10:50 2019 : publish_entry SCDSet() failed: Success!
Mon Sep 16 15:10:50 2019 : publish_entry SCDSet() failed: Success!
Mon Sep 16 15:10:50 2019 : Using interface ppp0
Mon Sep 16 15:10:50 2019 : Connect: ppp0 <--> /dev/ttys003
INFO:   Got addresses: [172.28.133.28], ns [10.215.137.194, 10.215.137.129]
INFO:   negotiation complete
INFO:   negotiation complete
Mon Sep 16 15:10:51 2019 : local  IP address 172.28.133.28
Mon Sep 16 15:10:51 2019 : remote IP address 192.0.2.1
Mon Sep 16 15:10:51 2019 : primary   DNS address 10.215.137.194
Mon Sep 16 15:10:51 2019 : secondary DNS address 10.215.137.129
Mon Sep 16 15:10:51 2019 : Committed PPP store
Mon Sep 16 15:10:51 2019 : Committed PPP store
INFO:   Interface ppp0 is UP.
INFO:   Setting new routes...
add host 195.76.89.2: gateway en0
delete net 0.0.0.0: gateway 192.168.1.1
add net 0.0.0.0: gateway ppp0
INFO:   Adding VPN nameservers...
INFO:   Tunnel is up and running.
INFO:   Cancelling threads...
INFO:   Setting ppp interface down.
INFO:   Restoring routes...
delete host 195.76.89.2: gateway en0
delete net 0.0.0.0: gateway ppp0
add net 0.0.0.0: gateway 192.168.1.1
INFO:   Removing VPN nameservers...
Mon Sep 16 15:11:12 2019 : Hangup (SIGHUP)
Mon Sep 16 15:11:12 2019 : Modem hangup
Mon Sep 16 15:11:12 2019 : Connection terminated.
Mon Sep 16 15:11:12 2019 : LCP close (User request).
[...]

PR #471, opened 2019-09-19
correctly shift masks for cidr notation on MAC

mrbaseman:
fix for #470

when parsing routing entries with cidr notation, the calculated
mask was shifted in the wrong direction. This has lead to the
result that ipv4_get_route has returned the wrong route in some
cases.


mrbaseman:
It's still not correct. The /4 mask hits the correct byte now, but the wrong bits of it. 224 is 0xe0 and we want to mask it with 0xf0 but the current code uses mask 0xf

mrbaseman:

PR #483, opened 2019-10-23
support dns-suffix feature

mrbaseman:
currently we just prepend a search line to resolv.conf when adding nameservers
only the first nameservers take effect and only the last search statement
openfortivpn competes with pppd when manipulating that file

### DIFF
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -91,12 +91,14 @@ Set if openfortivpn should add two 0.0.0.0/1 and 128.0.0.0/1 routes with
 higher priority instead of replacing the default route.
 .TP
 \fB\-\-set\-dns=\fI<bool>\fR, \fB\-\-no\-dns\fR
-Set if openfortivpn should add VPN nameservers in /etc/resolv.conf when
+Set if openfortivpn should add DNS name servers in /etc/resolv.conf when
 tunnel is up. If used multiple times, the last one takes priority.
 This option requires that the DNS entries are requested from the peer.
 So, \fB\-\-pppd\-no\-peerdns\fR conflicts with \fB\-\-set\-dns=\fI1\fR.
 Note that there may be other mechanisms to update /etc/resolv.conf
 which may require that openfortivpn is called with \fB\-\-no\-dns\fR.
+Also a dns\-suffix may be received from the peer and added to
+/etc/resolv.conf in the turn of adding the name servers.
 
 \fB\-\-no\-dns\fR is the same as \fB\-\-set\-dns=\fI0\fR.
 .TP
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -1,4 +1,4 @@
-.TH OPENFORTIVPN 1 "March 12, 2019" ""
+.TH OPENFORTIVPN 1 "October 24, 2019" ""
 
 .SH NAME
 openfortivpn \- Client for PPP+SSL VPN tunnel services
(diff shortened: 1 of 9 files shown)


mrbaseman:
While working on this I think I have found why #330 did not work as expected. With https://github.com/adrienverge/openfortivpn/pull/483/commits/60ab6d7b6ae5a760560bb5feb672b18d24e60056 I have corrected that. I have also changed the default for `pppd-use-peerdns` now that `use-dns` works independent from pppd. 
I have updated the man page and the help message.

I would propose **not** to squash the two commits during merge. If you prefer, or if we need a separate discussion, I can split this up into two pull requests.

mrbaseman:
I have just tested this with all my environments. It works as expected, at lest for my setups. 
By the way, I have updated my forticlient and could compare the behavior now. dns-suffix must have been one of the last features that have been implemented in the official linux vpn client...

DimitriPapadopoulos:
I agree, two separate commits for DNS domain and DNS servers. No need for a separate discussion or pull request though.

mrbaseman:
@DimitriPapadopoulos you have raised good points, and I have addressed them in https://github.com/adrienverge/openfortivpn/pull/483/commits/3e71563356a31d6571cc63520115abc6a7079755 
I'm not sure if we shall leave these corrections as a separate commit or if I should merge some of them into the first one (but then the second commit doesn't apply anymore cleanly). On the other hand, having two separate branches for individual pull requests wouldn't solve the problem.

DimitriPapadopoulos:
We need two commits, ideally in this order:
1. take into account the domain search list
2. modify the way we handle name servers.
Then we can release a new version - perhaps we should handle Coverity warnings first. If needed we can revert the name servers stuff easily.

Is it possible to rewrite the history to have to commits?

Perhaps creating two branches / pull requests can help, it's easier to review apply small changes.

mrbaseman:
I have merged the corrections into the original commits. Now we have two commits like you have proposed, @DimitriPapadopoulos 
(To achieve this, I have created a second local branch, then rewound the branch for the PR, cherry-picked the fixes from the local branch and manually thrown out everything that belongs to the second commit, squashed those two, then squashed all new commits in the local branch, rebased it on the original branch, and finally cherry-picked that resulting commit in the original branch).

DimitriPapadopoulos:
One last question. I'm wondering whether the VPN appliance may request multiple domains in the search list.

mrbaseman:
Yes, it can. Actually, the dns-suffix setting is a user defined string. If you want to send several domains, you can just supply a space separated list which can be directly used in the search statement of `resolv.conf`.

mrbaseman:
About the new release: I would like to address the Coverity warnings, improve the Readme a bit, and also do extensive testing this time. We have addend Smart card support, I have started another attempt to rationalize the dns options, but although I'm quite confident we should be careful this time

DimitriPapadopoulos:
If the dns-suffix setting can contain multiple domains, then it can be longer than 256.

mrbaseman:
In my understanding the whole string  is limited to 256 characters (https://unix.stackexchange.com/questions/245849/resolv-conf-limited-to-six-domains-with-a-total-of-256-characters).

DimitriPapadopoulos:
Perfect then. Perhaps you can add this link as a comment in the source code or in the commit itself. Other than that I suggest we merge this pull request - this will help testing.

mrbaseman:
I have added a comment in xml.h where MAX_DOMAIN_LENGTH is defined, squashed that change into the first commit, and then merged the PR as you have suggested for making testing easier

mrbaseman:

Issue #484, opened 2019-10-24
Coverity oustanding issues


DimitriPapadopoulos:
There are currently 7 Coverity issues. Some of them look like false positives, others may be theoretical issues that won't happen in real life. In any case, I suggest rewriting the relevant code to obtain less complex sources, easier to review (both for Coverity and human reviewers).

Especially reading/parsing file descriptors results in complex code. Would it make sense to use a parser generator such as ANTLR to simplify the code?

@mrbaseman Do you have access to the Coverity issues?

mrbaseman:
I have access, yes. These issues are not related to #483, right? They have either been there for a longer time or have been introduced by the other pull requests that we have merged in the last months.

DimitriPapadopoulos:
Most issues have been introduced by the other pull requests that we have merged in the last months. I haven't tested #483.

mrbaseman:
I had only 5 issues in the Dashboard. I guess two others are related to these ones and therfore not listed separately. 
Could you have a look at #489 please, and check if coverity is happy with those fixes?

DimitriPapadopoulos:
Unfortunately Coverity Scan seems to have problems right now. There were 304 builds ahead of _openfortivpn_ yesterday and 298 builds today. It might take weeks to obtain a result.

A short note on your changes: the macro _assert()_ generates no code when `NDEBUG` is defined. Perhaps it's best to shut up Coverity with code that is executed whatever the context, using our usual error management routines.

mrbaseman:
I have added a new commit to avoid _assert_. Do you know if Coverity is going to pick this up, or do we have to re-schedule another scan?

DimitriPapadopoulos:
I'll re-schedule a new manual scan. Automatic scans are supposed to happen on the `coverity_scan` branch of the `adrienverge` repository, as suggested by the Coverity documentation, but I seem to recall I haven't been able to finalize the setup.

In any case the Coverity Scan infrastructure seems broken again - last time it had occurred it had taken months to restore the Coverity Scan service.

DimitriPapadopoulos:
The previous scan is over. I haven't received a notification by mail, perhaps the scan server was in a real bad shape and they had to flush the queue / reboot it.

Any way, I have just submitted your branch mrbaseman/coverity for scanning: 72 builds ahead of it.

DimitriPapadopoulos:
Done. There are 3 issues left.

mrbaseman:
[...]

Issue #347, opened 2018-07-09
Use 3rd party HTTP parsing library?


DimitriPapadopoulos:
Would it make sense to use a 3rd party HTTP parsing library instead of maintaining our own code?

Examples of 3rd party libraries:
* [http-parser](https://github.com/nodejs/http-parser) - no syscalls or memory allocation
* [picohttpparser](https://github.com/h2o/picohttpparser) - no memory allocation, ×5 faster than [http-parser](https://github.com/nodejs/http-parser)

mrbaseman:
hmm... for what reason? 
* more features / flexibility that a 3rd party library provides? 
* issues we have/had with our own code which are properly handled there (if so, which ones?)? 
* just to shrink our code and do outsourcing of this particular task? 

personally, if the code for a particular feature is not too large, I prefer to maintain it locally for the following reasons:

* it's probably easier to understand and debug than a "black box" that is supposed to do a particular task
* we stay flexible to extend it on demand according to our requirements
* it's more portable than external libraries that might not exist on some platform

I'm not strictly against 3rd party libraries, just expressing my vague feelings about this topic

DimitriPapadopoulos:
Only this one:
* just to shrink our code and do outsourcing of this particular task

Both HTTP parsers are at least as portable as _openfortivpn_. Just and idea though, not a necessity.

mrbaseman:
hmm... we would replace (large) parts of [http.c](https://github.com/adrienverge/openfortivpn/blob/master/src/http.c). But we would presumably bundle the parser with _openfortivpn_, and redistribute it together with our source, right?

DimitriPapadopoulos:
Yes, that's what I had in mind. Linux packages might want to link to a library though.

mrbaseman:
good point: if the external library provides more detailed error messages. On the other hand one could just print out `res` to stderr before freeing it [here](https://github.com/adrienverge/openfortivpn/blob/2aca389/src/http.c#L571), at least in the verbose mode. 
But maybe this is just one more advantage of an external library, and there are many more if we just look close enough...

mrbaseman:
string parsing could also be done by an external library (see #502)

DimitriPapadopoulos:
#575: one more reason to use external files/libraries for everything that is not core functionality: HTTP or XML parsing

DimitriPapadopoulos:
[...]

PR #489, opened 2019-11-05
address coverity issues (fixes for #484)

mrbaseman:


mrbaseman:
Perhaps we should add more details to the commit message, but I was in a hurry, writing from my mobile now

mrbaseman:
@DimitriPapadopoulos could you please trigger another run with https://github.com/adrienverge/openfortivpn/pull/489/commits/92968856f6b5c9d9363121cf1e246ea7c278806d now?

DimitriPapadopoulos:
I've started a new run with c59d54473d. Two defects have been eliminated, one defect left.
Great work! I'm sorry I cannot contributing much these days.

DimitriPapadopoulos:
Can/should we eliminate the last defect?

mrbaseman:
The routing table [is read in](https://github.com/adrienverge/openfortivpn/blob/f2f084602971426ec45f7c4753240563d858a9f5/src/ipv4.c#L150) using `read()` and therefore `start` [is considered tainted](https://github.com/adrienverge/openfortivpn/blob/f2f084602971426ec45f7c4753240563d858a9f5/src/ipv4.c#L347). `strtok_r` internally uses it in a loop boundary. I think the theoretical problem is that the loop could iterate forever, if `start` contains a huge string which doesn't contain the token character. 

We can declare this as a false positive, because the routing table is to some extent trusted input, it's not that large, and the loop in `strtok_r` increments the pointer in each interation until it reaches the area where we have ensured that there is a delimiting `'\0'` character now by proper initialization. 

That's at least how I understand it by now. Perhaps we should add this explanation in a comment close to the initialization of the buffer.

mrbaseman:
I have rebased and squashed my commits

src/ipv4.c
```diff
@@ -122,6 +122,25 @@ static int ipv4_get_route(struct rtentry *route)
 	uint32_t rtdest, rtmask, rtgtw;
 	int rtfound = 0;
 
+	/* initialize the buffer with zeroes, aiming to address the
+	 * coverity issue "tainted_skalar passed to a tainted sink"
```
DimitriPapadopoulos: A minor typo here: skalar → scalar

I suggest either of:
"tainted scalar passed to a tainted sink"
"TAINTED_SCALAR passed to a tainted sink"

src/ipv4.c
```diff
@@ -122,6 +122,25 @@ static int ipv4_get_route(struct rtentry *route)
...
+	 * the token character, which we are parsing for.
+	 *
+	 * We can declare this as a false positive, because
+	 * - the routing table is to some extent trusted input,
+	 * - it's not that large,
+	 * - and the loop in strtok_r increments the pointer in each
+	 *   interation until it reaches the area where we have ensured
+	 *   that there is a delimiting '\0' character by proper
+	 *   initialization. We ensure this also when growing the buffer.
+	 */
```
DimitriPapadopoulos: Agreed.

That said perhaps we should use a 3rd party parser generator (some _yacc_ alternative?) instead of parsing everything manually. It would probably avoid such warnings. In the same way HTML parsing could use a 3rd party library as already suggested. Indeed I think most Coverity issues were found in parsing code. That would be for an upcoming release though...

src/tunnel.c
```diff
@@ -158,6 +168,11 @@ static int pppd_run(struct tunnel *tunnel)
 
 	slave_stderr = dup(STDERR_FILENO);
 
+	if (slave_stderr < 0) {
+		log_error("slave stderr %d\n", slave_stderr);
```
DimitriPapadopoulos: I suggest using `strerror(errno)` in the error message.

src/userinput.c
```diff
@@ -42,6 +42,8 @@ static char *uri_escape(const char *string)
 		if (allocated_len + 4 >= real_len) {
 			allocated_len += 16;
 			escaped = realloc(escaped, allocated_len);
+			// bail out if realloc fails
+			if (escaped == NULL) return NULL;
```
DimitriPapadopoulos: The [Linux kernel coding style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces) seems to suggest `if` and `return` should be split into two lines. I agree that's unclear but that's the style all examples use.

src/ipv4.c
mrbaseman: I have picked the second option

src/ipv4.c
mrbaseman: yes, that would be a bigger change. We could add this parsing topic to #347 or open a separate issue so that we don't forget it...

src/tunnel.c
mrbaseman: done.

src/userinput.c
mrbaseman: done.

mrbaseman:
I think I have seen the option to mark individual issues as false positives in the web interface, but I'm not sure anymore if it was codacy or coverity

mrbaseman:
I have found the option to ignore specific issues in the "Triage" tab

Issue #375, opened 2018-10-04
failed to connect just with a certificate


claudiuion:
Our server supports only authentication with a certificate. We don't have usernames and passwords. Is it possible to bypass the verification for a username and pass?
The error that I receive is:
VPN connection: failed to connect: 'Missing required option “user”.

mrbaseman:
I don't find the error message in the source (at least not the current one). Are you using the network manager applet?
But anyhow, it's true that openfortivpn currently requires that username and password are specified. However, maybe it works fine if just empty values for username and password are passed to the server. You could give it a try to comment out [lines 403-416 in main.c](https://github.com/adrienverge/openfortivpn/blob/master/src/main.c#L403).

claudiuion:
Yes I'm using the network manager applet. Thanks for the tip, I'll give it a try.

DimitriPapadopoulos:
The error in #366 is probably unrelated (it's not a Fortinet gateway).

DimitriPapadopoulos:
It would be best to provide your own verbose log (use option `openfortivpn -v`).

claudiuion:
DEBUG:  openfortivpn 1.7.1
DEBUG:  Loaded config file "config".
DEBUG:  Config host = "xxx.xxx.xxx.xxx"
DEBUG:  Config realm = ""
DEBUG:  Config port = "443"
DEBUG:  Config username = ""
DEBUG:  Config password = "********"
DEBUG:  server_addr: xxx.xxx.xxx.xxx
DEBUG:  server_port: 443
DEBUG:  gateway_addr: xxx.xxx.xxx.xxx
DEBUG:  gateway_port: 443
Enter PEM pass phrase:
DEBUG:  Gateway certificate validation succeeded.
INFO:   Connected to gateway.
ERROR:  Could not authenticate to gateway (Bad HTTP response code).
INFO:   Closed connection to gateway.
DEBUG:  server_addr: xxx.xxx.xxx.xxx
DEBUG:  server_port: 443
DEBUG:  gateway_addr: xxx.xxx.xxx.xxx
DEBUG:  gateway_port: 443
Enter PEM pass phrase:
DEBUG:  Gateway certificate validation succeeded.
INFO:   Logged out.

mrbaseman:
Ok, it could not authenticate you with an empty password. I can't test it easily, because local users always have a password, so your account is probably an external one (LDAP or Radius, ...). 

[...]

PR #493, opened 2019-11-27
allow logging in with a certificate only

mrbaseman:
...and with password set to an empty string, which denotes that no password shall be asked or sent
allow username to be empty for smartcard login
update man page and readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Check [#464](https://github.com/adrienverge/openfortivpn/issues/464) for a discu
 of known issues in this area.
 
 To make use of your smartcard put at least `pkcs11:` to the user-cert config or commandline
-option. It takes the full or a partial PKCS#11 token URI. 
+option. It takes the full or a partial PKCS#11 token URI.
 
 ```
 user-cert = pkcs11:
--- a/README.md
+++ b/README.md
@@ -55,8 +55,7 @@ Check [#464](https://github.com/adrienverge/openfortivpn/issues/464) for a discu
 of known issues in this area.
 
 To make use of your smartcard put at least `pkcs11:` to the user-cert config or commandline
-option. It takes the full or a partial PKCS#11 token URI. Also username and password currently
-may not be empty, but don't get used. So you should fill in dummy values.
+option. It takes the full or a partial PKCS#11 token URI. 
 
 ```
 user-cert = pkcs11:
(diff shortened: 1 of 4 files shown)


mrbaseman:
In #491 we were discussing if a dummy username is really needed. 

I had a similar branch in my repo for testing in the context of #375 

Although we didn't reach a final solution there, we could merge those lines, together with a few modifications that I have already squashed in so that we don't require dummy variables for the smartcard login.

mrbaseman:
@DimitriPapadopoulos could you run Coverity again on this PR, please? To make sure we don't introduce new defects right after we have cleaned up everything

DimitriPapadopoulos:
@mrbaseman There are a few warnings about _printf()_ that are easy to fix but I'm puzzled by the other errors: have they been introduced by the new code or had they been previously missed by Coverity?

mrbaseman:
the [new changes](https://github.com/adrienverge/openfortivpn/pull/493/files) aren't that big. Could it be that Coverity stops when it has found a certain amount of issues or one with a high-enough criticality-value? Or it randomly takes different branches for the evaluation in each run?

I'll try to solve them and push further commits to this PR. Once they are fixed we'll have to see how we squash the commits into two ones (one for the current topic of this PR and another one for the Coverity fixes).

mrbaseman:
Oh, this scan was not on the current branch. The fix for CID349832 [is already in my cert-only branch](https://github.com/mrbaseman/openfortivpn/blob/4a2848660ce1758385887048ec030f31d4a8c90b/src/userinput.c#L46). I have rebased that branch and force-pushed it to my account. Maybe you (or Coverity) have to (force-)pull again?

DimitriPapadopoulos:
Coverity scans have to be started manually - automatic runs are triggerred by commits to the adrienverge/coverity_scan branch as suggested by the Coverity documentation.

DimitriPapadopoulos:
Much better now: 0 defects!

DimitriPapadopoulos:
I suggest you merge.

mrbaseman:
@DimitriPapadopoulos  how is the `adrienverge/coverity_scan` branch updated? Can I push directly to that branch? Or can I make a pull request from master into that branch?

DimitriPapadopoulos:
I think you can merge `adrienverge/master` directly into `adrienverge/coverity_scan` - only `adrienverge/master` is protected. As far as I can recall this is what I had been applying:
[What is the best (and safest) way to merge a Git branch into master?](https://stackoverflow.com/questions/5601931/what-is-the-best-and-safest-way-to-merge-a-git-branch-into-master)

mrbaseman:

Issue #496, opened 2019-11-29
warning: comparison between pointer and zero character constant


DimitriPapadopoulos:
On Ubuntu 18.04 with _gcc version 7.4.0_:
```
src/http.c: In function ‘auth_log_in’:
src/http.c:581:32: warning: comparison between pointer and zero character constant [-Wpointer-compare]
   if (tunnel->config->password == '\0') {
                                ^~
src/http.c:581:7: note: did you mean to dereference the pointer?
   if (tunnel->config->password == '\0') {
       ^
```

mrbaseman:
ah that should probably read `if (tunnel->config->password[0] == '\0') {` which could be an explanation why this patch still didn't solve #375 

another issue in this context which has happened to a colleague of mine: If you have left a line
```
password =
```
in the config file, prior to 1.11.0 _openfortivpn_ has asked for a password interactively, now it connects and (the intention was not to send anything at all for a password, but instead...) it actually sends an empty string as password due to the bug that gcc 7.4 obviously notices whereas coverity and my gcc didn't stumble over this

mrbaseman:
created pull request #497 to fix this issue 

For the problem described [in my post above](https://github.com/adrienverge/openfortivpn/issues/496#issuecomment-559876218) I don't have a solution, because it was an intended change that _openfortivpn_ does not always ask for a password. I think we can leave it like that, but I wanted to have this documented so that users who run into that problem can find it (or at least we are aware of it and can refer to this)

I have noticed that the username is another required input, but at least in the context of tokens it can be empty (at least for some configurations) and maybe also in the usage scenario described in  #375 - and maybe the login mechanism is the same as that one used with the tokens from which we have at least a first report of success in #111

DimitriPapadopoulos:
Would it make sense to release a 0.11.1 for this fix?

mrbaseman:
let's wait if we get some positive feedback in #375

mrbaseman:
the fix for this in #497 has been merged into master, so I close this issue now

epsilon77:
> Would it make sense to release a 0.11.1 for this fix?

+1

mrbaseman:
this went into [release 1.12.0](https://github.com/adrienverge/openfortivpn/releases/tag/v1.12.0)

PR #497, opened 2019-12-02
fix for #496

mrbaseman:
...and try to solve login problems with certificate only (#375)


mrbaseman:

PR #498, opened 2019-12-04
Remove obsolete AC_TRY_RUN

DimitriPapadopoulos:
AC_TRY_RUN has been replaced by AC_RUN_IFELSE.

Additionally we do not need to run the snippet, just compile it.
Therefore use AC_COMPILE_IFELSE instead of AC_RUN_IFELSE.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -226,20 +226,19 @@ AC_ARG_WITH([rt_dst],
 	])
 )
 AS_IF([test "x$with_rt_dst" = "x" ], [
-AC_MSG_CHECKING(whether rtentry is available and has rt_dst )
-AC_TRY_RUN([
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Now that we just compile instead of running the code snippet, can't we get rid of `--with-rt_dst`?

Compiling snippets should work even when cross-compiling, there should be no need to force anything instead of testing automatically.

mrbaseman:
I would keep the `--with-rt_dst`. It can be useful to override the autoconf test if it fails for some reason. In a sane environment, even when cross-compiling, it should not be necessary, but if the compilation inside of autoconf doesn't find the right kernel headers for some reason, this is a handy way to influence the precompiler definitions.

DimitriPapadopoulos:
OK then, let's keep `--with-rt_dst` for now.

Issue #486, opened 2019-10-28
allow to use resolvconf instead of resolv.conf


luxferresum:
`--set-dns` should call `resolvconf` the same way the [`update-resolv-conf`](https://github.com/alfredopalhares/openvpn-update-resolv-conf) script does it for openvpn instead of directly modifying `resolv.conf`

DimitriPapadopoulos:
Isn't this a duplicate of #101? If not please explain the difference.

In any case we'll have to review issue #101 again and understand why it had been closed. I may be wrong but I seem to recall that's because this issue was beyond the reach of _openfortivpn_ - _pppd_ would be the culprit. Perhaps pull request #483 can help by preventing _pppd_ messing with DNS settings by default, which was not possible until now. Once this pull request is applied, we can make sure _openfortivpn_ calls _resolvconf_.

Also which Linux distribution are you using? For example on Ubuntu 16.04 it looks like _pppd_ is set to use `resolvconf`:
```
$ cat /etc/ppp/ip-up.d/000resolvconf 
#!/bin/sh
#
# PPP up hook script for resolvconf
#
# Reconfigures resolver to take into account
# the appearance of the ppp interface.
#
# This file is part of the resolvconf package.
#
[...]
echo -n "$R" | /sbin/resolvconf -a "${PPP_IFACE}.pppd"

$ 
```
How is _pppd_ set up on you Linux distribution?

mrbaseman:
I believe [this comment](https://github.com/adrienverge/openfortivpn/issues/101#issuecomment-395966476) by you Dimitri was the point at which (at least in my understanding) a fix was on the way. Unfortunately, we had to revert it, but with #483 I believe I have found what went wrong.

DimitriPapadopoulos:
Ah right. So I understand we need to replace all the code directly modifying `/etc/resolv.conf` when starting _openfortivpn_ by feeding:
```
namserver ns1
namserver ns2
search dns-suffix
```
to:
```
/sbin/resolvconf -a "<ppp_interface>.openfortivpn"
```

Then when shutting down:
```
/sbin/resolvconf -d "<ppp_interface>.openfortivpn"
```

Should we replace the code modifying directly `/etc/resolv.conf`, or perhaps attempt to use `resolvconf` and if not available fall back to modifying `/etc/resolv.conf`? How common is `resolvconf` on reasonably recent Linux distributions? What about BSD and macOS?

mrbaseman:
[...]

PR #506, opened 2019-12-10
use resolvconf if available to update resolv.conf

mrbaseman:
this implements the feature request from #486 and #101

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -9,7 +9,8 @@ openfortivpn_SOURCES = src/config.c src/config.h src/hdlc.c src/hdlc.h \
 openfortivpn_CFLAGS = -Wall -pedantic -std=gnu99
 openfortivpn_CPPFLAGS = -DSYSCONFDIR=\"$(sysconfdir)\" \
 			-DPPP_PATH=\"@PPP_PATH@\" \
-			-DNETSTAT_PATH=\"@NETSTAT_PATH@\"
+			-DNETSTAT_PATH=\"@NETSTAT_PATH@\" \
+			-DRESOLVCONF_PATH=\"@RESOLVCONF_PATH@\"
 
 openfortivpn_CPPFLAGS += $(OPENSSL_CFLAGS) $(LIBSYSTEMD_CFLAGS)
 openfortivpn_LDADD = $(OPENSSL_LIBS) $(LIBSYSTEMD_LIBS)
--- a/configure.ac
+++ b/configure.ac
@@ -246,6 +246,7 @@ cast_addr(&(&route)->rt_dst)->sin_family = AF_INET;
 
 NETSTAT_PATH=""
 PPP_PATH=""
+RESOLVCONF_PATH=""
 
 # prepare possibility to override default locations
 AC_ARG_WITH([netstat],
@@ -272,6 +273,12 @@ AC_ARG_WITH([ppp],
 		with_ppp="yes"
 		with_pppd="no"
 	])
+)
+# resolvconf if present
+AC_ARG_WITH([resolvconf],
+        AS_HELP_STRING([--with-resolvconf],
+                       [Set the path to the resolvconf executable]),
+        RESOLVCONF_PATH="$withval"
 )
 
 # override for /proc/net/route detection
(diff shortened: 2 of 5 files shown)


mrbaseman:
perhaps we first merge my other pull request and I'll rebase this one manually afterwards.
The nice thing of `resolvconf`is that it automatically merges the domain search path with an existing one (we have discussed this a bit when implementing the dns-suffix feature, but decided to start with the most simple approach).

mrbaseman:
Just a side note: My resolvconf thinks that it should not add nameservers - but that might be due to the fact that I have a local dns resolver running on my client machine. It also doesn't take the one it receives from dhcp (which is by intention, so it might be a configuration issue of my resolvconf.d)

mrbaseman:
I have rebased this PR on the current master branch.  I have found the reason why my resolvconf didn't update the nameservers: Like I have presumed, the reason was the following section in my `/etc/resolvconf.conf`:
```
# If you run a local name server, you should uncomment the below line and
# configure your subscribers configuration files below.
name_servers=127.0.0.1
```
If I uncomment the line with `name_servers` it correctly updates the nameservers in the resolv.conf file.

DimitriPapadopoulos:
OK, so I guess _resolvconf_ does the right thing here.

Should we always mimic _FortiClient_ , even in this case where it might not do the right thing? What does _FortiClient_ actually do in this case? I do understand that relying on _resolvconf_ might change the behaviour of _openfortivpn_ compared to _FortiClient_ and that might be a good thing - I'm just trying to characterize this change.

mrbaseman:
_FortiClient_ in this aspect does what _openfortivpn_ is doing currently (it just prepends search and nameservers in resolv.conf). With this change, _openfortivpn_ does use resolvconf if it is installed. If resolvconf is not available (or if its path is set to something nonexisting at configure time, i.e. `configure --with-resolvconf=/sbin/resolvconf_disabled`), _openfortivpn_ also will behave (nearly) the same way as before. It would check if an executable `/sbin/resolvconf_disabled` is there and when it notices that it's not it should fall back to the previous behavior.
I think if someone installs resolvconf, he does it for some reason, namely because he runs into trouble with the different programs (dhcp, various vpn solutions) that compete about changing resolv.conf. In that sense, I think doing something different than Forticlient it is a good move forward to provide a better user experience.

mrbaseman:
Maybe we should update the man page and perhaps README.md and describe the change there.
Not sure about the `--help` output - perhaps just add a hint and refer to the man page?

DimitriPapadopoulos:
Perhaps a very short hint in the `--help` summary. No need to refer to the man page or the README any more than we do now - that should be obvious.

mrbaseman:
yes, we are anyway close to the limit of 4095 characters. I have reworked the `--help` output a bit and the man page, too. I have mentioned resolvconf in the `--help` output. I think that's more important than mentioning that the last option takes priority when `--set-dns` is used multiple times (that's something I would consider normal unless an error is reported in this case)

mrbaseman:
At your edit of the title I've noticed I had a typo in the commit message - I've used this occasion to squash my later commits into one larger change, in the turn of the edit of the commit message

DimitriPapadopoulos:
Looks pretty good to me. Perhaps it should be merged into _master_ for wider testing?

mrbaseman:
Ok. Thanks for reviewing.

DimitriPapadopoulos:
Started testing it. Noticed we have _autoconf_ options for _pppd_/_ppp_ and _netstat_ but not _resolvconf_:
```
$ ./configure --help
[...]
Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
  --with-rt_dst           disable rtentry with rt_dst testing (for linux
                          target when cross compile)
  --with-netstat          Set the path to the netstat executable on MacOSX or
                          FreeBSD
  --with-pppd             Set the path to the pppd daemon executable
  --with-ppp              Set the path to the ppp userspace client on FreeBSD
[...]
```
Shouldn't we have such an option?

DimitriPapadopoulos:
Ah, but there is one! I'm probably not testing the proper changeset!

DimitriPapadopoulos:
Works well on Ubuntu 18.04.
Found missing newlines in a couple sol _log_debug_ calls while testing and fixed them in #512 - yes these newlines are counter-intuitive.

mrbaseman:
this went into [release 1.12.0](https://github.com/adrienverge/openfortivpn/releases/tag/v1.12.0), but we have a new issue #555 due to this change on Fedora, see also #556

Issue #507, opened 2019-12-11
blank password in default config


andreworg:
Hi,
I have 1.6.0 in my Ubuntu system, coming from the package manager.
It features a default config file `/etc/openfortivpn/config`, with null values.
```
# config file for openfortivpn, see man openfortivpn(1)
host =
port =
username =
password =
```

I just compiled 1.11.0 from git and tried running it before installing.
I found out that the blank password in the default configuration file will cause openfortivpn not to ask for a password at the CLI and my connections to fail just as if I provided a wrong password.

I don't know if this is by design. With the same default config, 1.6.0 prompts for the password.

Having a default config with blank values is probably not the best of ideas - I see in the repo there is a file named /etc/openfortivpn/config.**template**, probably that's what gets installed in the current release - but I think that the current behaviour is at least somehow misleading. At the very least, the verbose debug output should hint at the fact that a password is being picked up from the config file. I don't think it does now.
Better still, the log could show more clearly what is being taken from the config file and what is being picked up/overridden from the CLI parameters.

mrbaseman:
This is an intended change (although it is admittedly not a very lucky one) in order to support login with a smartcard or user certificate only.
Just comment out the password line in your config file (or remove it) - and openfortivpn 1.11.0 will ask you again for the password.

mrbaseman:
we should comment out the line in the template to avoid this problem in the future. The template is copied to the default config location if that file doesn't exist during installation.

mrbaseman:
a fix for this is landed on the master branch now. I would leave this ticket open until the next release, just to have it visible for other users who upgrade to the 1.11.0 release and run into the same problem

andreworg:
I still contend that the debug output could be clearer about where the parameters are taken from. I'll propose a PR if I find the time.

mrbaseman:
Additional information in the debug output usually helps understanding problems. In this context however, I don't really see the need anymore. In Version 1.8.0 Issue #258 was solved via #346. 

[...]

PR #508, opened 2019-12-11
comment out line for password in config.template (fix for #507)

mrbaseman:


DimitriPapadopoulos:
Perhaps we should completely remove this file, or install it somewhere it doesn't get read from - for example `/usr/share/openfortivpn/config.template`. Besides I believe many setups require a personal configuration file, not a system-wide configuration file.

DimitriPapadopoulos:
I'm also OK with this change, at least in the short term.

My only concern is that an example file should contain commented out items.

mrbaseman:
you mean something like this?

```
# config file for openfortivpn, see man openfortivpn(1)
# host = vpn.example.org
# port = 443
# username = vpnuser
# password = VPNpassw0rd
```

DimitriPapadopoulos:
I think we may install a version with lines commented out into `/etc/openfortivpn/config` (_sysconfdir_):
```
# config file for openfortivpn, see man openfortivpn(1)
# host = vpn.example.org
# port = 443
# username = vpnuser
# password = VPNpassw0rd
```
or a pristine version  into `/usr/share/openfortivpn/config.template` (_datadir_):
```
# config file for openfortivpn, see man openfortivpn(1)
host = vpn.example.org
port = 443
username = vpnuser
password = VPNpassw0rd
```
Or both. How would you feel about it?

mrbaseman:
oh, you have edited [your comment](https://github.com/adrienverge/openfortivpn/pull/508#issuecomment-564926485) and added a **not** so, [both files](https://github.com/adrienverge/openfortivpn/pull/508#issuecomment-565073326), commented and uncommented would be good, maybe [like this](https://github.com/adrienverge/openfortivpn/pull/508/commits/0dac3e91ce86363c84eaaf6f253481128550dcd5) we use `sed` anyway

DimitriPapadopoulos:
Great!

Just nitpicking here: perhaps the source version should **not** be commented because this is easier to read and maintain. If so, _sed_ should be used to add `#` in _sysconfdir_ instead of removing `#` in _datadir_.

mrbaseman:
Ok, I can change this. One more thought as we talk about the install hook:
Uninstall doesn't remove everything. If I recall it right, the config file is not removed and maybe the example files. Should we add an uninstall hook, too, that takes care of these? And if so, how should it handle files that have been edited? I.e. remove the config file only if it contains no modifications?

DimitriPapadopoulos:
I would uninstall the template in _datadir_ and leave the system-wide config file in _sysconfdir_. Not a big deal and probably standard practice.

mrbaseman:
oh, that's what actually happens already. I have made the source version without comments and updated the Makefile to insert comments for the default config file.

mrbaseman:
this fix went into [release 1.12.0](https://github.com/adrienverge/openfortivpn/releases/tag/v1.12.0)

PR #509, opened 2019-12-12
Use snake_case naming style for Python files

DimitriPapadopoulos:


mrbaseman:
@DimitriPapadopoulos by the way, I was thinking of adding [checkpatch](https://github.com/torvalds/linux/blob/master/scripts/checkpatch.pl) to the test-suite, but I'm not sure if travis-ci has a perl interpreter in its testing container. If we first have to download and compile perl it's too much overhead i.m.o.

Anyhow, I have added a bash function to my environment, now:
```
function checkpatch(){ 
  diff=$(git diff)
  if [ "$diff" = "" -a "$1" = "" ] ; then
     echo "trying staged changes" >&2
     diff=$(git diff --staged)
  fi
  if [ "$diff" = "" -o "$1" != "" ] ; then
    commit=$1
    if [ "$commit" = "" ] ; then commit="HEAD" ; fi
    echo "result for commit $commit" >&2
    diff=$(git show $commit)
  else
    echo "result for uncommited changes" >&2
  fi
  echo "$diff" | $HOME/bin/checkpatch.pl --no-tree --terse --max-line-length=90 --ignore MISSING_SIGN_OFF,CODE_INDENT,LEADING_SPACE 2>/dev/null && echo "no problems found" >&2 
}
```
**If** I remember to call it on each pull request you won't have to correct so many coding style issues anymore :-)

DimitriPapadopoulos:
Right now we still have issues caught by _checkpatch_. I haven't removed all of them - some seem far-fetched. So we're not ready to fully automate this - I'd have to check if we can selectively ignore some classes of warnings.

DimitriPapadopoulos:
Oh, just noticed:
```
--max-line-length=90 --ignore MISSING_SIGN_OFF,CODE_INDENT,LEADING_SPAC
```

DimitriPapadopoulos:
Perl is pretty mainstream. I'd give it a try. DO you want me to look into it?

mrbaseman:
Yes, that would be good. If you want more classes to be ignored: checkpatch.pl offers an option to list them, but if you have a specific error, in my impression was that the easiest is to grep for the message in the perl source and find the identifier for it. 

I have added the space classes because checkpatch.pl and our lint test implement different indentation styles.

DimitriPapadopoulos:
Also see #262.

mrbaseman:
I have noticed that checkpatch complains about long strings, strings split over several lines, and line  continuations. So, for using it in an automatic test we probably have to exclude a few more warning classes.

DimitriPapadopoulos:
Yes, I have fixed what can be easily and reasonably fixed in our current sources. Remaining warning classes should be excluded. One that should perhaps not be excluded is "strings split over several lines" as I have fixed most of not all these warnings.

BTW, how do we implement this?
- As far as I can see `checkpatch.pl --no-tree` is best run on patches, hence it is [ideally run as pre-commit hook](http://blog.vmsplice.net/2011/03/how-to-automatically-run-checkpatchpl.html), however I don't how to implement that directly on the GitHub server.
- Instead we probably need to run `checkpatch.pl --no-tree -f` on all C sources and header files from Travis.

Ah, just found how to run `checkpatch.pl --no-tree` on new patches only, instead of checking the whole existing code base - which could be handy to throw additional warning classes to new patches only:
https://github.com/linux-rdma/opensm/blob/master/scripts/travis-checkpatch

DimitriPapadopoulos:
@mrbaseman Not sure about this error class:
```
src/log.c:30: ERROR: do not initialise statics to 0
src/log.c:34: ERROR: do not initialise statics to 0
```
Should we enforce it?

mrbaseman:
well, we could move the initialization to `init_logging()` to comply with this coding style rule

DimitriPapadopoulos:
In ISO C static variables are automatically initialized to 0. Explicit initialization is useless. That's the reason for this rule.

See for example:
[Why are global and static variables initialized to their default values?](https://stackoverflow.com/questions/2091499/why-are-global-and-static-variables-initialized-to-their-default-values)

mrbaseman:
Hmm... does Codacy go ok with this, too? Or does it recommend to initialize all variables unconditionally?

DimitriPapadopoulos:
I don't know, Codacy has been out of order for quite a few days now.

mrbaseman:
I've seen that you have opened #524 for this. So, let's move the discussion over there.

PR #513, opened 2019-12-13
fix segfault when config file is empty and connecting without password

mrbaseman:


mrbaseman:

PR #518, opened 2019-12-18
add a hint in debug output when password is read from config file

mrbaseman:


mrbaseman:
this is for #507

DimitriPapadopoulos:
Perhaps add a distinction between empty password and non-empty password?

mrbaseman:

PR #520, opened 2019-12-19
Linux kernel coding style

DimitriPapadopoulos:
* However, never break user-visible strings such as printk
  messages, because that breaks the ability to grep for them.


mrbaseman:
ok, so this has higher priority than the enforced line length limit of 90 characters.

Issue #407, opened 2018-12-04
-vv shows password in clear text


UweSauter:
When running openfortivpn with "-v -v" it will show the password in clear text inside the POST request.

With this config file `test.vpn`:
```
host = XXX.XXX.XXX.XXX
port = 443
username = testuser
password = testpw
set-routes = 0
set-dns = 0
```

I get (important part)
`username=testuser&credential=testpw&realm=&ajax=1&redir=%2Fremote%2Findex&just_logged_in=1`

complete output:
```
DEBUG:  openfortivpn 1.8.0
DEBUG:  Loaded config file "test.vpn".
DEBUG:  Config host = "XXX.XXX.XXX.XXX"
DEBUG:  Config realm = ""
DEBUG:  Config port = "443"
DEBUG:  Config username = "testuser"
DEBUG:  Config password = "********"
DEBUG:  Resolving gateway host ip
DEBUG:  Establishing ssl connection
DEBUG:  server_addr: XXX.XXX.XXX.XXX
DEBUG:  server_port: 443
DEBUG:  gateway_addr: XXX.XXX.XXX.XXX
DEBUG:  gateway_port: 443
DEBUG:  Gateway certificate validation succeeded.
INFO:   Connected to gateway.
DEBUG:  http_send : 
POST /remote/logincheck HTTP/1.1
Host: XXX.XXX.XXX.XXX:443
User-Agent: Mozilla/5.0 SV1
Accept: text/plain
Accept-Encoding: identify
Content-Type: application/x-www-form-urlencoded
Cookie: 
Content-Length: 90

username=testuser&credential=testpw&realm=&ajax=1&redir=%2Fremote%2Findex&just_logged_in=1
DEBUG:  http_receive : 
HTTP/1.1 405 Method Not Allowed
Date: Tue, 04 Dec 2018 22:45:43 GMT
Server: xxxxxxxx-xxxxx
Set-Cookie:  SVPNCOOKIE=; path=/; expires=Sun, 11 Mar 1984 12:00:00 GMT; secure; httponly;
Set-Cookie: SVPNNETWORKCOOKIE=; path=/remote/network; expires=Sun, 11 Mar 1984 12:00:00 GMT; secure; httponly
Transfer-Encoding: chunked
Content-Type: text/html
X-Frame-Options: SAMEORIGIN
Content-Security-Policy: frame-ancestors 'self'
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
[...]

PR #523, opened 2019-12-20
add a clear warning about sensitive information in the debug output

mrbaseman:
We had another issue in which the password was disclosed. In openfortivpn we suppress it in the -vv output since #443 as a reaction on #407 and we have introduced -vvv as a new verbosity level, but now users submit -vvv debug output without a closer look, so a clear warning at the beginning of the output might help to avoid disclosure by accident.


mrbaseman:

PR #524, opened 2019-12-20
Run checkpatch.pl to enforce Linux kernel coding style

DimitriPapadopoulos:
• For now we disable many warnings or even errors to avoid brutal changes to the code base.
• For now we check the whole C code base instead of checking only new patches.

I'm still working on it. will improve this iteratively directly in our production environment.


mrbaseman:
currently, we have a lot of these:
 ```
ERROR: code indent should use tabs where possible
WARNING: please, no spaces at the start of a line
WARNING: please, no space before tabs
WARNING: Missing a blank line after declarations
```
in the code. If you like I could try to fix those between the years
Maybe we start with warnings disabled globally (if such an option exists, otherwise we would have to use a long list of warning classes).

Let's have at the errors that remain and see if they can be fixed easily.

**Comment by DimitriPapadopoulos** [comment]
We currently disable these errors/warnings:
```
SPACE_BEFORE_TAB
LEADING_SPACE
LINE_SPACING
SPDX_LICENSE_TAG
CODE_INDENT
NAKED_SSCANF
VOLATILE
NEW_TYPEDEFS
```
Remaining errors/warnings are:
```
ERROR: do not initialise statics to 0
WARNING: line over 90 characters
```

**Comment by mrbaseman** [comment]
we could add LONG_LINE to the ignore list (currently line_length.py checks this), 
and LONG_LINE_STRING and LONG_LINE_COMMENT are ignored automatically, then.

With the statics initialized to 0, even if it does not make a difference, I[ still tend to](https://github.com/adrienverge/openfortivpn/pull/509#issuecomment-567536251) an explicit initialization in [init_logging()](https://github.com/adrienverge/openfortivpn/blob/0157f735bc306c70322c4641840ca5d5c2975fc2/src/log.c#L52), just to make it clearer what happens, to readers of the code and to code analysis tools as well 

I'd clean up some white space issues, and perhaps we can remove these from the ignore list then. If the barrier is too high for pull requests in the future, we still have the possibility to push commits that fix violations of those coding style rules

**Comment by DimitriPapadopoulos** [comment]
I had to add both `LONG_LINE` and `LONG_LINE_STRING` to shut up all errors and warnings.

I tend to agree with you about statics, however another option might simply be:
```
static int is_a_tty; // static variables of arithmetic type are initialized to zero in C99
```
instead of:
```
static int is_a_tty = 0;
```

**Comment by DimitriPapadopoulos** [comment]
My main concern is, should we run _checkpatch_ on new patches only or on the whole code base as I do now?

The nice thing with running _checkpatch_ on new patches only is that we do not need to change all the existing code base. Of course we would disable errors such as `SPACE_BEFORE_TAB` or `LEADING_SPACE` to allow for consistency in indentation of the new existing code base and new patches.

A possible side effect of running _checkpatch_ on new patches is that it may catch errors directly inherited from the existing code base and disturb casual contributors.

**Comment by DimitriPapadopoulos** [comment]
What about `NAKED_SSCANF` and `NEW_TYPEDEFS`?

**Comment by DimitriPapadopoulos** [comment]
Finally shall I add [SPDX License List identifiers](https://spdx.org/licenses/) to the headers? That would get rid of `SPDX_LICENSE_TAG`.

**Comment by mrbaseman** [comment]
If possible I would run checkpatch on the new submissions only, but anyhow try to fix most of the findings in the existing code. This makes it more acceptable for contributors. If the existing code does not fulfill the requirements, the high barrier for new contributions seems difficult to understand.

Somehow , the NAKED_SSCANF and NEW_TYPEDEDFS have disapeared in my branch. The typedefs are ok imo, but maybe we should look at the sscanf issue

**Comment by DimitriPapadopoulos** [comment]
Disabled errors/warnings down to:
```
LEADING_SPACE
SPDX_LICENSE_TAG
CODE_INDENT
NAKED_SSCANF
VOLATILE
NEW_TYPEDEFS
LONG_LINE
LONG_LINE_STRING
```

**Comment by DimitriPapadopoulos** [comment]
For now I run `checkpatch.pl` on all files, not new submissions only.

File: tests/lint/checkpatch.sh [hunk 1]
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Copyright (C) 2015 Adrien Vergé
```
adrienverge:
```suggestion
# Copyright (C) 2020 Dimitri Papadopoulos
```
:wink:

DimitriPapadopoulos:
We already run `checkpatch.pl` on all files, not only new patches. Point taken though, I won't change to checking new patches only. Martin has already cleaned up the existing code base so it's not an issue.

I'll add `checkpatch.pl` to the repo instead of downloading it, excellent idea. I'd rather avoid tweaking it to be able to upgrade to new versions easily. I'll see what I can do about it.

My intention was indeed to keep some warnings disabled for now. I don't want to modify teh source code in this PR.

DimitriPapadopoulos:
Spelling uses `/usr/share/codespell/dictionary.txt`,`spelling.txt` is just a small ancilary file which can easily be included.

DimitriPapadopoulos:
> * storing `checkpatch.pl` in the repo, under `tests/lint/`,

I have stored `checkpatch.pl` and its ancillary files under `./tests/ci/checkpatch/` as I understand we put the infrastructure for continuous integration under `./tests/ci/` and our tests under `.:tests/lint/`.

DimitriPapadopoulos:
On my workstation it takes a few seconds indeed but nothing unbearable. Probably more annoying in VMs though.
```
$ time ./tests/lint/checkpatch.sh $(git ls-files '*.[ch]' | grep -v openssl_hostname_validation)

real	0m6,039s
user	0m5,821s
sys	0m0,164s
$ 
```

mrbaseman:
On old hardware (AMD Athlon(TM) XP 2400+, 2 GHz):
```
real    0m51.102s
user    0m42.264s
sys     0m1.486s
```
on a netbook (AMD C-60 1 Ghz):
```
real    1m26.020s
user    1m19.970s
sys     0m2.702s
```
a VM however, if the hardware below is strong enough (Intel(R) Core(TM) i5-4250U CPU, 1.30GHz):
```
real    0m10.782s
user    0m10.140s
sys     0m0.328s
```
that's more or less what I see on my laptop (which is ok imho).

Issue #526, opened 2019-12-20
Possibility to set a cutom dns server via command flag


tuxBurner:
Sometimes you want to set a custom dns sever, because the ones provided from the fortigate are not sufficiant.

The flag could be like this: --use-dnsServer=10.0.0.3
This will lead to, that the dns server is set at the top of the resolv.conf before the other dns server provided by the fortigate server.

DimitriPapadopoulos:
Should we really make it easy to bypass the corporate policy? My feeling is that this is very specific and best left to a an _ip-up_ script. But that's only me.

What is the use case? Why would the DNS server pushed by the FortiGate device not be sufficient?

mrbaseman:
I feel the same. There are a couple of ways to influence how /etc/resolv.conf is updated. We support [resolvconf](http://manpages.ubuntu.com/manpages/bionic/man8/resolvconf.8.html) now, there is the possibility to use a [pppd ip-up script](https://www.tldp.org/HOWTO/PPP-HOWTO/x1455.html), when using networkmanager it [can be configured there](https://www.tldp.org/HOWTO/PPP-HOWTO/x1455.html) I think (unless you have several different vpn clients that use the same interface name), or just call openfortivpn with `--no-dns` and call it from a wrapper script that also takes care of the resolv.conf file.

tuxBurner:
Well sometimes you have a vpn where you don't get a dns server from the other side provided.
Mostly i see people polluting there /etc/hosts with entries to get a "name resolving". 

This is not a good way IMHO because often you have virtual hosts which than have to be added individually. 

So i created a dnsmasq server which the people can use but this ends at the moment in the workflow:

open vpn
edit  /etc/resolv.conf

lfuelling:
IMO this is a quality of life improvement to allow people that are less used to configuring Linux networking by themselves. 
Like with everything you should of course ask your system administrator if this violates **a** company policy (I have yet to find **the** policy you refer to, doesn't every company has their own?)

Regarding the use-case, it actually is **a** company policy that there is no DNS provided in the VPN and usually we use the hosts file to manage name resolving in the VPN. 
The addition of the DNS came out of necessity due to new/changed IPs which caused a lot of overhead due to everyone having to keep their hosts file up to date. 

[...]

PR #527, opened 2019-12-20
#526 add flag to set custom dns server

tuxBurner:


DimitriPapadopoulos:
Why add such an option to _openfortivpn_? Setting a custom DNS server seems orthogonal to setting up a VPN tunnel. I would rather add hooks to run a script when the tunnel is up. There's already the `systemd` notifications. Then we should add such hooks and move the routing/DNS stuff to a shell script. See many other issues on this last subject.

PR #529, opened 2019-12-20
Add SPDX License identifiers

DimitriPapadopoulos:
See https://spdx.org/licenses.

Not sure about GPL with OpenSSL exception, so documenting it as GPL-3.0-or-later like the rest.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
 /*
  *  Copyright (C) 2015 Adrien Vergé
  *
--- a/src/config.h
+++ b/src/config.h
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: GPL-3.0-or-later */
 /*
  *  Copyright (C) 2015 Adrien Vergé
  *
--- a/src/hdlc.c
+++ b/src/hdlc.c
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
 /*
  *  Copyright (C) 2015 Adrien Vergé
  *
--- a/src/hdlc.h
+++ b/src/hdlc.h
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: GPL-3.0-or-later */
 /*
  *  Copyright (C) 2015 Adrien Vergé
  *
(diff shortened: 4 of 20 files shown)


mrbaseman:
I'm not an expert in licensing topics. So, here just my thoughts without warranty :)
* we already ship a license file for GPL v3, and as specified in the source, it can be changed to newer versions (I believe that then anyone can redistribute the code with a newer version of GPL without asking the authors for permission to do so).
* we also ship the Openssl license file. I think we don't have to do so, because we don't **redistribute** the library, neither it's source nor binary packages. We only link against it and use it's API. Including the license file in my understanding is a gimmick for packagers who might provide statically linked builds in some rare cases. 
Our code is released under GPL, and the licesnse file is included and referenced in the source code files. So, I don't see a big advantage in adding the SPDX License identifiers, but well, another line of comment doesn't hurt either. I just don't see the advantage. But as I said, I'm not an expert in these topics.
@adrienverge maybe you can comment on this, too?

adrienverge:
Hey Martin & Dimitri,

I wasn't aware of these new licensing markers by the Linux Foundation.

My understanding is that they can be useful for large projects that have multiple licenses and/or many dependencies with various licenses. So I don't see a direct advantage of having them for openfortivpn, but they don't hurt and it's always nice to promote cool initiatives like this one!

About OpenSSL, linking against the library (without shipping it in the default configuration) is enough to have to distribute the license file, isn't it? I'm not an expert neither. Lubomir (@lkundrak), you know more than me on this topic, what do you think?

Anyway, I don't have a strong opinion on that: if this PR does not create OpenSSL-related problems, I'm fine with either merging or not!

PS: By the way I've seen a lot of activity and style fixing recently, thanks for all that good work guys!

DimitriPapadopoulos:
I'm suggesting these SPDX identifiers because _checkpatch.pl_ requires them.

DimitriPapadopoulos:
On the other hand I can add `SPDX_LICENSE_TAG` to the ignore list to shut up _checkpatch.pl_.

This adds noise to the source files that might make sense in a large project with multiple licenses, but without clear benefit for _openfortivpn_.

Which do you prefer? This patch or shutting up _checkpatch.pl_ with `SPDX_LICENSE_TAG`?

adrienverge:
I'm for ignoring SPDX tags using `SPDX_LICENSE_TAG` :+1:

DimitriPapadopoulos:
OK, closing this pull request then.

PR #530, opened 2019-12-20
Coding style

mrbaseman:


mrbaseman:
some coding style issues which can easily be fixed as far as I can see. 
Indentation seems to be a problem: astyle suggests something else than checkpatch
I should have rebased my branch before starting this... anyhow, maybe we just cherry-pick individual commits from here

DimitriPapadopoulos:
To remain in the spirit of the Linux kernel coding style instead of working around it, I'd rather add a comment after static variable declarations instead of explicitly initializing in _init_logging_().
```
static int is_a_tty; // static variables of arithmetic type are initialized to zero in C99
```

mrbaseman:
done, and rebased on current master. Unfortunately we have a line with 91 characters, now.

Also, there are a lot of ```
WARNING: struct  should normally be const
```
Unfortunately, they can't be fixed automatically, and I'm not sure if they are appropriate in our code. If we declare all structs as const, can we still change the values and const refers to the offsets of the properties inside of the struct.

DimitriPapadopoulos:
Actually if the struct is const, the value of struct members cannot be changed. If the struct members are pointers we can change the contents of the memory pointed to but not the pointer itself.

Therefore in most cases we cannot use const struct in _openfortivpn_. Perhaps we don't follow some Linux idiom where all struct must be created const and later copied, not modified. I don't know.

mrbaseman:
You are right, `const struct` does not match the data concept in _openfortivpn_. It may compile but woh't work at runtime. Anyhow, I have to rebase these commits and I'll double-check which ones make sense and for which ones we should exclude the corresponding rule in checkpatch.pl

mrbaseman:
Probably it's much easier to do the coding style work again on a fresh branch based on the current master

PR #535, opened 2020-01-09
always hide cleartext password in -vv output

mrbaseman:
when special characters were contained that need to be url_encoded
the anonymization mechanism did not mask the password in the debug
output of the http request


mrbaseman:

Issue #540, opened 2020-01-19
Memory leak


dovi5988:
Hi,

I have one host with multiple sessions out. It seems there is some sort of memory leak. dmesg is showing:
[5112108.572250] [ 2381]  1002  2381     5956       37      13        0             0 perl
[5112108.577103] Out of memory: Kill process 24130 (openfortivpn) score 295 or sacrifice child
[5112108.581677] Killed process 24130 (openfortivpn), UID 0, total-vm:1245576kB, anon-rss:884460kB, file-rss:0kB, shmem-rss:0kB

How would I go about troubleshooting this? Is there any logs that I can connect to post here to debug further?

DimitriPapadopoulos:
I don't know much about the OOM killer of the Linux kernel. As far as I can understand your machine is our of memory and the kernel selects a process to kill in an attempt to recover. Which exact part of the log suggest a memory leak in _openfofrtivpn_? Perhaps `total-vm:1245576kB` which seems to imply that _openfortivpn_ is consuming > 1 GB of RAM? Also an `oom_score` of 295 is pretty bad as far as I can understand.

Which exact Linux distribution is this? Which exact version of the kernel?

DimitriPapadopoulos:
Also how much RAM do you have on your system? How many _openfortivpn_ processes do you run in parallel and how much memory do they consume?

DimitriPapadopoulos:
See for example :
* [How to diagnose causes of oom-killer killing processes - Server Fault](https://serverfault.com/questions/134669/how-to-diagnose-causes-of-oom-killer-killing-processes#136956)
* [Out of Memory (OOM) in Linux machines](https://tunnelix.com/out-of-memory-oom-in-linux-machines/)

dovi5988:
Sorry for not updating sooner. The box is a VPS running on Digital Ocean. I have two public IP's on the FortiNet device and I have a connectivity to each one. The VPS has 3GB of ram. Top shows each connection using 13 of ram:
25264 root      20   0  737748 383988   2828 S  0.0 13.2  10:14.85 openfortivpn                                                                                                                               
25301 root      20   0  728668 383196   2828 S  0.0 13.1  12:59.74 openfortivpn    

The box shows the actual total as 2846:
[root@noc2 ~]# free -m
              total        used        free      shared  buff/cache   available
Mem:           2846        1515         250         152        1080        1022
Swap:             0           0           0

[...]

PR #544, opened 2020-02-25
Check the return value of close()

DimitriPapadopoulos:
I don't think there are significant cases where _close()_ and friends fail out there. Yet check return value and emit a warning if it ever happens


DimitriPapadopoulos:
I don't think this will help with #540 but it's good practice to check return values anyway - even though chances are these _close()_ calls will never go wrong.

PR #547, opened 2020-02-27
free resolvconf_call after last usage

mrbaseman:
this  "Memory - illegal accesses  (USE_AFTER_FREE)" was found by coverity scan


mrbaseman:
Hmm...  Note that _file_ is either the `resolv.conf` file or a pipe to the `resolvconf` process.
The same values _ns1_, _ns2_, _dns_suffix_ have to be prepared and written. We could pull that apart, but probably we would end up with duplicated code for the rest of the function.

DimitriPapadopoulos:
The common part could be added to a function as well - I agree it shouldn't be duplicated.

PR #550, opened 2020-02-27
Explicitly initialize use_engine

DimitriPapadopoulos:
It actually worked because use_engine is intialized to 0 here or -1 elsewhere
and we later test:
	use_engine > 0
Yet the expected values seem to be -1 and 1, not 0. Perhaps we should
make use_engine a "boolean" with 0/1 values, but the expected values
clearly seem to be -1/1 for now.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -578,7 +578,7 @@ int auth_log_in(struct tunnel *tunnel)
 
 	tunnel->cookie[0] = '\0';
 
-	if (tunnel->config->use_engine
+	if (tunnel->config->use_engine > 0
 	    || (username[0] == '\0' && tunnel->config->password[0] == '\0')) {
 		snprintf(data, sizeof(data), "cert=&nup=1");
 		ret = http_request(tunnel, "GET", "/remote/login",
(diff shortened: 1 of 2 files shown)


mrbaseman:
We don't check use_engine > 0 [here](https://github.com/adrienverge/openfortivpn/blob/4c9057a92711df6341344df8899654ca9c66f213/src/http.c#L581), but properly initializing it to 0 would be better. The value of -1 [here](https://github.com/adrienverge/openfortivpn/blob/4c9057a92711df6341344df8899654ca9c66f213/src/config.c#L64) is for invalid_config, although that value is nowhere used yet, because it is not a mandatory value that we require being set either on the command line or in the config file.

DimitriPapadopoulos:
Ah right. Should we change to checking use_engine > 0 then? Also I understand we should initialize to the same value in both cases, either 0 or -1, shouldn't we?

DimitriPapadopoulos:
Ah sorry, got it for the initialization values.

DimitriPapadopoulos:
But what about checking use_engine > 0?

mrbaseman:
yes, that looks good now. Can you rebase it? Or maybe the "Update branch" button on github does the job, and then squash and merge.

PR #554, opened 2020-02-28
Fixed Makefile compatibility with BSD sed

Bo98:
As discussed in https://github.com/Homebrew/homebrew-core/pull/50728.


mrbaseman:
looks good to me. I should have tested on Mac or BSD. I can remember when I have done the change that was responsible for this incompatibility. My thought was "we are using sed anyway" but I didn't pay enough attention on how the arguments to `sed` are supposed to be used - and that there are differences between the platforms. I think I did a quick check if the `-i` option was available, but obviously the devil is more in the details. I'm sorry for that.

PR #558, opened 2020-03-05
Consistency in HDLC support

DimitriPapadopoulos:
Applied some minor changes while trying to understand and review this source file:
* Const-correctness
* Consistency with RFC 1662 (same vocabulary and use the exact same code for FCS calculation)
* Consistency within file (re-use existing variables such as the pre-calculated checksum)
* Reworked code comments

### DIFF
--- a/src/hdlc.c
+++ b/src/hdlc.c
@@ -17,13 +17,24 @@
 
 #include "hdlc.h"
 
+/*
+ * Encode and decode PPP packets from and into HDLC frames.
+ *
+ * RFC1622 describes the use of HDLC-like framing for PPP encapsulated packets:
+ *
+ * https://www.rfc-editor.org/info/rfc1662
+ */
+
 #define in_sending_accm(byte) \
 	((byte) < 0x20 || ((byte) & 0x7f) == 0x7d || ((byte) & 0x7f) == 0x7e)
 
 #define in_receiving_accm(byte) \
 	((byte) < 0x20)
 
-static uint16_t fcs_tab[] = {
+/*
+ * Lookup table used to calculate the FCS, as generated in RFC1662.
+ */
+static const uint16_t fcs_tab[] = {
 	0x0000, 0x1189, 0x2312, 0x329b, 0x4624, 0x57ad, 0x6536, 0x74bf,
 	0x8c48, 0x9dc1, 0xaf5a, 0xbed3, 0xca6c, 0xdbe5, 0xe97e, 0xf8f7,
 	0x1081, 0x0108, 0x3393, 0x221a, 0x56a5, 0x472c, 0x75b7, 0x643e,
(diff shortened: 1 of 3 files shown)


mrbaseman:
at first glance this looks good to me. I believe it is logically correct, although one might find surprises when changing bit operations, depending on the compiler, hardware, etc.

DimitriPapadopoulos:
Yes, I'm not 100% certain this is useful: why change code that works? But then when trying to understand the code I find it's easier to have code that has been copied/pasted directly from the RFC 1662 sample code.

mrbaseman:
You are right, the code is much better understandable when the sections are copied 1:1 from the RFC. We just should not forget to do extensive testing on this change.

DimitriPapadopoulos:
The code has been working so far for me - but that is not extensive testing :-)

DimitriPapadopoulos:
Also I am worried about this **static** variable:
```
static int need_flag_sequence = 1;
```

Shouldn't this variable be re-initalized at each loop after adding the `--persistent` loop? This may be true of other static variables elsewhere. Am I missing something here?

mrbaseman:
good catch! absolutely. 

Also the mutexes and semaphores in io.c are static variables, but I'm not really familiar with these mechanisms, so I can't tell if we have another problem there. 

The other occurrences are `static const` or static functions which  should be ok.

DimitriPapadopoulos:
I believe this `--persistent` option might not be such a good idea, not because of the static variables (for that we can add a module initialization/reset function for each of the files with static variables and call it before each loop) but also because of the potential leaks in programs exec'ed by _openfortivpn_ such as _pppd_ (we cannot do much about it).

mrbaseman:
I went back in the history here and found that I have added the `--persistent` option in #190 after it has been requested several times and successfully tested in #92 . We all were not really happy with this feature, but there was / there is a demand for it by several users. However, it's a separate topic, so I'll open a new issue for this.

DimitriPapadopoulos:
Do you want a separate pull request for `need_flag_sequence`?

mrbaseman:
No. I just have opened #560 so that we don't forget to double check other files in the context of this topic. If we come to the conclusion that `need_flag_sequence` is the only problematic area, it will be  fixed once this PR gets merged. But if we find out that we need some more reset functions elsewhere, #560 is our reminder for that.

DimitriPapadopoulos:
I've tested this pull request as much as possible on my end. OK for merging?

src/hdlc.c
```diff
@@ -59,53 +70,80 @@ static uint16_t fcs_tab[] = {
...
- * 16-bit frame check sequence from RFC1662.
+ * Calculates a new FCS from the current FCS and new data.
+ *
+ * @param[in] sum      Current FCS value.
+ * @param[in] seq      Array of new data.
+ * @param[in] length   Length of the array of new data.
+ * @return             new FCS value.
  */
-static uint16_t frame_checksum_16bit(uint16_t sum, uint8_t *seq, size_t length)
+static uint16_t frame_checksum_16bit(uint16_t sum, const uint8_t *seq, size_t length)
```
mrbaseman: is this `const` ok here? there is `*seq++` on the new line 83

src/hdlc.c
```diff
@@ -59,53 +70,80 @@ static uint16_t fcs_tab[] = {
...
 
 	return sum;
 }
 
+/*
+ * Precalculated FCS for Address and Control fields.
+ *
+ *     address_control_checksum = frame_checksum_16bit(0xffff, { 0xff, 0x03 }, 2);
+ */
+static const uint16_t address_control_checksum = 0x3de3;
```
mrbaseman: I didn't double check this, hope it's correct :)

src/hdlc.c
DimitriPapadopoulos: If it weren't const the compiler would complain. We don't assign to `*seq++` so that's OK.

src/hdlc.c
DimitriPapadopoulos: That value was already used in the source code:
https://github.com/adrienverge/openfortivpn/blob/b596a0c/src/hdlc.c#L244
I double-checked it by compiling a small test program.

src/hdlc.c
mrbaseman: oh, yes, you are right. I've found a good write-up about this topic: https://www.geeksforgeeks.org/difference-const-char-p-char-const-p-const-char-const-p/

src/hdlc.c
mrbaseman: I have missed that the value was already in the source code. I'm sorry.

Issue #555, opened 2020-03-01
openfortivpn 1.12 / DNS configuration problems


slartibart70:
Since using openfortivpn 1.12 i do see issues with the DNS resolution, the tunnel is up and running, but no name resolution is possible (see log).
My current workaround is to stay with version 1.11 which runs properly
(running on fedora31, latest updates)

Using ip addresses instead of names is ok, querying the (remote/tunneled) nameserver directly with 'dig @a.b.c.d hostname' is ok, but a 'cat /etc/resolv.conf' only gives me my local router address, there are no remote dns server entries at all. (resolv.conf is managed by networkmanager on my system)

see error line: 'Failed to set DNS configuration: Could not activate remote peer.'

Logs:
INFO:   Connected to gateway.
INFO:   Authenticated.
INFO:   Remote gateway has allocated a VPN.
Using interface ppp0
Connect: ppp0 <--> /dev/pts/3
INFO:   Got addresses: [10.147.x.z], ns [10.62.z.zz, 10.62.z.zz]
INFO:   negotiation complete
INFO:   negotiation complete
local  IP address 10.147.x.x
remote IP address 192.0.x.x
INFO:   Interface ppp0 is UP.
INFO:   Setting new routes...
WARN:   Route to gateway exists already.
INFO:   Adding VPN nameservers...
Failed to set DNS configuration: Could not activate remote peer.
INFO:   Tunnel is up and running.
^CINFO:   Cancelling threads...
INFO:   Setting ppp interface down.
INFO:   Restoring routes...
INFO:   Removing VPN nameservers...
Failed to revert interface configuration: Could not activate remote peer.
Hangup (SIGHUP)
Modem hangup
Connect time 12.2 minutes.
Sent 113414 bytes, received 1988486 bytes.
Connection terminated.
INFO:   pppd: The link was terminated by the modem hanging up.
INFO:   Terminated pppd.
INFO:   Closed connection to gateway.
INFO:   Logged out.

Any help with this?

DimitriPapadopoulos:
The only relevant change is:
* [+] use resolvconf if available to update resolv.conf file

This is interesting:
```
Failed to set DNS configuration: Could not activate remote peer.
```
as this error message is associated to `resolvconf`. See #514.

Yet `resolvconf` is not available on Fedora by default as far as I know. Let's see:
* What is the output of `rpm -qa | grep resolvconf`?
* What is the output of `which resolvconf` and `locate resolvconf`?

[...]

PR #565, opened 2020-03-13
add configure and command line option for resolvconf

mrbaseman:
see discussion in #555

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -414,7 +414,11 @@ AS_IF([test "x$enable_resolvconf" = "x" ], [
     AS_IF([$RESOLVCONF_PATH -l &>/dev/null], [
      enable_resolvconf="yes"
     ],[
-     enable_resolvconf="no"
+      AS_IF([ $RESOLVCONF_PATH &>/dev/null ; test $? -eq 99 ], [
+       enable_resolvconf="yes"
+      ],[
+       enable_resolvconf="no"
+      ])
     ])
   ],[
      enable_resolvconf="no"
--- a/configure.ac
+++ b/configure.ac
@@ -411,7 +411,7 @@ AC_ARG_ENABLE([resolvconf],
 # Checks for to determine how resolvconf works is it is installed
 # openresolv supports a -l option that lists the active configurations and returns 0
 # resolvctl supports -a -f which silently exits if no interface is suppplied
-# the resolvconf script in Ubuntu/Debian does not support listing 
+# the resolvconf script in Ubuntu/Debian does not support listing
 # but returns 99 if invoked without parameters
 #
 AS_IF([test "x$enable_resolvconf" = "x" ], [
--- a/configure.ac
+++ b/configure.ac
@@ -408,16 +408,25 @@ AC_ARG_ENABLE([resolvconf],
 			[Enable usage of resolvconf by default, \
 			use --disable-resolvconf for the opposite]))
 
-# Checks for existence of specific files if not cross-compiling.
+# Checks for to determine how resolvconf works is it is installed
+# openresolv supports a -l option that lists the active configurations and returns 0
+# resolvctl supports -a -f which silently exits if no interface is suppplied
+# the resolvconf script in Ubuntu/Debian does not support listing 
+# but returns 99 if invoked without parameters
+#
 AS_IF([test "x$enable_resolvconf" = "x" ], [
   AC_CHECK_FILE([$RESOLVCONF_PATH],[
     AS_IF([$RESOLVCONF_PATH -l &>/dev/null], [
      enable_resolvconf="yes"
     ],[
-      AS_IF([ $RESOLVCONF_PATH &>/dev/null ; test $? -eq 99 ], [
+      AS_IF([$RESOLVCONF_PATH -a -f &>/dev/null], [
        enable_resolvconf="yes"
       ],[
-       enable_resolvconf="no"
+	AS_IF([$RESOLVCONF_PATH &>/dev/null ; test $? -eq 99], [
+	 enable_resolvconf="yes"
+	],[
+	 enable_resolvconf="no"
+	])
       ])
     ])
   ],[
--- a/configure.ac
+++ b/configure.ac
@@ -274,12 +274,6 @@ AC_ARG_WITH([ppp],
 		with_pppd="no"
 	])
 )
-# resolvconf if present
-AC_ARG_WITH([resolvconf],
-        AS_HELP_STRING([--with-resolvconf],
-                       [Set the path to the resolvconf executable]),
-        RESOLVCONF_PATH="$withval"
-)
 
 # override for /proc/net/route detection
 AC_ARG_ENABLE([proc],
@@ -358,6 +352,14 @@ AS_IF([test "x$with_ppp" = "xyes" -a "x$with_pppd" = "xyes"], [
 
 AC_PATH_PROG(RESOLVCONF_PATH, [resolvconf], [/sbin/resolvconf], "$PATH:/sbin:/usr/sbin")
 
+# resolvconf if present
+AC_ARG_WITH([resolvconf],
+        AS_HELP_STRING([--with-resolvconf],
+                       [Set the path to the resolvconf executable]),
+        RESOLVCONF_PATH="$withval"
+)
+
+
 AS_IF([test "x$with_ppp" = "xyes"], [
 	AC_DEFINE(HAVE_USR_SBIN_PPP, 1)
 	AC_MSG_NOTICE([HAVE_USR_SBIN_PPP... 1])
@@ -391,6 +393,13 @@ AS_IF([test "x$NETSTAT_PATH" != "x" ], [
 AC_SUBST(RESOLVCONF_PATH)
 AS_IF([test "x$RESOLVCONF_PATH" != "x" ], [
     AC_MSG_NOTICE([RESOLVCONF_PATH...] $RESOLVCONF_PATH)
+])
+AS_IF([test "x$RESOLVCONF_PATH" = "xDISABLED"], [
+	AC_DEFINE(DISABLE_RESOLVCONF, 1)
+	AC_MSG_NOTICE([DISABLE_RESOLVCONF... 1])
+],[
+	AC_DEFINE(DISABLE_RESOLVCONF, 0)
+	AC_MSG_NOTICE([DISABLE_RESOLVCONF... 0])
 ])
 
 AC_CONFIG_COMMANDS([timestamp], [touch src/.dirstamp])
--- a/configure.ac
+++ b/configure.ac
@@ -350,16 +350,6 @@ AS_IF([test "x$with_ppp" = "xyes" -a "x$with_pppd" = "xyes"], [
    with_ppp="no"
 ])
 
-AC_PATH_PROG(RESOLVCONF_PATH, [resolvconf], [/sbin/resolvconf], "$PATH:/sbin:/usr/sbin")
-
-# resolvconf if present
-AC_ARG_WITH([resolvconf],
-        AS_HELP_STRING([--with-resolvconf],
-                       [Set the path to the resolvconf executable]),
-        RESOLVCONF_PATH="$withval"
-)
-
-
 AS_IF([test "x$with_ppp" = "xyes"], [
 	AC_DEFINE(HAVE_USR_SBIN_PPP, 1)
 	AC_MSG_NOTICE([HAVE_USR_SBIN_PPP... 1])
@@ -390,6 +380,16 @@ AS_IF([test "x$NETSTAT_PATH" != "x" ], [
     AC_MSG_NOTICE([NETSTAT_PATH...] $NETSTAT_PATH)
 ])
 
+# use resolvconf if present
+AC_PATH_PROG(RESOLVCONF_PATH, [resolvconf], [/sbin/resolvconf], "$PATH:/sbin:/usr/sbin")
+
+# allow override at configure time
+AC_ARG_WITH([resolvconf],
+        AS_HELP_STRING([--with-resolvconf],
+                       [Set the path to the resolvconf executable]),
+        RESOLVCONF_PATH="$withval"
+)
+
 AC_SUBST(RESOLVCONF_PATH)
 AS_IF([test "x$RESOLVCONF_PATH" != "x" ], [
     AC_MSG_NOTICE([RESOLVCONF_PATH...] $RESOLVCONF_PATH)
@@ -401,6 +401,7 @@ AS_IF([test "x$RESOLVCONF_PATH" = "xDISABLED"], [
 	AC_DEFINE(DISABLE_RESOLVCONF, 0)
 	AC_MSG_NOTICE([DISABLE_RESOLVCONF... 0])
 ])
+
 
 AC_CONFIG_COMMANDS([timestamp], [touch src/.dirstamp])
 AC_OUTPUT(Makefile)
--- a/configure.ac
+++ b/configure.ac
@@ -402,6 +402,32 @@ AS_IF([test "x$RESOLVCONF_PATH" = "xDISABLED"], [
 	AC_MSG_NOTICE([DISABLE_RESOLVCONF... 0])
 ])
 
+# the default for the --use-resolvconf runtime command line option
+AC_ARG_ENABLE([resolvconf],
+	AS_HELP_STRING([--enable-resolvconf],
+			[Enable usage of resolvconf by default, \
+			use --disable-resolvconf for the opposite]))
+
+# Checks for existence of specific files if not cross-compiling.
+AS_IF([test "x$enable_resolvconf" = "x" ], [
+  AC_CHECK_FILE([$RESOLVCONF_PATH],[
+    AS_IF([$RESOLVCONF_PATH -l &>/dev/null], [
+     enable_resolvconf="yes"
+    ],[
+     enable_resolvconf="no"
+    ])
+  ],[
+     enable_resolvconf="no"
+  ])
+])
+
+AS_IF([test "x$enable_resolvconf" = "xyes"], [
+	AC_DEFINE(USE_RESOLVCONF, 1)
+	AC_MSG_NOTICE([USE_RESOLVCONF... 1])
+],[
+	AC_DEFINE(USE_RESOLVCONF, 0)
+	AC_MSG_NOTICE([USE_RESOLVCONF... 0])
+])
 
 AC_CONFIG_COMMANDS([timestamp], [touch src/.dirstamp])
 AC_OUTPUT(Makefile)
--- a/configure.ac
+++ b/configure.ac
@@ -381,7 +381,7 @@ AS_IF([test "x$NETSTAT_PATH" != "x" ], [
 ])
 
 # use resolvconf if present
-AC_PATH_PROG(RESOLVCONF_PATH, [resolvconf], [/sbin/resolvconf], "$PATH:/sbin:/usr/sbin")
+AC_PATH_PROG(RESOLVCONF_PATH, [resolvconf], [DISABLED], "$PATH:/sbin:/usr/sbin")
 
 # allow override at configure time
 AC_ARG_WITH([resolvconf],
@@ -395,11 +395,11 @@ AS_IF([test "x$RESOLVCONF_PATH" != "x" ], [
     AC_MSG_NOTICE([RESOLVCONF_PATH...] $RESOLVCONF_PATH)
 ])
 AS_IF([test "x$RESOLVCONF_PATH" = "xDISABLED"], [
-	AC_DEFINE(DISABLE_RESOLVCONF, 1)
-	AC_MSG_NOTICE([DISABLE_RESOLVCONF... 1])
+	AC_DEFINE(HAVE_RESOLVCONF, 0)
+	AC_MSG_NOTICE([HAVE_RESOLVCONF... 0])
 ],[
-	AC_DEFINE(DISABLE_RESOLVCONF, 0)
-	AC_MSG_NOTICE([DISABLE_RESOLVCONF... 0])
+	AC_DEFINE(HAVE_RESOLVCONF, 1)
+	AC_MSG_NOTICE([HAVE_RESOLVCONF... 1])
 ])
 
 # the default for the --use-resolvconf runtime command line option
--- a/src/main.c
+++ b/src/main.c
@@ -62,7 +62,7 @@
 
 #if !DISABLE_RESOLVCONF
 #define RESOLVCONF_USAGE \
-"                    [--use-resolvconf=<0|1>] "
+"[--use-resolvconf=<0|1>] "
 #define RESOLVCONF_HELP \
 "  --use-resolvconf=[01]         if possible use resolvconf to update /etc/resolv.conf\n"
 #endif
@@ -74,8 +74,7 @@
 "                    [--otp-prompt=<prompt>] [--set-routes=<0|1>]\n" \
 "                    [--half-internet-routes=<0|1>] [--set-dns=<0|1>]\n" \
 PPPD_USAGE \
-RESOLVCONF_USAGE \
-"                    [--ca-file=<file>]\n" \
+"                    " RESOLVCONF_USAGE "[--ca-file=<file>]\n" \
 "                    [--user-cert=<file>] [--user-key=<file>]\n" \
 "                    [--trusted-cert=<digest>] [--use-syslog]\n" \
 "                    [--persistent=<interval>] [-c <file>] [-v|-q]\n" \
--- a/src/main.c
+++ b/src/main.c
@@ -60,6 +60,13 @@
 "                                /etc/ppp/ppp.conf\n"
 #endif
 
+#if !DISABLE_RESOLVCONF
+#define RESOLVCONF_USAGE \
+"                    [--use-resolvconf=<0|1>] "
+#define RESOLVCONF_HELP \
+"  --use-resolvconf=[01]         if possible use resolvconf to update /etc/resolv.conf\n"
+#endif
+
 #define usage \
 "Usage: openfortivpn [<host>[:<port>]] [-u <user>] [-p <pass>]\n" \
 "                    [--pinentry=<program>]\n" \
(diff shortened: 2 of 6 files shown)


DimitriPapadopoulos:
@mrbaseman Looks perfect at first sight. Has this been tested on FreeBSD and macOS? What happens on these platforms?
@adrienverge Could you perhaps test and comment how it works on Fedora? You probably have to disable _resolvconf_ explicitly on these platforms. If so future _Fedora_/_CentOS_/_EPEL_ packages need to be modified accordingly.

DimitriPapadopoulos:
In _autoconf_ I would prefer **HAVE_RESOLVCONF** instead of **DISABLE_RESOLVCONF** for consistency with other _autoconf_ options. Perhaps there's a reason why it's implemented this way?

src/main.c
```diff
@@ -60,14 +60,21 @@
 "                                /etc/ppp/ppp.conf\n"
 #endif
 
+#if !DISABLE_RESOLVCONF
+#define RESOLVCONF_USAGE \
+"[--use-resolvconf=<0|1>] "
+#define RESOLVCONF_HELP \
+"  --use-resolvconf=[01]         if possible use resolvconf to update /etc/resolv.conf\n"
```
DimitriPapadopoulos: Capital first letter:
if possible → If possible

adrienverge:
> @adrienverge Could you perhaps test and comment how it works on _Fedora_/_CentOS_/_EPEL_? You probably have to disable _resolvconf_ explicitly on these platforms. If so future _Fedora_/_CentOS_/_EPEL_ packages need to be modified accordingly.

It's been a while since I haven't manually changed/checked resolvconf, or used a Fortinet gateway... I wouldn't know where to start. However if you give me precise instructions or commands to run, I can test on my system and report results.

DimitriPapadopoulos:
@adrienverge If you have a _Fedora_ machine I suggest the following tests:
* Do not configure _resolvconf_ in any way, leave the _Fedora_ machine in its default state.
* Check out Martlin's `resolvconf_fix` branch for testing.
* Run `configure` and check that _resolvconf_ support is enabled by default (`openfortivpn --help` should show a line starting with `--use-resolvconf=[01]`).
* Run _openfortivpn_ against the test FortiGate: `openfortivpn <host> -u <user>` where `<host>` and `<user>` are in Martin's mail as well as the password. You should be able to reproduce the bug and verify that `/etc/resolv.conf` is not updated. Then run `openfortivpn <host> -u <user> --use-resolvconf=0`. This time `/etc/resolv.conf` should be updated.
* Rebuild _openfortivpn_ from scratch, run `configure --disable-resolvconf`. This time _resolvconf_ support should be disabled (`openfortivpn --help` should **not** show a line starting with `--use-resolvconf=[01]`).
* Run _openfortivpn_ against the test FortiGate: `openfortivpn <host> -u <user>` where `<host>` and `<user>` are in Martin's mail as well as the password. This time _openfortivpn_ should modify `/etc/resolv.conf` itself and by default.

DimitriPapadopoulos:
@adrienverge If you can verify all the above, then change the `openfortivpn.spec` file of upcoming _Fedora_ / _CentOS_ / _EPEL_ packages so that _configure_ is run with option `--disable-resolvconf`. Not sure how to do that though:
```
%build
autoreconf -fi
%configure
make %{?_smp_mflags} V=1
```
This command should show you how `%configure` is evaluated on your system:
```
rpm --eval '%configure' 
```
You'll probably end up with something like this:
```
%build
autoreconf -fi
CFLAGS="$RPM_OPT_FLAGS" ./configure--disable-resolvconf --prefix=%{_prefix} --mandir=%{_mandir} --sysconfdir=/etc`
make %{?_smp_mflags} V=1
```

mrbaseman:
@DimitriPapadopoulos I haven't done much testing on this branch yet, but

- On Ubuntu I'm working with this branch now from my home office and unfortunately, it doesn't work as expected yet. I have just noticed that on Ubuntu 16.04 _openresolv_ and _resolvconf_ are two independent packages that both provide a `resolvconf` command with different options, and the `-l` is not a good test in the configure script, because it is only available with _openresolv_
- when _resolvconf_ is **not** used, the nameservers are not removed again from `/etc/resolv.conf` so we have a bug there
- My Mac is broken, so I can't test on OS X anymore, until a new one is available.
- The machine with BSD is powered off in the office, so I can not test there until I'm back in the office or I find the time to install a FreeBSD VM at home

About the autoconf variable, yes, you are right HAVE_RESOLVCONF would be much better. I'll push a commit for that.

mrbaseman:
I have fixed 

- the capital letter in the help string,
- use HAVE_RESOLVCONF
- use DISABLED as a sane default if autoconf can't find resolvconf at configure time
- nameservers are removed again in the case when resolvconf is not used

The question how to detect at configure time if resolvconf is working is still open. I think it works for _openresolv_, but not for _resolvconf_

DimitriPapadopoulos:
I'll check with/without _resolvconf_ support on Ubuntu 18.04 again. Both seem to have worked for me.

I only have package `resolvconf` installed on Ubuntu 18.04. I have never installed `openresolv`. I'll give it a try...

DimitriPapadopoulos:
On Ubuntu 18.04 with package `resolvconf` installed and package `openresolv` **not** installed, with your branch of _openfortivpn_ configured to support _resolvconf_:

Before running _openfortivpn_:
```
$ cat /etc/resolv.conf 
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN


# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.

nameserver 127.0.0.53
$ 
```
After starting _openfortivpn_ with `openfortivpn --use-resolvconf=1`:
```
$ cat /etc/resolv.conf 
nameserver xxx.xxx.xxx.x
nameserver xxx.xxx.xxx.x
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.

nameserver 127.0.0.53
$ 
```
After stopping _openfortivpn_:
```
$ cat /etc/resolv.conf 
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.
nameserver 127.0.0.53
$ 
```
After starting _openfortivpn_ with `openfortivpn --use-resolvconf=0`:
```
$ cat /etc/resolv.conf
nameserver xxx.xxx.xxx.x
nameserver xxx.xxx.xxx.x
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.
nameserver 127.0.0.53
$ 
```
Everything seems to be working smoothly.

DimitriPapadopoulos:
What I'm not certain about is the effect of `./configure --disable_resolvconf`. I see _Autoconf_ still searching for `resolvconf` in the path and just setting `-DHAVE_RESOLVCONF=1 -DUSE_RESOLVCONF=0`. Thinking about it, I would expect `--disable_resolvconf` to skip `resolvconf` detection and set `HAVE_RESOLVCONF=0`.

mrbaseman:
Hi Dimitri,

unfortunately, the only thing we have learned from this test is that the **fallback** works smoothly. The fact that the nameservers are prepended to the original context of `resolv.conf` shows that the builtin fallback is active in all cases. Could you repeat the test with the `--enable-resolvconf` configure option, please? The nameservers should be merged **below** the generated comment.

On my Ubuntu 16.04 this works, with _resolvconv_ installed. _openresolv_ on the other hand should already be detected properly by the configure script, but for _resolvconf_ we still have to find a good _autoconf_ check.

I have rebased my branch to pick up your changes on master already

mrbaseman:
our messages have crossed. There are several important details:

- the configure option is `--disable-resolvconf` and not `--disable_resolvconf`
- it sets the default for the command line option `--use-resolvconf` at runtime to 0
- resolvconf support in general is compiled in, as long as the configure script finds a `resolvconf` command
- to completely disable resolvconf support at configure time you would have to use `./configure --with-resolvconf=DISABLED`
- to fully enable resolvconf support either the autoconf check must find a `resolvconf` command and either `resolvconf -l` must return a 0 return value or you need the `--enable-resolvconf` configure option (or we find a better test than the currently implemented check of the return value of `resolvconf -l`)

DimitriPapadopoulos:
On my Ubuntu 18.04 machine I don't have _openresolv_, just _resolvconf_. It used to work. Now it doesn't work any more because your branch expects a 0 return value. The return value is not 0 although _resolvconf_ does work:
```
$ dpkg -l resolvconf
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                        Version            Architecture       Description
+++-===========================-==================-==================-===========================================================
ii  resolvconf                  1.79ubuntu10.18.04 all                name server information handler
$ 
$ dpkg -l openresolv
dpkg-query: no packages found matching openresolv
$ 
$ resolvconf ; echo $?
resolvconf: Error: Command not recognized
Usage: resolvconf (-d IFACE|-a IFACE|-u|--enable-updates|--disable-updates|--updates-are-enabled)
99
$ 
$ resolvconf -l ; echo $?
resolvconf: Error: Command not recognized
Usage: resolvconf (-d IFACE|-a IFACE|-u|--enable-updates|--disable-updates|--updates-are-enabled)
99
$ 
```

DimitriPapadopoulos:
I think a return value of 99 is actually a marker of a functioning _resolvconf_. On _Fedora_ the return value is 1, not 99:
```
$ resolvconf ; echo $?
Expected either -a or -d on the command line.
1
$ 
$ resolvconf -l ; echo $?
resolvconf: option requires an argument -- 'l'
1
$ 
```

DimitriPapadopoulos:
As a follow-up to the above discussion and your suggestion to configure with `--enable-resolvconf`:

When running _Autoconf_ **without** `--enable-resolvconf` on Ubuntu 18.04 _openfortivpn_ runs _resolvconf_ only with explicit option `--use-resolvconf=1`:
```
$ ./configure --prefix=/opt/openfortivpn
[...]
checking for resolvconf... /sbin/resolvconf
configure: RESOLVCONF_PATH... /sbin/resolvconf
configure: HAVE_RESOLVCONF... 1
checking for /sbin/resolvconf... yes
[...]
$ 
```
When running _Autoconf_ **with** `--enable-resolvconf` on Ubuntu 18.04 _openfortivpn_ does run _resolvconf_ by default:
```
$ ./configure --prefix=/opt/openfortivpn --enable-resolvconf
[...]
checking for resolvconf... /sbin/resolvconf
configure: RESOLVCONF_PATH... /sbin/resolvconf
configure: HAVE_RESOLVCONF... 1
configure: USE_RESOLVCONF... 1
[...]
$ 
```
And in that case I do see `_openfortivpn_/_resolvconf_ properly modifying `/etc/resolv.conf` without the fallback mechanism from:
```
$ cat /etc/resolv.conf 
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.

nameserver 127.0.0.53
$ 
```
to:
```
$ cat /etc/resolv.conf 
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.

nameserver xxx.xx.252.252
nameserver xxx.xx.252.202
nameserver 127.0.0.53
search xxxx.de xxx.xxxx.de xxx.de
$ 
```

DimitriPapadopoulos:
On my Ubuntu 18.04 machine, just tried with _openresolv_ instead of _resolvconf_:
```
$ dpkg -l openresolv
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                        Version            Architecture       Description
+++-===========================-==================-==================-===========================================================
ii  openresolv                  3.8.0-1            amd64              management framework for resolv.conf
$ 
$ dpkg -l resolvconf
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                        Version            Architecture       Description
+++-===========================-==================-==================-===========================================================
un  resolvconf                  <none>             <none>             (no description available)
$ 
$ resolvconf ; echo $?
Usage: resolvconf [options] command [argument]

Inform the system about any DNS updates.

Commands:
  -a $INTERFACE    Add DNS information to the specified interface
                   (DNS supplied via stdin in resolv.conf format)
  -d $INTERFACE    Delete DNS information from the specified interface
  -h               Show this help cruft
  -i [$PATTERN]    Show interfaces that have supplied DNS information
                   optionally from interfaces that match the specified
                   pattern
  -l [$PATTERN]    Show DNS information, optionally from interfaces
                   that match the specified pattern

  -u               Run updates from our current DNS information

Options:
  -f               Ignore non existant interfaces
  -m metric        Give the added DNS information a metric
  -p               Mark the interface as private
  -x               Mark the interface as exclusive

Subscriber and System Init Commands:
  -I               Init the state dir
  -r $SERVICE      Restart the system service
                   (restarting a non-existent or non-running service
                    should have no output and return 0)
  -R               Show the system service restart command
  -v [$PATTERN]    echo NEWDOMAIN, NEWSEARCH and NEWNS variables to
  		   the console
  -V [$PATTERN]    Same as -v, but only uses configuration in
                   /etc/resolvconf.conf
0
$ resolvconf -l ; echo $?
0
$ 
```

Indeed the _Autoconf_ script detects _openresolv_ and enables `/sbin/resolvconf` support by default:
```
$ ./configure --prefix=/opt/openfortivpn
[...]
checking for resolvconf... /sbin/resolvconf
configure: RESOLVCONF_PATH... /sbin/resolvconf
configure: HAVE_RESOLVCONF... 1
checking for /sbin/resolvconf... yes
configure: USE_RESOLVCONF... 1
[...]
$ 
```

Then it works just fine, like the real _resolvconf_, changing from:
```
$ cat /etc/resolv.conf 
# This file is managed by man:systemd-resolved(8). Do not edit.
#
# This is a dynamic resolv.conf file for connecting local clients to the
# internal DNS stub resolver of systemd-resolved. This file lists all
# configured search domains.
#
# Run "systemd-resolve --status" to see details about the uplink DNS servers
# currently in use.
#
# Third party programs must not access this file directly, but only through the
# symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a different way,
# replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 127.0.0.53
$ 
```
to:
```
$ cat /etc/resolv.conf 
# Generated by resolvconf
search xxxx.de xxx.xxxx.de xxx.de
nameserver xxx.xx.252.252
nameserver xxx.xx.252.202
$ 
```

adrienverge:
Dimitri, Martin, I'm sorry, I didn't find time to test on Fedora nor CentOS, and I don't think I will be able until this week-end. I hope it's not blocking for you (Dimitri I see you got a Fedora VM working :+1:)

mrbaseman:
Ok, I have added the case of the 99 return value to configure.ac. @DimitriPapadopoulos, your observations on Ubuntu 18.04 fit well with my experiences on 16.40 (I know, I have to update my systems soon). The interesting question, however, does this help us on Fedora now?

DimitriPapadopoulos:
Yes, I'm certain it helps on _Fedora_. I will double-check. when I have time - I have much less time at my disposal than initially expected!

Then it all depends on how we want to package for _Fedora_:
1. totally remove `resolvconf` support
2. disable attempts to run `resolvconf`  by default but `--use-resolvconf=1` to run `resolvconf`

I believe 2 is better - someone might install a working _resolvconf_ on _Fedora_ and want to use it. For 2 I'm certain we don't need to change anything, just release a new version and repackage for _Fedora_ /_CentOS_/_EPEL_ - provided the build/packaging machine is representative of runtime machines and ships with a broken `resolvconf` like most _Fedora_ machines out there.

configure.ac
```diff
@@ -388,10 +380,58 @@ AS_IF([test "x$NETSTAT_PATH" != "x" ], [
...
+# the default for the --use-resolvconf runtime command line option
+AC_ARG_ENABLE([resolvconf],
+	AS_HELP_STRING([--enable-resolvconf],
+			[Enable usage of resolvconf by default, \
+			use --disable-resolvconf for the opposite]))
+
+# Checks for existence of specific files if not cross-compiling.
+AS_IF([test "x$enable_resolvconf" = "x" ], [
+  AC_CHECK_FILE([$RESOLVCONF_PATH],[
+    AS_IF([$RESOLVCONF_PATH -l &>/dev/null], [
```
DimitriPapadopoulos: Perhaps document the target distributions where you would expect `resolvconf -l` to return an exit status of 0?

configure.ac
```diff
@@ -388,10 +380,58 @@ AS_IF([test "x$NETSTAT_PATH" != "x" ], [
...
+     enable_resolvconf="yes"
+    ],[
+      AS_IF([ $RESOLVCONF_PATH &>/dev/null ; test $? -eq 99 ], [
```
DimitriPapadopoulos: Perhaps document the target distributions where you would expect `resolvconf` to return an exit status of 99?

DimitriPapadopoulos:
I have just tested you branch on _Fedora 31_ and it looks good to me:
```
$ ./configure --prefix=/opt/openfortivpn
[...]
checking for resolvconf... /usr/sbin/resolvconf
configure: RESOLVCONF_PATH... /usr/sbin/resolvconf
configure: HAVE_RESOLVCONF... 1
checking for /usr/sbin/resolvconf... yes
[...]
$ 
```
That's what we expect: _Autoconf_ does find `/usr/sbin/resolvconf` but identifies it as a broken substitute of the real _resolvconf_ and it is not used by default.

configure.ac
mrbaseman: done

configure.ac
mrbaseman: I've found another case (resolvectl acting as resolvconf drop in replacement):
https://www.freedesktop.org/software/systemd/man/resolvconf.html
and we have the resolvconf script which exits with 99 if an unknown command or not enough arguments are supplied https://salsa.debian.org/debian/resolvconf/-/blob/master/bin/resolvconf#L49

mrbaseman:
@DimitriPapadopoulos could you test once more on Fedora, please, so that we know if my latest commits still handle the broken resolvconf issue on Fedora correctly?

DimitriPapadopoulos:
Still works as expected on _Fedora_ at build-time:
```
$ ./configure --prefix=/opt/openfortivpn
[...]
checking for resolvconf... /usr/sbin/resolvconf
configure: RESOLVCONF_PATH... /usr/sbin/resolvconf
configure: HAVE_RESOLVCONF... 1
checking for /usr/sbin/resolvconf... yes
[...]
$ 
```
and at run-time:
```
$ sudo /opt/openfortivpn/bin/openfortivpn -u vpnuser fg60d.hlrs.de
VPN account password: 
[...]
INFO:   Got addresses: [10.212.134.200], ns [xxx.xx.252.252, xxx.xx.252.202]
INFO:   negotiation complete
INFO:   Got addresses: [10.212.134.200], ns [xxx.xx.252.252, xxx.xx.252.202]
[...]
$ 
```
with `/etc/resolv.conf` modified by the fallback mechanism as expected on _Fedora_ from:
```
$ cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 10.0.2.3
$ 
```
to:
```
$ cat /etc/resolv.conf 
nameserver xxx.xx.252.252
nameserver xxx.xx.252.202
search xxxx.de xxx.xxxx.de xxx.de
# Generated by NetworkManager
nameserver 10.0.2.3
$ 
```

configure.ac
DimitriPapadopoulos: And that `resolvectl` acting as a drop-in replacement for `resolvconf` is different from the one on _Fedora 31_ which doesn't work?

DimitriPapadopoulos: Because on _Fedora_ we have:
```
$ ls -l /usr/sbin/resolvconf
lrwxrwxrwx. 1 root root 17 Feb 11 14:25 /usr/sbin/resolvconf -> ../bin/resolvectl
$ 
$ ls -l /usr/bin/resolvectl 
-rwxr-xr-x. 1 root root 132368 Feb 11 14:25 /usr/bin/resolvectl
$ 
$ rpm -qf /usr/bin/resolvectl
systemd-243.7-1.fc31.x86_64
$ 
$ /usr/bin/resolvectl --help
resolvectl [OPTIONS...] {COMMAND} ...

Send control commands to the network name resolution manager, or
resolve domain names, IPv4 and IPv6 addresses, DNS records, and services.

  -h --help                    Show this help
     --version                 Show package version
     --no-pager                Do not pipe output into a pager
  -4                           Resolve IPv4 addresses
  -6                           Resolve IPv6 addresses
  -i --interface=INTERFACE     Look on interface
  -p --protocol=PROTO|help     Look via protocol
  -t --type=TYPE|help          Query RR with DNS type
  -c --class=CLASS|help        Query RR with DNS class
     --service-address=BOOL    Resolve address for services (default: yes)
     --service-txt=BOOL        Resolve TXT records for services (default: yes)
     --cname=BOOL              Follow CNAME redirects (default: yes)
     --search=BOOL             Use search domains for single-label names
                                                              (default: yes)
     --raw[=payload|packet]    Dump the answer as binary data
     --legend=BOOL             Print headers and additional info (default: yes)

Commands:
  query HOSTNAME|ADDRESS...    Resolve domain names, IPv4 and IPv6 addresses
  service [[NAME] TYPE] DOMAIN Resolve service (SRV)
  openpgp EMAIL@DOMAIN...      Query OpenPGP public key
  tlsa DOMAIN[:PORT]...        Query TLS public key
  status [LINK...]             Show link and server status
  statistics                   Show resolver statistics
  reset-statistics             Reset resolver statistics
  flush-caches                 Flush all local DNS caches
  reset-server-features        Forget learnt DNS server feature levels
  dns [LINK [SERVER...]]       Get/set per-interface DNS server address
  domain [LINK [DOMAIN...]]    Get/set per-interface search domain
  default-route [LINK [BOOL]]  Get/set per-interface default route flag
  llmnr [LINK [MODE]]          Get/set per-interface LLMNR mode
  mdns [LINK [MODE]]           Get/set per-interface MulticastDNS mode
  dnsovertls [LINK [MODE]]     Get/set per-interface DNS-over-TLS mode
  dnssec [LINK [MODE]]         Get/set per-interface DNSSEC mode
  nta [LINK [DOMAIN...]]       Get/set per-interface DNSSEC NTA
  revert LINK                  Revert per-interface configuration

See the resolvectl(1) man page for details.
$ 
```
I thought this `resolvctl` doesn't work, at least not as a ***drop-in*** replacement for``resolvconf`. But it might work with different set of command lline options.

DimitriPapadopoulos: OK, this is indeed the man page of the broken `resolvconf`/`resolvctl` on _Fedora_:
https://www.freedesktop.org/software/systemd/man/resolvconf.html

And indeed `resolvctl` is **not** a drop-in replacement for `resolvconf` in our use case. I've created this file:
```
$ cat PIPE_TO_RESOLVCONF 
nameserver xxx.xx.252.252
nameserver xxx.xx.252.202
search xxxx.de xxx.xxxx.de xx.de
```
On _Ubuntu_ 18.04:
```
$ cat /etc/resolv.conf
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.

nameserver 127.0.0.53
$ 
$ cat PIPE_TO_RESOLVCONF | /sbin/resolvconf -a "ppp0.openfortivpn"
$ 
$ cat /etc/resolv.conf
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.

nameserver xxx.xx.252.252
nameserver xxx.xx.252.202
nameserver 127.0.0.53
search xxxx.de xxx.xxxx.de xx.de
$ 
```
On _Fedora_ 31:
```
$ cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 10.0.2.3
$ 
$ cat PIPE_TO_RESOLVCONF | /sbin/resolvconf -a "ppp0.openfortivpn"
$ 
$ cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 10.0.2.3
$ 
```
I guess the `resolvctl` authors had not thought of the possibility to **pipe** parameters to `resolvconf`. They thought of it as a drop-in replacement with respect to command line options only.

My suggestion:
* In the short term, let's say for 1.12.1, the `resolvconf_fix` branch looks perfect in its current form: on _Fedora_ _openfortivpn_ uses the fallback mechanism.
* In the long term, let's say for 1.13.0, we could support `resolvctl` explicitly in addition to `resolvconf`. In both cases I would recommend using commend line arguments instead of piping into `resolvconf`.

mrbaseman:
Ok, then. I have already rebased this on the current master, yesterday. So, I'll squash and merge this, so that we can cherry-pick it as a single commit for a 1.12.1 bugfix release. Or maybe we release that version from the master branch as we always did? I think https://github.com/adrienverge/openfortivpn/commit/32f1949ffb78b9b7e447bff8ddd51bb7b58995ae is the only commit that might be a bit problematic and where we have said that we should do some more testing on different platforms.

PR #569, opened 2020-03-15
Compliant header guards

### LINKED ISSUE #567: reserved identifier violation
I would like to point out that an identifier like “[`_OPENFORTIVPN_LOG_H`](https://github.com/adrienverge/openfortivpn/blob/f3909d8cfb36c0660fc4bcbb9167d891c67016d3/src/log.h#L18 "Update candidate")” [does not fit](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28IncludeGuard%29 "Do not use identifiers which are reserved for the compiler implementation.") to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

DimitriPapadopoulos:
Fixes #567.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -15,8 +15,8 @@
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef _OPENFORTIVPN_CONFIG_H
-#define _OPENFORTIVPN_CONFIG_H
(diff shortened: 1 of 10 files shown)


elfring:
Thanks for your small source code improvement.

PR #576, opened 2020-03-19
Doubled BUFSZ to fix #575

### LINKED ISSUE #575: Received bad header from gateway when you have tons of remote subnets
Hi

It seems you get the "Received bad header from gateway" error when your vpn server has tons of routes to remote subnets to push to the client.

I have narrowed the issue down to a hardcoded buffer size. 

Here is the log, and a generous size increase of the buffer size... is it safe ? 

It looks good - works well here at least.

[openfortivpn.log](https://github.com/adrienverge/openfortivpn/files/4356262/openfortivpn.log)

m33m33:
Seems a simplistic fix for issue #575 

"it works here" ;-)

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -30,7 +30,7 @@
 #include <unistd.h>
 #include <arpa/inet.h>
 
-#define BUFSZ 0xF000
+#define BUFSZ 0x10000
 
 /*
  * URL-encodes a string for HTTP requests.
--- a/src/http.c
+++ b/src/http.c
@@ -30,7 +30,7 @@
 #include <unistd.h>
 #include <arpa/inet.h>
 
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Ah, nice catch! And we thought `0x8000` "would be enough for everyone"! Perhaps use `0x10000`? For no peculiar reason. I just happen to like it more 😊

m33m33:
> Ah, nice catch! And we thought `0x8000` would be "enough for everyone"! Perhaps use `0x10000`? For no peculiar reason. I just happen to like it more

Looks good to me.

DimitriPapadopoulos:
Yes,  I'm terribly fond of round numbers.

PR #587, opened 2020-03-22
Always define RESOLVCONF_USAGE / RESOLVCONF_HELP

DimitriPapadopoulos:
While either `HAVE_USR_SBIN_PPPD` or `HAVE_USR_SBIN_PPP` **must** be defined,
`HAVE_RESOLVCONF` might not be defined - typically when _resolvconf_ is not
available at build-time. `RESOLVCONF_USAGE` and `RESOLVCONF_HELP` should always
be defined, if needed as empty strings, so as to not break subsequent
_printf_() calls.


mrbaseman:
@DimitriPapadopoulos thanks for fixing this. I would have solved it exactly the same way as you have already done. I have spent the weekend with  my family.

Issue #590, opened 2020-03-23
Setting DNS parameters on Fedora


DimitriPapadopoulos:
I suspect the official way to modify DNS parameters on _CentOS_ >= 8 and _Fedora_ is running `resolvectl`, at least when `resolvectl` is available.

Therefore I would recommend the following mechanism when `--set-dns=1`:
1. If `resolvectl` is available and works properly, run it to modify DNS parameters. Decide at build-time or run-time?
2. Else if `resolvconf` is available and works properly (either [_openresolv_](https://roy.marples.name/projects/openresolv/) or [_resolvconf_](https://sources.debian.org/src/resolvconf/) but **not** [_resolvectl_](https://www.freedesktop.org/software/systemd/man/resolvectl.html)), run it to modify DNS parameters. Decide at build-time or run-time?
3. Else fall back on modifying `/etc/resolv.conf` directly.

I recommend the above order, `resolvectl` before `resolvconf`, because `resolvconf` might be available but doesn't seem to be working properly on _Fedora_.

Man pages:
* [resolvectl(1)](https://www.freedesktop.org/software/systemd/man/resolvectl.html) from _systemd_
* [resolvconf(8)](https://manpages.debian.org/stretch/resolvconf/resolvconf.8.en.html) from [_Debian_](https://sources.debian.org/src/resolvconf/)
* [resolvconf(8)](https://manpages.debian.org/testing/openresolv/resolvconf.8.en.html) from [_openresolv_](https://roy.marples.name/projects/openresolv/)

Hopefully we can get some info from this bug report:
https://bugzilla.redhat.com/show_bug.cgi?id=1815605

grey-olli:
When using vpn nameservers I need nameserver records to be _inserted_ rather than appended by openfortivpn. It is unclear for me is this mode is completely impossible or not in my configuration. I'm currently using fedora 31 xfce spin, /etc/resolv.conf is a symlink to systemd-resolved /usr/lib/systemd/resolv.conf  , NetworkManager is in 'default' mode . It receives dns from dhcp and in /run/systemd/resolve/resolve.conf I see:

[root@dell15 ~]# tail -n 4 /run/systemd/resolve/resolv.conf 
nameserver 8.8.4.4
nameserver 10.126.126.137
nameserver 172.16.116.194
search dell15.localdomain
[root@dell15 ~]# ls -l /etc/resolv.conf 
[...]

PR #591, opened 2020-03-23
Skip resolvectl for now

DimitriPapadopoulos:
On Fedora _resolvectl_ does not work as expected when invoked as `resolvconf`.
Also the exit status is not 0 so the current test never works actually:
```sh
$ resolvconf -a -f ; echo $?
Expected interface name as argument.
1
$ 
```

I suggest this goes in _openfortivpn_ 1.13.2.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -408,25 +408,20 @@ AC_ARG_ENABLE([resolvconf],
 			[Enable usage of resolvconf by default, \
 			use --disable-resolvconf for the opposite]))
 
-# Checks for to determine how resolvconf works is it is installed
-# openresolv supports a -l option that lists the active configurations and returns 0
-# resolvctl supports -a -f which silently exits if no interface is suppplied
-# the resolvconf script in Ubuntu/Debian does not support listing
-# but returns 99 if invoked without parameters
-#
+# Determine how resolvconf works at build-time if it is installed:
+# * openresolv supports option -l that lists active configurations and returns 0
+# * resolvconf in Ubuntu/Debian does not support listing but returns 99
+#   if invoked without parameters
+# * skip resolvectl which does not work as expected when invoked as resolveconf
 AS_IF([test "x$enable_resolvconf" = "x" ], [
   AC_CHECK_FILE([$RESOLVCONF_PATH],[
     AS_IF([$RESOLVCONF_PATH -l &>/dev/null], [
      enable_resolvconf="yes"
     ],[
-      AS_IF([$RESOLVCONF_PATH -a -f &>/dev/null], [
+      AS_IF([$RESOLVCONF_PATH &>/dev/null ; test $? -eq 99], [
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Instead implement #590.

Hopefully we will have some input from:
https://bugzilla.redhat.com/show_bug.cgi?id=1815605

mrbaseman:
The intention actually was to catch this return value of 1. If resolvctl works as described in the man page it should support the -f option when called es resolvconf and silently ignore a missing interface. If it does not work properly, we should not use it.

But maybe I have misunderstood this sentence. Maybe "missing interface" means an interface name supplied to the call, but not describing an existing interface.

DimitriPapadopoulos:
Let me detail what happens on _Fedora_:

I may be wrong but I had understood that you were expecting the exit status of `resolvconf -a -f` to be 0 based on the [resolvectl(1)](https://www.freedesktop.org/software/systemd/man/resolvectl.html) man page. I would not necessarily interpret "_will silently execute no operation_" as meaning that the exit status is 0, just that no message is written to the terminal. In any case both assumptions are wrong without an interface passed as an argument to `-a`:  an error message is written to the terminal and the exist status is 1:
```sh
$ sudo resolvconf -a -f ; echo $?
Expected interface name as argument.
1
$ 
$ sudo resolvconf -d -f ; echo $?
Expected interface name as argument.
1
$ 
```

Unless I am missing something, the existing test attempts to catch an exit status of 0, not an exit status of 1, doesn't it?

Finally invoking _resolvectl_ as `resolvconf` in "real life" doesn't seem to be working either, at least on _Fedora_, see:
https://bugzilla.redhat.com/show_bug.cgi?id=1815605

Running this as root on _Fedora_ 31:
```sh
$ cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 10.0.2.3
$ 
$ cat MY_FILE 
namsserver 1.1.1.1
nameserver 1.0.0.1
$ 
$ cat MY_FILE | /usr/sbin/resolvconf -a enp0s3 -f ; echo $?
0
$ 
$ cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 10.0.2.3
$ 
$ cat MY_FILE | /usr/sbin/resolvconf -a enp0s3 ; echo $?
0
$ 
$ cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 10.0.2.3
$ 
```

DimitriPapadopoulos:
Ah, the exit status is 0 when using a fake interface name with `-f` - but of course not without `-f``:
```sh
$ /usr/sbin/resolvconf -a foobar -f ; echo $?
0
$ 
$ usr/sbin/resolvconf -a foobar; echo $?
Unknown interface 'foobar': No such device
1
$ 
```

That doesn't change the fact that invoking _ressolvectl_ as `resolvconf` does not seem to be working.

PR #592, opened 2020-03-23
Cosmetic fixes in configure.ac

DimitriPapadopoulos:
Depends on #591.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -219,13 +219,13 @@ LIBS="${save_openssl_LIBS}"
 # When cross compile, don't run the tests.
 AC_ARG_WITH([rt_dst],
 	AS_HELP_STRING([--with-rt_dst],
-			  [disable rtentry with rt_dst testing (for linux target when cross compile)]),
+	               [disable rtentry with rt_dst testing (for linux target when cross compile)]),
 	AS_IF([test "x$with_rt_dst" = "xyes"],[
-	AC_MSG_NOTICE([HAVE_RT_ENTRY_WITH_RT_DST... 1])
-	AC_DEFINE(HAVE_RT_ENTRY_WITH_RT_DST)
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
@mrbaseman OK, we have few changes in `master` for now so we can release a 1.13.2 with fixes for FreeBSD. Do you agree making this priority, before merging other pull requests liek this one. This way we might avoid a 1.13 branch.

mrbaseman:
I would merge #592 #593 and #595 (all of them are really minor fixes or code cleanup) for the 1.13.2 and then continue working on #524 for instance

DimitriPapadopoulos:
OK for merging them #592, #593, #595. Could you do that and then test on FreeBSD again?

By the way, could these fixes be of interest on macOS too? Homebrew checks for 1.13.1 have not completed yet so I don't know if there are build issues:
https://github.com/Homebrew/homebrew-core/pull/52021

mrbaseman:
hmm... I have to rebase the other ones first. This time I have created new branches that are based on the other pull requests (in the hope that I don't have to rebase, but it didn't work).

The macOS issue was the same as on Fedora (as you have already mentioned [in the Homebrew repo](https://github.com/Homebrew/homebrew-core/pull/52021#issuecomment-602813155)), but I thought it makes sense to repeat this statement here.

While writing this answer I have merged the other ones in the meantime, and I'll re-test on FreeBSD now (a colleague of mine was on-site this morning and has powered up the machine on which I have a couple of VMs, including the FreeBSD testing installation).

mrbaseman:
I have tested the build system on FreeBSD again with  #592, #593, #595 merged, and even if the ppp binary is renamed during configure time the fallback to the default path works fine. Also output redirection during the configure tests works now.

PR #593, opened 2020-03-23
improve help strings for resolvconf configure options

mrbaseman:
this depends on #592

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -382,7 +382,10 @@ AC_PATH_PROG(RESOLVCONF_PATH, [resolvconf], [DISABLED], "$PATH:/sbin:/usr/sbin")
 # allow override at configure time
 AC_ARG_WITH([resolvconf],
 	AS_HELP_STRING([--with-resolvconf],
-	               [Set the path to the resolvconf executable]),
+	               [Set the path to the resolvconf executable. \
+		       Set this to "DISABLED" to fully disable resolvconf support. \
+		       In that case it will not be compiled in and therefore be \
+		       unavailable at runtime.]),
 	RESOLVCONF_PATH="$withval"
 )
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -382,7 +382,10 @@ AC_PATH_PROG(RESOLVCONF_PATH, [resolvconf], [DISABLED], "$PATH:/sbin:/usr/sbin")
 # allow override at configure time
 AC_ARG_WITH([resolvconf],
 	AS_HELP_STRING([--with-resolvconf],
-	               [Set the path to the resolvconf executable]),
+	               [Set the path to the resolvconf executable. \
+		       Set this to "DISABLED" to fully disable resolvconf support. \
```
DimitriPapadopoulos: Perhaps indent by one more space here?

configure.ac
```diff
@@ -401,8 +404,10 @@ AS_IF([test "x$RESOLVCONF_PATH" = "xDISABLED"], [
 # the default for the --use-resolvconf runtime command line option
 AC_ARG_ENABLE([resolvconf],
 	AS_HELP_STRING([--enable-resolvconf],
-	               [Enable usage of resolvconf by default, \
-	                use --disable-resolvconf for the opposite]))
+	               [Enable usage of resolvconf at runtime by default. \
+	                Use --disable-resolvconf for the opposite, note that \
```
DimitriPapadopoulos: Compare indentation above with here.

mrbaseman:
I had this one already merged, when I have seen your comment. However, I have squeezed the fix into #596. I have noticed that in several places spaces were used for indentation (which is not compliant with the linux coding style)

DimitriPapadopoulos:
Yes, I don't know about the spaces:
* The Linux kernel coding style applies to C, not necessarily _Autoconf_.
* This is the only rule I don't like - it is currently disabled by the way.  I'm OK with tabs for indentation, except when I want to align with the above line and tabs are not sufficient:
```
	const char variable[] = "indented with a tab only" \
	                        "Indented with a tab followed by spaces";
```
Also I know Linux suggest editors should display tabs as 8 spaces and that anything else is crap. Yet with more and more online editors, this is not always possible or simply convenient.

adrienverge:
My two cents: I agree using tabs + spaces *within the same line*, like on Dimitri's example, and like in the kernel source code, is not very practical. I'd be in favor of using only spaces in autoconf, and why not, one day, in all openfortivpn source code (if it's easy to do, considering linting scripts we have).

mrbaseman:
Ok. Me too, I didn't care much about a consistent indentation in configure.ac so far. 

It just happened that Dimitri's suggestion above resulted in 2 tabs + exactly 8 space characters. 

Therefore I did a global search&replace of 8 spaces by 1 tabulator in the whole file, and my editor has found a few more indentations it could "fix".

DimitriPapadopoulos:
@adrienverge It looks like you have moved from C/C++ to Python - like myself!

adrienverge:
> Therefore I did a global search&replace of 8 spaces by 1 tabulator in the whole file, and my editor has found a few more indentations it could "fix".

Sure, I understand.

> @adrienverge It looks like you have moved from C/C++ to Python - like myself!

Haha, this explains that!

PR #596, opened 2020-03-24
openfortivpn version 1.13.2

mrbaseman:


mrbaseman:
I have updated the changelog. 
My feeling is that the other commits since 1.13.1 are not important enough to be mentioned in the high level changelog. Ok?

mrbaseman:
I have squeezed another commit into this PR, which should probably stay separated, so please do not squash. Unfortunately travis-ci does not report back now, although the tests were run successfully.

mrbaseman:
I've updated the time stamp of my 1.13.2 commit. This has convinced github to fetch the test result from travis-ci.

adrienverge:
OK, thanks Martin! I won't squash.

Building packages for Fedora and CentOS is some work, do you think this new release needs some? (compared to 1.13.1 compiled this morning)

mrbaseman:
No, we don't need Fedora or Centos packages. The changes are all about the build system and cosmetic code cleanup.

Issue #566, opened 2020-03-15
Complete build options for Pthread API


elfring:
Would you like to add the configuration script “[AX_PTHREAD](https://www.gnu.org/software/autoconf-archive/ax_pthread.html "How to build C programs using POSIX threads")” to [your build specification](https://github.com/adrienverge/openfortivpn/blob/cd9368c6a1b4ef91d77bb3fdbe2e5bc34aa6f4c4/configure.ac#L22 "Update candidate")?

DimitriPapadopoulos:
Sure, why not. But can you expand on why it is needed on _openfortivpn_ supported platforms?

elfring:
I suggest to improve software portability also for the affected build script.

DimitriPapadopoulos:
I had a look at AX_PTHREAD. I'm not convinced. It requires additional dependencies and is complicated. _openfortivpn_ is portable enough as is.

DimitriPapadopoulos:
When running `configure` error I see:
```
$ ./configure
[...]
checking for pthread_cancel... no
configure: error: Required function not found
$ 
```
From `config.status`:
```
[...]
configure:6092: checking for pthread_cancel
configure:6092: gcc -o conftest -g -O2   conftest.c -lutil  >&5
/tmp/ccbMcx1t.o: In function `main':
/foo/bar/openfortivpn/conftest.c:138: undefined reference to `pthread_cancel'
collect2: error: ld returned 1 exit status
configure:6092: $? = 1
configure: failed program was:
[...]
| char pthread_cancel ();
| /* The GNU C library defines this for functions which it implements
|     to always fail with ENOSYS.  Some functions are actually named
|     something starting with __ and the normal name is an alias.  */
| #if defined __stub_pthread_cancel || defined __stub___pthread_cancel
| choke me
| #endif
| 
| int
| main ()
| {
| return pthread_cancel ();
|   ;
|   return 0;
| }
configure:6092: result: no
configure:6099: error: Required function not found
[...]
```

elfring:
>It requires additional dependencies and is complicated.

The safe determination of appropriate build parameters can be challenging.
Would you like to reuse any corresponding system variables?


>`checking for pthread_cancel... no`

Do you really stumble on a software library without support for thread cancelability?

DimitriPapadopoulos:
I'm sorry. I'm unable to understand your questions.

DimitriPapadopoulos:
[...]

PR #603, opened 2020-03-28
Use AX_PTHREAD autoconf macro

DimitriPapadopoulos:
[AX_PTHREAD](https://www.gnu.org/software/autoconf-archive/ax_pthread.html) is supposed to be more portable/standard.
Not sure I'm using it properly though. Not sure it's very useful either.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -33,7 +33,13 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 # Checks for libraries.
 PKG_CHECK_MODULES(OPENSSL, [libssl >= 0.9.8 libcrypto >= 0.9.8], [], [AC_MSG_ERROR([Cannot find OpenSSL 0.9.8 or higher.])])
-AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthread.])])
+AC_MSG_CHECKING([POSIX threads support])
+AX_PTHREAD([
+	CLIBS="$LIBS $PTHREAD_LIBS"
+	CFLAGS="$CFLAGS $PTHREAD_CFLAGS"
+	LDFLAGS="$LDFLAGS $PTHREAD_CFLAGS"
+	CC="$PTHREAD_CC"],
+AC_MSG_ERROR([Cannot configure POSIX threads support.]))
 AC_CHECK_LIB([util], [forkpty], [], [AC_MSG_ERROR([Cannot find libutil.])])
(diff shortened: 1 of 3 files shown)


DimitriPapadopoulos:
Need to add [autoconf-archive](https://www.gnu.org/software/autoconf-archive/) as a build-time dependency.

DimitriPapadopoulos:
@mrbaseman Could you maybe check it works on FreeBSD?

Meanwhile I will try to get Travis CI to build on mac0S and FreeBSD.

configure.ac
```diff
@@ -19,7 +19,13 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 # Checks for libraries.
 PKG_CHECK_MODULES(OPENSSL, [libssl >= 0.9.8 libcrypto >= 0.9.8], [], [AC_MSG_ERROR([Cannot find OpenSSL 0.9.8 or higher.])])
-AC_CHECK_LIB([pthread], [pthread_create], [], [AC_MSG_ERROR([Cannot find libpthread.])])
+AC_MSG_CHECKING([POSIX threads support])
+AX_PTHREAD([
+	CLIBS="$LIBS $PTHREAD_LIBS"
+	CFLAGS="$CFLAGS $PTHREAD_CFLAGS"
+	LDFLAGS="$LDFLAGS $PTHREAD_CFLAGS"
```
mrbaseman: should this perhaps be `LDFLAGS="$LDFLAGS $PTHREAD_LDFLAGS"` here?

mrbaseman:
I understand this is due to the suggestions in #566 but actually I fail to see the benefit of this change.

configure.ac
DimitriPapadopoulos: Mmmh... Probably!

DimitriPapadopoulos:
Yes, I'm not convinced myself: _openfortivpn_ works and is portable enough so why change? On the other hand it means we have less code to maintain ourselves, at the expense of possible initial issues caused by the change.

mrbaseman:
Maybe I miss something, but where do we have to maintain less code? 

The [documetnation of ax_pthread](https://www.gnu.org/software/autoconf-archive/ax_pthread.html) sais "you might link with $PTHREAD_CC $CFLAGS $PTHREAD_CFLAGS $LDFLAGS ... $PTHREAD_LIBS $LIBS" - so $PTHREAD_CFLAGS was counter intuitive but correct. Sorry.

DimitriPapadopoulos:
While there is no clear benefit in switching to `AX_PTHREAD`, the additional dependency on the Autoconf Archive adds complexity. Closing this PR.

PR #606, opened 2020-03-28
Get Travis CI to build on macOS and FreeBSD

### LINKED ISSUE #605: Travis CI: support for macOS and FreeBSD
Add `os: osx` and `os: freebsd` to `.travis.yml`:
[Build Environment Overview](https://docs.travis-ci.com/user/reference/overview/)

DimitriPapadopoulos:
Fixes #605.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+os:
+- linux
+- osx
+- freebsd
+
 language: c
 
 sudo: false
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Need to switch to the OpenSSL from Homebrew when bulding](https://medium.com/@timmykko/using-openssl-library-with-macos-sierra-7807cfd47892).

DimitriPapadopoulos:
* We now attempt to build on mac0S and FreeBSD too.
* We run tests on Linux only.

By the way, do we still need to install _astyle_ ourselves? Version 3.1 has been released in 2018 and is available as a package in most Linux distributions and _Homebrew_. Even if a new version 3.2 was released I wouldn't expect it to much disruptive - we will be running `checkpatch.pl` which is more stern.

mrbaseman:
@DimitriPapadopoulos  Wow. Thanks a lot for this work. So, we have a faster feedback if something gets broken (at least what's detectable at compile time) on all the platforms that we support.

DimitriPapadopoulos:
Ouch! The downside is that Travis CI takes for ever to run!

Is this a temporary situation or not? Hard to tell as the only feedback is:
> **continuous-integration/travis-ci** *Expected — Waiting for status to be reported*

mrbaseman:
this has happened to me occasionally. If I had a browser window open, it helped to navigate to travis and watch the output. If not, the only solution that I have found was to trigger another run (just update the date of the latest commit for instance) and quickly open the travis page to watch how the build proceeds :)  If you find a better way to see the details, please let me know.

PR #609, opened 2020-03-30
Fix Coverity defects

DimitriPapadopoulos:
CID 200508: String not null terminated (STRING_NULL)

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -570,6 +570,7 @@ static int tcp_connect(struct tunnel *tunnel)
 		        tunnel->config->gateway_port,
 		        inet_ntoa(tunnel->config->gateway_ip),
 		        tunnel->config->gateway_port);
+
 		ssize_t bytes_written = write(handle, request, strlen(request));
 
 		if (bytes_written != strlen(request)) {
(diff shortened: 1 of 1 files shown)


src/tunnel.c
```diff
@@ -584,27 +585,22 @@ static int tcp_connect(struct tunnel *tunnel)
...
 			static const char HTTP_STATUS_200[] = "200";
 
+			if (bytes_read < 1) {
+				log_error("Proxy response does not contain \"%s\" as expected.\n",
+				          HTTP_STATUS_200);
+				goto err_proxy_response;
+			}
+
+			// detect "200"
 			response = strstr(request, HTTP_STATUS_200);
```
mrbaseman: I don't understand how you can tell that the HTTP status code 200 has not been found before you do the strstr(). 
Sure, if bytes_read < 1 no character is returned by the read() command, which is a situation which we should catch at this point, but the status code is a different thing which we have to catch later, or maybe I'm missing something.

DimitriPapadopoulos:
The important thing is that this is a loop and that `if (bytes_read < 1` then we jump to `err_proxy_response` and we will not have seen 200.

I guess the real question here is the cause of `(bytes_read < 1)`. My guess was that the HTTP status is not 200 hence we reach the end of the inbound data. But it's true there could be other reasons - but whatever the reasons we won't have seen 200.

mrbaseman:
Ah, yes, that's the most probable case. It's just confusing that this error message comes right before the actual check. Maybe add a comment "// we have reached the end of the inbound data and not seen HTTP status code 200"?

DimitriPapadopoulos:
Done.

I'm afraid I'll have to revert builds on macOS and/or FreeBSD by the way. Travis CI runs last forever since that change.

DimitriPapadopoulos:
This part of the code would benefit from a 3rd party HTTP library that would simply read the HTTP header until its end and return the HTTP status code - whether it is 200 or not.

DimitriPapadopoulos:
OK to merge the pull request?

mrbaseman:
yes. thanks for taking care about this issue. Can you merge? I don't want to disturb you work on several branches by an unexpected merge in the wrong moment...

Issue #564, opened 2020-03-12
Add keepalive feature


pgporada:
I have a test fortigate device that hits the 15 minute timeout just like #485 . The forticlientsslvpn_cli has a keepalive flag and doesn't time me out. Could something similar be implemented in openfortivpn? Let me know if you need any information. There are long periods of time when I won't need to access anything on the fortigate network, but when I do and have to reauth it's a bit annoying.

```
forticlientsslvpn_cli [--proxy proxyaddress:proxyport] --server vpnserveraddress:vpnport [--proxyuser proxyuser] [--vpnuser vpnuser] [--pkcs12 pkcs12path] [--keepalive]
```
![forticlientvpn](https://user-images.githubusercontent.com/2382565/76531571-c9409880-644b-11ea-8ef8-b0c039b6bcc6.png)

DimitriPapadopoulos:
I've tried _FortiClient_ 4.4.2334 on a _Fedora_ virtual machine. I do see TCP Keep-Alive packets from the _Fedora_ machine to the _FortiGate_ appliance every 27 seconds,. I see these TCP Keep-Alive packets **with or without** the `--keepalive` option, so I'm wondering whether the Keep-Alive can be enabled by the FortiGate appliance in addition to the local `--keep-alive` option. And where do these 27 seconds come from?
![Wireshark](https://user-images.githubusercontent.com/3234522/77051440-92c6c880-69cb-11ea-9c8a-145f9d73c5de.png)

DimitriPapadopoulos:
Now the strange thing is that the _FortiGate_ appliance we connect to in the above example sends the following XML configuration:
```
<client-config save-password='off' keep-alive='off' auto-connect='off' />
```
My understanding is that TCP Keep-Alive should be off, at least by default. I cannot explain why it is on by default, with or without the `--keep-alive` option. And again where do these 27 seconds come from?

DimitriPapadopoulos:
This is related to DTLS, probably not relevant here:
```
<dtls-config heartbeat-interval='10' heartbeat-fail-count='10' heartbeat-idle-timeout='10' client-hello-timeout='10' />
```

mrbaseman:
I assume it's our test appliance. I can see a line `set keep-alive disable` in the configuration. And I can't find the 27 seconds configured anywhere.

DimitriPapadopoulos:
[...]

PR #613, opened 2020-03-31
implement tcp keep-alive feature

mrbaseman:
- this commit attempts to solve #564 
- still to be tested (I'm not sure if it is really that easy)
- I have squashed and rebased my branch on current master now
- I have fixed the coding style and squashed this change, too

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -596,6 +596,7 @@ int io_loop(struct tunnel *tunnel)
 	int tcp_nodelay_flag = 1;
 	int ret = 0;		// keep track of pthread_* return value
 	int fatal = 0;		// indicate a fatal error during pthread_* calls
+	int flags = 1;
 
 	pthread_t pty_read_thread;
 	pthread_t pty_write_thread;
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
I am very far from understanding this keep alive stuff. Just wondering though, should we apply options to the SSL socket or use `pppd` LCP options instead?
`lcp-echo-adaptive`
`lcp-echo-interval`

mrbaseman:
I think we can close this one, since keep-alive is already handled on a lower level. With _openfortivpn_ 1.15.0 the error messages from `pppd` are properly shown:
```
INFO:   Interface ppp0 is UP.
INFO:   Setting new routes...
No response to 4 echo-requests
Serial link appears to be disconnected.
Connect time 18.0 minutes.
Sent 230882 bytes, received 576408 bytes.
Connection terminated.
INFO:   Cancelling threads...
INFO:   Cleanup, joining threads...
INFO:   Setting ppp0 interface down.
INFO:   Restoring routes...
ERROR:  pppd: The link was terminated because the peer is not responding to echo requests.
INFO:   Terminated pppd.
INFO:   Logged out.
``` 
The only advantage might be that the socket options would allow us to increase the the values, but I haven't received any positive feedback on this change.

Issue #600, opened 2020-03-26
Improve DNS handling?


DimitriPapadopoulos:
### Rationale

Setting DNS parameters has been raising different issues:
* race condition between _openfortivpn_ itself and _pppd_ in earlier versions of _openfortivpn_ prior to #486 and version 1.12.0
* race condition between _openfortivpn_ and _NetworkManager_ (see for example #590)
* mechanisms for setting DNS might be available at build-time but not run-time, or at run-time but not build-time (for example `resolvectl` from _systemd_ is often available at run-time on _Fedora_ machines but is not available at build-time by default)
* mechanisms for setting DNS might be available but not working (for example `resolvconf` from _systemd_ does not work as expected on _Fedora_ as seen in #555 )

Especially _Fedora_ and _CentOS_ have been challenging so far. Short of interacting with _NetworkManager_ directly when _NetworkManager_ is installed **and** used, I suspect the official way to modify DNS parameters on _CentOS_ >= 8 and _Fedora_ is running `resolvectl` which has been part of the _systemd_ package since _CentOS_ >= 8 (not _CentOS_ 7).

### Proposal for setting DNS parameters

Is it worth trying to further improve the DNS configuration mechanism when `--set-dns=1`? Perhaps using the following mechanisms and other ones you can think of?
1. If _NetworkManager_ is installed **and** in use (how to detect that?) interact with _NetworkManager_ directly? Perhaps through `nmcli`?
2. If `resolvectl` from _systemd_ is available run it to modify DNS parameters.
3. If `resolvconf` is available and works properly (either [_openresolv_](https://roy.marples.name/projects/openresolv/) or [_resolvconf_](https://sources.debian.org/src/resolvconf/) but **not** [_resolvectl_](https://www.freedesktop.org/software/systemd/man/resolvectl.html)), run it to modify DNS parameters.
4. Fall back on modifying `/etc/resolv.conf` directly.

_NetworkManager_ is a special case: I am not sure about supporting _NetworkManager_ because _NetworkManager_ already supports _openfortivpn_ through [_NetworkManager-fortisslvpn_](https://github.com/GNOME/NetworkManager-fortisslvpn). It's the other way round.

### Documentation

Man pages:
* [nmcli(1)](https://developer.gnome.org/NetworkManager/stable/nmcli.html) from _NetworkManager_
* [resolvectl(1)](https://www.freedesktop.org/software/systemd/man/resolvectl.html) from _systemd_
[...]

PR #615, opened 2020-04-01
Add support for systemd-resolved

da-x:
These two commits make it possible to update systemd-resolved with the DNS server configuration upon a successful connect. First commit adds a generic script invocation, and the second adds an example script for updated `systemd-resolved`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Releases
 This high level changelog is usually updated when a release is tagged.
 On the master branch there may be changes that are not (yet) described here.
 
+### 1.20.2
+
+* [-] fix regression: do attempt to apply duplicate routes, log INFO instead of WARN
+* [-] minor changes in log messages
+
 ### 1.20.1
 
 * [-] fix version string in configure.ac
--- a/configure.ac
+++ b/configure.ac
@@ -2,7 +2,7 @@
 # Process this file with autoconf to produce a configure script.
 
 AC_PREREQ([2.63])
-AC_INIT([openfortivpn], [1.20.1])
+AC_INIT([openfortivpn], [1.20.2])
 AC_CONFIG_SRCDIR([src/main.c])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 


DimitriPapadopoulos:
Perhaps it would be worth discussing this pull request in #600.

Tatsh:
Version of this patch for version 1.17.1 (works on Gentoo with its user patches system): https://gist.github.com/Tatsh/9e527bcc760bee143e22df9dd380cead

DimitriPapadopoulos:
I think it would be best to strandardize on [vpnc-script](https://gitlab.com/openconnect/vpnc-scripts/), like [OpenConnect](https://gitlab.com/openconnect/openconnect).

Tatsh:
In terms of systemd stuff, it should be detected if the user is actually using it. It's not required by systemd to be working. If that check fails for some reason (false positive), a new setting like `systemd-dns` should be added so it can be set to `0`.

Lastly, rather than introducing a generic scripting thing here, the code could directly call the D-Bus API.

Main thing for me is to have the D-Bus calls do the DNS changing, and not touch `/etc/resolv.conf`.

Tatsh:
This does fix my issue where Chrome (and not any others) could not resolve based on changes made to `/etc/resolv.conf`. Many things read it and take the first nameservers, but not Chrome as of recently.

Now that those changes are made to the PPP link with systemd, Chrome can correctly resolve certain VPN-specific domains.

DimitriPapadopoulos:
I believe [vpnc-script](https://gitlab.com/openconnect/vpnc-scripts/) does the right thing. It would be great to remove the routing/DNS code written in C and call `vpnc-script` instead. Unfortunately, I don't have the resources for that change.

DimitriPapadopoulos:
Finally, I suppose Chrome is resolving using systemd. Other software resolve using glibc.

You can reproduce this using `systemd-resolve` (obviously uses systemd) and `nslookup` (uses glibc).

While systemd and glibc have been unified in recent Fedora distributions (https://fedoraproject.org/wiki/Changes/systemd-resolved), this might not be the case on Gentoo.

Tatsh:
They won't be unified on Gentoo, unless Gentoo decides that systemd will be the only supported init system. This is very unlikely to happen.

Tatsh:
macOS needs equivalent call to `networksetup -setdnsservers IF ...`, but I do not see vpnc-scripts supporting it at this time.

DimitriPapadopoulos:
Yes, there are open issues about that:
[openconnect/vpnc-scripts#15](https://gitlab.com/openconnect/vpnc-scripts/-/issues/15)
[openconnect/vpnc-scripts#27](https://gitlab.com/openconnect/vpnc-scripts/-/issues/27)

Yet I believe [vpnc-scripts](https://gitlab.com/openconnect/vpnc-scripts/) remain the best bet in the long term.

Tatsh:
I looked at some of the logic to see if it's feasible to use as-is with the current openfortivpn code (with a change similar to this PR). Right now the main issue is the `VPNGATEWAY` environment variable which then calls `set_vpngateway_route`, where this is already taken cared of by openfortivpn. Otherwise, the `INTERNAL_IP4_DNS` env var with the `connect` reason will set the correct DNS, while the `disconnect` reason will undo it.

So in summary:

* openfortivpn has to pass `VPNGATEWAY` (and others) for routing, and `INTERNAL_IP4_DNS` for DNS, via environment
* Make a code path to avoid doing routing/DNS in openfortivpn and instead pass it to `vpnc-script`. Perhaps an option like `--use-vpnc-script [PATH]` (default to searching for `vpnc-script` in `PATH`)
* Call the script at the appropriate times: `vpnc-script pre-init`, `vpnc-script connect`, `vpnc-script disconnect`

DimitriPapadopoulos:
Yes, most of [`ipv4.c`](https://github.com/adrienverge/openfortivpn/blob/master/src/ipv4.c) would be replaced by [`vpnc-script`](https://gitlab.com/openconnect/vpnc-scripts/-/blob/master/vpnc-script), or an alternative script such as [`vpnc-slice`](https://github.com/dlenski/vpn-slice).

The nice thing with that solution is that IPv6 support would (more) straightforward to add.

Tatsh:
@DimitriPapadopoulos this change is now in master?

DimitriPapadopoulos:
It shouldn't. Let me see.

DimitriPapadopoulos:
This pull request has been closed inadvertently by GitHub. Re-opened as #1092.

PR #617, opened 2020-04-02
do not use interface ip for routing on linux

mrbaseman:
we do this already on mac and bsd

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -890,12 +890,10 @@ static int ipv4_set_split_routes(struct tunnel *tunnel)
 		route_iface(route) = strdup(tunnel->ppp_iface);
 		if (!route_iface(route))
 			return ERR_IPV4_NO_MEM;
-#ifndef HAVE_RT_ENTRY_WITH_RT_DST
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
About the revision, may I suggest something similar to what I have been using in the _openfortivpn_ "snap"? It's the standard way in Snapcraft:
https://forum.snapcraft.io/t/getting-the-git-commit-tag-as-a-version/14716/7

```sh
version="$(git describe --tags | sed -e 's/-/+git/;y/-/./')"
```

If checking out a tag such as `v1.13.3` the version will be `1.13.3`. Otherwise it will be `1.13.3+git3.gb4c0a3d` where `3` is the number of commits since last tag `v1.13.3` and `gb4c0a3d` is the commit hash.

DimitriPapadopoulos:
Indeed #627 works for `master` and I have just verified it works on this branch too. As you say perhaps #627 needed to be merged into `master` for some reason. Anyway, I cannot really comment on this pull request, you know much better than me. I suggest you merge...

PR #619, opened 2020-04-03
Fix OpenSSL build

DimitriPapadopoulos:
The location of old releases has been moved from:
https://www.openssl.org/source/
to:
https://www.openssl.org/source/old/1.0.2/


DimitriPapadopoulos:
Travis seems to be stuck... Closing this pull request.

PR #622, opened 2020-04-04
Check malloc results to silence GCC 10 analyzer

jmakovicka:


mrbaseman:
Looks good. Thanks for this. Unfortunately, the result of the Travis-CI check was not reported back and I can not merge this directly. @DimitriPapadopoulos any idea how to fetch the "ok" from Travis manually?

DimitriPapadopoulos:
No clue. [Travis shows this job passed](https://travis-ci.org/github/adrienverge/openfortivpn/builds/670945969) but GitHub has no clue about it. This has happened to me recently and the only way I have found to work around it is to delete the pull request and submit a new one (see pull requests "Fix OpenSSL builds").

travis

DimitriPapadopoulos:
I think only @adrienverge may have the rights to force merge this pull request. See for example:

I understand these recent problems may be related to "Legacy integration" and "GitHub Apps should not exhibit this issue" but I don't know what that means exactly and how to migrate. Also I think only @adrienverge can modify Travis integration.

adrienverge:
I've seen many Travis strange status recently, not only on openfortivpn. As you point out, this may be related to "legacy integration", but should work theoretically.

Anyway, since the PR looks good to you, I'll merge.

mrbaseman:
Uh, my GCC 5.4.0 complains now:
```
src/ipv4.c: In function ‘ipv4_get_route’:
src/ipv4.c:195:6: warning: ‘fd’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  if (close(fd))
      ^
src/ipv4.c: In function ‘ipv4_protect_tunnel_route’:
src/ipv4.c:798:9: warning: ‘ret’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  return ret;
         ^
```

DimitriPapadopoulos:
Yes, Coverity complains as well.

CID 355472: Uninitialized scalar variable (UNINIT)
5. **uninit_use**: Using uninitialized value `ret`.

mrbaseman:
should be fixed by #626

PR #627, opened 2020-04-06
include git commit id as revision in debug output

mrbaseman:
We update the version number only when we create a new release. When testing different branches, it sometimes makes sense to have a more fine grained hint at which level of the sources a binary was built. The commit id of the git HEAD could be such a little piece that may help to distinguish if it's the released version or if it includes (at least one) more patch(es).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,7 +12,8 @@ openfortivpn_CFLAGS = -Wall -pedantic -std=gnu99
 openfortivpn_CPPFLAGS = -DSYSCONFDIR=\"$(sysconfdir)\" \
 			-DPPP_PATH=\"@PPP_PATH@\" \
 			-DNETSTAT_PATH=\"@NETSTAT_PATH@\" \
-			-DRESOLVCONF_PATH=\"@RESOLVCONF_PATH@\"
+			-DRESOLVCONF_PATH=\"@RESOLVCONF_PATH@\" \
+			-DREVISION=\"@REVISION@\"
 
 openfortivpn_CPPFLAGS += $(OPENSSL_CFLAGS) $(LIBSYSTEMD_CFLAGS)
 openfortivpn_LDADD = $(OPENSSL_LIBS) $(LIBSYSTEMD_LIBS)
--- a/configure.ac
+++ b/configure.ac
@@ -12,6 +12,14 @@ AC_GNU_SOURCE
 m4_ifndef([PKG_PROG_PKG_CONFIG], [m4_fatal([Please install pkg-config.])])
 PKG_PROG_PKG_CONFIG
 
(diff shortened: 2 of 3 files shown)


mrbaseman:
Now it works for me, too. Perhaps I haven't merged the tags into my fork. I don't know.

Shall we also show the revision in the output of `openfortivpn --version`? 
I tend to say "no", because we don't have verbosity levels in that code branch, and some user scripts may already depend on the output being just the version number and nothing else.

DimitriPapadopoulos:
No, I agree revision should be shown only between releases.

DimitriPapadopoulos:
Is this sufficient?
```
	log_debug("openfortivpn " VERSION "\n");
	if (REVISION != VERSION)
		log_debug("    revision " REVISION "\n");
```

mrbaseman:
Oh, there are several aspects here: 
- Now we show the revision in the debug output of `openfortivpn -v`. Your [comments](https://github.com/adrienverge/openfortivpn/pull/627#issuecomment-610288299) aims in the direction that if no commits have been submitted since the last tag, the revision is basically the version, but I think it's not exactly the same. It would look like:
```
$ ./openfortivpn -v
DEBUG:  openfortivpn 1.13.4
DEBUG:      revision v1.13.4
```
and we can't compare the preprocessor defines in C with `!=`.
- the revision string repeats the version. That might be good from what I have seen yesterday on my branch where `git --describe` has counted the commits since `v1.12.0` for some reason. If there are no commits since the last tag it's somewhat redundant. Perhaps we could attempt to suppress the output if `strlen()` reports less than let's say 10 characters.
- Independent from these considerations, my point [above](https://github.com/adrienverge/openfortivpn/pull/627#issuecomment-610266227) was that `./openfortivpn --version` does not show the revision. It only shows the version. 
Appending a `-v` does not help, even if we add a `log_debug("revision " REVISION "\n");` statement there, because the verbosity level is determined later. But  `./openfortivpn -v --version` would work.

I have implemented these ideas in #630

DimitriPapadopoulos:
Instead of  `if (strlen(REVISION) > 10)` which doesn't make the intent very clear and is making use of a "magic number", I would recommend either comparing `REVISION` and `VERSION`. My intent was to compare them at run-time - not at compile-time using the C preprocessor.

Something like this:
```
if (strcmp(&REVISION[1], VERSION))
```
or even:
```
if (strstrREVISION, "+git")
```

mrbaseman:
yes, good point.
```
if (strcmp(&REVISION[1], VERSION))
```
works well, and also shows `unavailable` when `git` is not available at configure time. That makes it at least clear that one can **not** be sure that there were no further commits. We simply don't know in that case (like it has been all the time so far).

PR #630, opened 2020-04-07
improve handling of revision string

mrbaseman:


mrbaseman:
I've found two more issues:
- if the .git directory is not there (e.g. source downloaded and copied manually, or extracted from a src.rpm), the `REVISION` string was empty and therefore `&REVISION[1]` outside of bounds 
- the result of `AC_PATH_PROG` for `git` was not used at all in the configure script

Issue #629, opened 2020-04-07
SmartCard + username and password


simowdas:
Version: 1.12.0
When using username,password and user-cert from smartcard the username and pass are not send to server.

/remote/login is send only cert=&nup=1

DimitriPapadopoulos:
Ins't this a duplicate of #507?

DimitriPapadopoulos:
Or perhaps not because #507 is supposed to have been fixed in 1.12.0.

DimitriPapadopoulos:
It would help if you could show the (sanitized) output of `openfortivpn -v -v` and the (sanitized) contents of `/etc/openfortivpn/config`.

simowdas:
`DEBUG:  openfortivpn 1.12.0                                                                                                                                                                     [66/801]
DEBUG:  Loaded config file "/etc/openfortivpn/config".                                                                                                                                                  
DEBUG:  Loaded password from config file "/etc/openfortivpn/config"                                                                                                                                     
DEBUG:  Config host = "***"                                                                                                                                                                             
DEBUG:  Config realm = "someone"                                                                                                                                                                        
DEBUG:  Config port = "443"                                                                                                                                                                             
DEBUG:  Config username = "someone"                                                                                                                                                                     
DEBUG:  Resolving gateway host ip                                                                                                                                                                       
DEBUG:  Establishing ssl connection                                                                                                                                                                     
DEBUG:  server_addr:*****                                                                                                                                                                               
DEBUG:  server_port: 443                                                                                                                                                                                
DEBUG:  gateway_addr: ******                                                                                                                                                                            
DEBUG:  gateway_port: 443                                                                                                                                                                               
DEBUG:  Setting cipher list to: HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4                                                                                                                                   
DEBUG:  Gateway certificate validation failed.                                                                                                                                                          
DEBUG:  Gateway certificate digest found in white list.                                                                                                                                                 
INFO:   Connected to gateway.                                                                                                                                                                           
DEBUG:  http_send:                                                                                                                                                                                      GET /remote/login HTTP/1.1                                                                                                                                                                              Host: ***********:443                                                                                                                                                                                   User-Agent: Mozilla/5.0 SV1                
Accept: text/plain                                
Accept-Encoding: identity                                                                           
Content-Type: application/x-www-form-urlencoded
Cookie: 
Content-Length: 11

cert=&nup=1
DEBUG:  http_receive:
[...]

PR #631, opened 2020-04-07
allow logging in with smartcard and username+password

mrbaseman:
This should perhaps be tested by someone who actively uses smartcard login without username and password. I remember that usual practice was to have dummy entries in username and password which is a configuration which would be broken by this commit. Anyhow, if those users change their config file to include an empty line for username and password, I would expect it to work again, but I would have a better feeling if someone could confirm this.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -605,8 +605,7 @@ int auth_log_in(struct tunnel *tunnel)
 
 	tunnel->cookie[0] = '\0';
 
-	if (tunnel->config->use_engine > 0
-	    || (username[0] == '\0' && tunnel->config->password[0] == '\0')) {
+	if (username[0] == '\0' && tunnel->config->password[0] == '\0') {
 		snprintf(data, sizeof(data), "cert=&nup=1");
 		ret = http_request(tunnel, "GET", "/remote/login",
 		                   data, &res, &response_size);


mrbaseman:
@simowdas [has confirmed](https://github.com/adrienverge/openfortivpn/issues/629#issuecomment-610466947) that this works for him and solves #629

mrbaseman:
we have the [feature request](https://github.com/adrienverge/openfortivpn/issues/629) for logging in with username+password+smartcard and we have a solution for this in  [my smartcard branch](https://github.com/mrbaseman/openfortivpn/tree/smartcard). I would like to have this change tested in other scenarios (smartcard without username+password or only regular user certificate, i.e. without username+password etc.). 
I went through the older issues and the pull requests in this area and have collected the users who have contributed or reported.
@rmuehl @kamushadenes @claudiuion @nicolaerosia @martinetd @sabino @epsilon77  could you help us and test this change in your environment, please?

If you have a dummy entry for the username in the config file and you log in without username+password, please put empty entries in the config file:
```
username =
password =
```

If you need a password, either put it in the config file or remove the `password` line from the config (or comment it out).

It would be great if you could provide your feedback to this. Thank you.

DimitriPapadopoulos:
@mrbaseman Not much feedback about this change. Perhaps we should merge and hope it is tested a little bit more in the master branch.

mrbaseman:
Ok. It's a pity that we didn't receive much feedback, even in those times of close down, when VPN solutions have gathered more importance.
As you have suggested, I have rebased, squashed and merged it, in the hope of getting some more testing and perhaps some feedback before the next release.

PR #634, opened 2020-04-09
Standardize Automake use: additional changes

DimitriPapadopoulos:
* Add missing target to "make" `etc/openfortivpn/config`.
* Add missing target to "make clean" `etc/openfortivpn/config`.
* Add missing target to "make uninstall" `$(confdir)/config`.


mrbaseman:
oh, I have noticed this, too, some time ago but forgot about it

PR #636, opened 2020-04-09
Add dns suffix information to informative message

charlesgoyard:
Hi,
I found out that printing the dns suffixes could be a first step to enable networknanager-fortisslvpn to scrape this information, and then report it back to whatever handles resolv.conf in NetworkManager. For now this information is not used by pppd.

On the other end, I think adding a stable parsable output would be a more appropriate, proper solution.

Sample output:
```
VPN account password:
INFO:   Connected to gateway.
Two-factor authentication token:
INFO:   Authenticated.
Using interface ppp0
Connect: ppp0 <--> /dev/pts/1
INFO:   Got addresses: [0.0.0.0], ns [0.0.0.0, 0.0.0.0], ns_suffix [sub.example.com example.com]
INFO:   negotiation complete
INFO:   negotiation complete
Cannot determine ethernet address for proxy ARP
local  IP address 0.0.0.0
remote IP address 0.0.0.0
INFO:   Interface ppp0 is UP.
INFO:   Setting new routes...
INFO:   Adding VPN nameservers...

```

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -473,7 +473,7 @@ static void *ssl_read(void *arg)
 
 		if (tunnel->state == STATE_CONNECTING) {
 			if (packet_is_ip_plus_dns(packet)) {
-				char line[ARRAY_SIZE("[xxx.xxx.xxx.xxx], ns [xxx.xxx.xxx.xxx, xxx.xxx.xxx.xxx]")];
+				char line[ARRAY_SIZE("[xxx.xxx.xxx.xxx], ns [xxx.xxx.xxx.xxx, xxx.xxx.xxx.xxx], ns_suffix []") + MAX_DOMAIN_LENGTH];
 
 				set_tunnel_ips(tunnel, packet);
 				strcpy(line, "[");
@@ -482,6 +482,8 @@ static void *ssl_read(void *arg)
 				strncat(line, inet_ntoa(tunnel->ipv4.ns1_addr), 15);
 				strcat(line, ", ");
 				strncat(line, inet_ntoa(tunnel->ipv4.ns2_addr), 15);
+				strcat(line, "], ns_suffix [");
+				strncat(line, tunnel->ipv4.dns_suffix, MAX_DOMAIN_LENGTH);
 				strcat(line, "]");
 				log_info("Got addresses: %s\n", line);
 			}
--- a/src/ipv4.h
+++ b/src/ipv4.h
@@ -64,6 +64,10 @@ struct rtentry {
 #define MAX_SPLIT_ROUTES 65535
 #define STEP_SPLIT_ROUTES 32
 
+// see https://unix.stackexchange.com/questions/245849
+// ... /resolv-conf-limited-to-six-domains-with-a-total-of-256-characters
+#define MAX_DOMAIN_LENGTH 256
+
 struct ipv4_config {
 	struct in_addr	ip_addr;
 
--- a/src/xml.c
+++ b/src/xml.c
@@ -16,6 +16,7 @@
  */
 
 #include "xml.h"
+#include "ipv4.h"
 #include "log.h"
 
 #include <string.h>
--- a/src/xml.h
+++ b/src/xml.h
@@ -18,10 +18,6 @@
 #ifndef OPENFORTIVPN_XML_H
 #define OPENFORTIVPN_XML_H
 
-#define MAX_DOMAIN_LENGTH 256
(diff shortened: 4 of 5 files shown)


DimitriPapadopoulos:
I understand NetworkManager-fortivpnssl does not get the name servers and search domains directly from _openfortivpn_, instead the information is passed to _pppd_ (hence the [`--pppd-use-peerdns=1` patch](https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn/-/merge_requests/15)) and retrieved by this _pppd_ plugin:
https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn/-/blob/master/src/nm-fortisslvpn-pppd-plugin.c

And indeed the **`usepeerdns`** option of _pppd_ handles only two DNS servers, no search domains:
> **usepeerdns**
>    Ask the peer for up to 2 DNS server addresses. The addresses supplied by the peer (if any) are passed to the /etc/ppp/ip-up script in the environment variables DNS1 and DNS2, and the environment variable USEPEERDNS will be set to 1. In addition, pppd will create an /var/run/ppp/resolv.conf file containing one or two nameserver lines with the address(es) supplied by the peer. 

Here you argue the information should be retrieved directly from _openfortivpn_.

Am I correct?

charlesgoyard:
We understand the same thing about pppd, and you are correct, I'm talking about enabling a direct conversation between the networkmanager plugin and openfortivpn.

My thinking is that networkmanager can handle the resolv.conf file by itself, without the help of pppd.
So in order to lift the pppd limitation, yes, the resolver configuration could be handled by networkmanager.

Also, this could be a first step to getting the networkmanager plugin handle the "Two-factor authentication token" prompt (which is another topic).

But all of this is a bit of a stretch, it requires a lot more work to have a nice integration between networkmanager and openfortivpn.

DimitriPapadopoulos:
It would help to document the desired interaction between _openfortivpn_ and _NetworkManager-fortivpnssl_.

For example we could add _openfortivpn_ plugins, the same way _pppd_ supports plugins. But frankly this looks like overkill. I don't think _NetworkManager_ needs _openfortivpn_ to execute specific code. Rather some kind of limited interprocess communication (IPC) is required.

On Linux [D-BUS](https://www.freedesktop.org/wiki/Software/dbus/) has become a standard for IPC between system components. Note that _openfortivpn_ already notifies [_systemd_](https://www.freedesktop.org/wiki/Software/systemd/), precisely after setting routes and DNS - or not:
https://github.com/adrienverge/openfortivpn/blob/8b0c235b63e47c86c83c3380f72a5b9f09b11f92/src/tunnel.c#L127

Perhaps it wouldn't be too complicated to get _openfortivpn_ to notify _NetrowkManager_ as well as _systemd_. Then it would be up to the calling program to manage routing and DNS (whether the calling program is _NetworkManager_ or a script bundled with _openfortivpn_). Thus _openfortivpn_ could focus on what it does best: initiate the communication with a FortiGate appliance and "spawn a pppd process and operate the communication between the gateway and this process".

charlesgoyard:
I totally agree with what you say, an proper, simple IPC channel is the way. Please reject my PR and let's start thinking about it.

Also, I found out that there is something started in this direction with how TFA is handled (--pinentry). At the moment I could not make it work with SMS, but maybe it's just me.

DimitriPapadopoulos:
@charlesgoyard That said printing the search domain makes sense, not only to help improving _NetworkManager-fortisslvpn_ but also for debugging and information.

A couple comments:
* I believe `MAX_DOMAIN_LENGTH` should be moved from [src/xml.h](https://github.com/adrienverge/openfortivpn/blob/master/src/xml.h) to [src/ipv4.h](https://github.com/adrienverge/openfortivpn/blob/master/src/ipv4.h) - why `#include "xml.h"` if we don't need to parse XML?
* In the long term we probably need something more readable than these `1 + 15 + 7 + 15 + 2 + 15 + 1 + 1 + 7 + MAX_DOMAIN_LENGTH(256) + 1` comments, but that I will fix that in a separate PR. It's the length of this line that breaks checks. Can you break the comment in two separate lines or a single line above the declaration?

charlesgoyard:
Thanks! Nice cleanups as a bonus.

DimitriPapadopoulos:
Note that callbacks for setting IP routes and name resolution might not be a bad idea after all. See:
* #600
* #615
* #650

PR #640, opened 2020-04-14
Reorganize http_receive()

DimitriPapadopoulos:
Prepare for adding an external HTTP library later on.

The pull request will help use external HTTP / XML libraries later on. At this point I am not certain we will benefit from using [PicoHTTPParser](https://github.com/h2o/picohttpparser) in terms of maintainability and less code. Yet I'm opening the pull request to perform some code clean up, add a few assertions here and there, and raise a few issues.

For example we do not catch SSL errors if the header has already been read. Here is the relevant part in the new code:
https://github.com/DimitriPapadopoulos/openfortivpn/blob/230b158/src/http.c#L177-L181
and in the old code:
https://github.com/adrienverge/openfortivpn/blob/8b0c235/src/http.c#L217-L221
@adrienverge Is this on purpose? Are we lenient on purpose? On my machine I have modified the code to be strict on SSL errors, whether the HTTP header has been read or not. I haven't seen any occurrence of SSL errors.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -22,6 +22,7 @@
 #include "userinput.h"
 #include "log.h"
 
+#include <assert.h>
 #include <ctype.h>
 #include <string.h>
 #include <stdarg.h>
@@ -30,7 +31,8 @@
 #include <unistd.h>
 #include <arpa/inet.h>
 
-#define BUFSZ 0x10000
+static const uint32_t HTTP_BUFFER_SIZE = 0x8000;
+
 
 /*
  * URL-encodes a string for HTTP requests.
@@ -58,6 +60,7 @@ static void url_encode(char *dest, const char *str)
 	*dest = '\0';
 }
 
+
 /*
  * Sends data to the HTTP server.
  *
@@ -68,13 +71,13 @@ static void url_encode(char *dest, const char *str)
 int http_send(struct tunnel *tunnel, const char *request, ...)
 {
 	va_list args;
-	char buffer[BUFSZ];
-	char logbuffer[BUFSZ];
+	char buffer[HTTP_BUFFER_SIZE];
+	char logbuffer[HTTP_BUFFER_SIZE];
 	int length;
 	int n = 0;
 
 	va_start(args, request);
-	length = vsnprintf(buffer, BUFSZ, request, args);
+	length = vsnprintf(buffer, HTTP_BUFFER_SIZE, request, args);
 	va_end(args);
 	strcpy(logbuffer, buffer);
 	if (loglevel <= OFV_LOG_DEBUG_DETAILS && tunnel->config->password[0] != '\0') {
@@ -96,7 +99,7 @@ int http_send(struct tunnel *tunnel, const char *request, ...)
 
 	if (length < 0)
 		return ERR_HTTP_INVALID;
-	else if (length >= BUFSZ)
+	else if (length >= HTTP_BUFFER_SIZE)
 		return ERR_HTTP_TOO_LONG;
 
 	log_debug_details("%s:\n%s\n", __func__, logbuffer);
@@ -112,6 +115,7 @@ int http_send(struct tunnel *tunnel, const char *request, ...)
 
 	return 1;
 }
+
 
 static const char *find_header(
         const char *res,
@@ -135,6 +139,7 @@ static const char *find_header(
 	return NULL;
 }
 
+
 /*
  * Receives data from the HTTP server.
  *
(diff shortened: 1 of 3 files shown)


adrienverge:
@DimitriPapadopoulos I don't know about SSL errors, I can't remember.

All I can tell is that I tried to avoid having external dependencies, hence the HTTP decoding inside our source.

DimitriPapadopoulos:
Oh, agreed about the dependencies. I'm after small libraries that can be included in the source code. 

[PicoHTTPParser](https://github.com/h2o/picohttpparser) is just two files `picohttpparser.c` and `picohttpparser.h`.

adrienverge:
> I'm after small libraries that can be included in the source code.

:+1:

DimitriPapadopoulos:
I have pushed the code that is stricter on SSL errors. I will keep testing it for a few days.

DimitriPapadopoulos:
OK, it appears we are lenient on SSL errors as soon as the header has been read because of change https://github.com/adrienverge/openfortivpn/commit/70dadda. Specifically the change from:
```
	if (n < 0) {
```
to:
```
	if (!header_size) {
```

The comment reads:
> So, let's just obey the Content-Length header. If it's missing (chunked
encoding or standards violation) then read until the EOF, but still
succeed if the header was complete. This way we're hopefully on the safe
side.

Yet I don't see fundamental differences in the way we process chunked encoding vs. `Content-Length`-based responses. If `Content-Length` is known, we stop as soon as we read the expected length:
https://github.com/adrienverge/openfortivpn/blob/8b0c235b63e47c86c83c3380f72a5b9f09b11f92/src/http.c#L204-L205
If on the other hand the response is chunked-encoded, we stop as soon as we detect the last chunk terminator:
https://github.com/adrienverge/openfortivpn/blob/8b0c235b63e47c86c83c3380f72a5b9f09b11f92/src/http.c#L198-L202
So what is this "'safe side" about? I have verified that checking strictly for SSL errors still works when parsing chunked-encoded responses.

The code that handles the chunked encoding has been introduced  in https://github.com/adrienverge/openfortivpn/commit/f1bfe16, a few days **after** https://github.com/adrienverge/openfortivpn/commit/70dadda.  My guess is that the "safe side" part had probably been left over because the author wasn't yet certain about the robustness of the chunked encoding code:
> FortiOS 5 talks chunked. We don't really follow; but let's assume that
the if the packet ends with a zero-sized chunk we're done.
> 
> Kind of a change that works in testing and breaks in production.

Yet I believe the chunked encoding code has proven quite astute and robust over the years. It believe i's time we parse HTTP following the rules - so that we can move to 3rd party HTTP parsing code if needed.

src/http.c
```diff
@@ -150,75 +155,87 @@ int http_receive(
...
 
-	buffer = malloc(res_size);
+	buffer = malloc(capacity);
 	if (buffer == NULL)
 		return ERR_HTTP_NO_MEM;
 
-	do {
-		n = safe_ssl_read(tunnel->ssl_handle,
-		                  (uint8_t *) buffer + bytes_read,
-		                  BUFSZ - 1 - bytes_read);
```
mrbaseman, line 168: Did you drop this "- 1" intentionally?

src/http.c
```diff
@@ -150,75 +155,87 @@ int http_receive(
...
 
-			if (bytes_read == BUFSZ - 1) {
-				log_warn("Response too big\n");
+		/* expand the buffer if necessary */
+		if (bytes_read == capacity) {
+			char *new_buffer;
+
+			assert(UINT32_MAX / capacity >= 2);
+			capacity *= 2;
+			new_buffer = realloc(buffer, capacity);
```
mrbaseman, line 232: ...maybe you have decided to read to the very end of the buffer because you reallocate it here?

src/http.c
```diff
@@ -150,75 +155,87 @@ int http_receive(
...
+			if (new_buffer == NULL) {
 				free(buffer);
-				return ERR_HTTP_SSL;
+				return ERR_HTTP_NO_MEM;
 			}
+			buffer = new_buffer;
 		}
-	} while (n >= 0);
-
```
mrbaseman, line 216: the fact that you drop this condition is because you handle ssl errors in the code above, right?

src/http.c
DimitriPapadopoulos: Indeed I don't understand why it should be needed - we don't add a terminal `\0` until later on - and when we do we first realloc anyway.

src/http.c
DimitriPapadopoulos: That's not related - I think.

The only difference is that we used to bail out with `HTTP_ERR_NO_MEM`. Now we increase the size of the buffer.

src/http.c
DimitriPapadopoulos: Absolutely, I have added a `return` inside the loop, triggered by `if (n < 0)`:
https://github.com/DimitriPapadopoulos/openfortivpn/blob/c2c467ed067372a0568ee7e7f1bf409034e2467f/src/http.c#L176
This is equivalent to continuing `while (n >= 0)` but in my view has the advantage of showing explicitly which part of the code is responsible for errors.

Note that I have added a test for `n == ERR_SSL_AGAIN` (⇔ `n == 0`) early in the loop:
https://github.com/DimitriPapadopoulos/openfortivpn/blob/c2c467ed067372a0568ee7e7f1bf409034e2467f/src/http.c#L172-L175
I believe it improves readability through proximity of conditional expressions and assignment of the variables `n` these conditionals operate on. Additionally this helps remove the `if (n> 0)` conditional in the loop since we now handle the `n < 0` and `n == 0` cases earlier in the loop.

DimitriPapadopoulos: The main change is that we now **return** from the function if `n < 0`.

We would previously only **leave the loop** if `n < 0`. Then immediately after the loop we would return only if `!header` - discarding possible SSL violations.

src/http.c
DimitriPapadopoulos: Besides I don't think I am now reading more than previously, unless of course we were hitting `ERR_HTTP_TOO_LONG` which is always a bug (see #575 for example). On the contrary I will read less and bail out in case of an SSL violation.

src/http.c
mrbaseman: maybe this was just a left-over from a previous version of the code that didn't realloc yet.

src/http.c
mrbaseman: I see.

src/http.c
mrbaseman: thank you for the explanations

src/http.c
DimitriPapadopoulos: Indeed. Not sure why we always realloc, even if the capacity is sufficient for the final `\0`. Sure, libc may recover a few bytes that might (perhaps) be re-used elsewhere but at the expense of copying the contents of `buffer`. Is it worth it?

DimitriPapadopoulos:
I'll try to provide more comprehensive explanations in the commit.

src/http.c
mrbaseman: perhaps just to make sure the final `'\0'` fits in - without having checked if it's necessary at all

mrbaseman:
maybe just a few buzzwords like "code cleanup and improve readability" would have lead my thoughts in the right direction :-)

src/http.c
DimitriPapadopoulos: Sure, but then that's what the "-1" I have removed was about - I think. Why have both? Strange.

mrbaseman: They have been there since the [initial commit](https://github.com/adrienverge/openfortivpn/blob/171cebae7db0c0dbd5d0609c2d4a52ed98a05182/src/http.c#L71-L81). Maybe the intention was really to shorten the buffer to the needed length.

PR #645, opened 2020-04-15
Better document buffer lengths

### LINKED ISSUE #644: Improve (?) comments about buffer length
For example I suggest changing from:
https://github.com/adrienverge/openfortivpn/blob/8b0c235b63e47c86c83c3380f72a5b9f09b11f92/src/io.c#L476
to:
```
	char line[ARRAY_SIZE("[xxx.xxx.xxx.xxx], ns [xxx.xxx.xxx.xxx, xxx.xxx.xxx.xxx]")];
```

DimitriPapadopoulos:
Fixes #644.


DimitriPapadopoulos:
Will help fix check failures in #636.

PR #662, opened 2020-04-21
Get Travis CI to build on macOS and FreeBSD

DimitriPapadopoulos:
Attempt to limit FreeBSD and macOS to specific branches.


DimitriPapadopoulos:
Giving up. The Travis CI - GitHub interface seems too fragile, I don't want to spend time on that.

PR #676, opened 2020-04-28
Attempt to warn of outdated FortiGate appliances

DimitriPapadopoulos:
FortiGate appliances that are unable to provide configuration
in XML format are probably running FortiOS 4 or earlier which
has been umaintained and vulnerable for some time now.

Warn the user that the FortiGate appliance is outadated and that
it might not be compatible with recent OpenSSL 1.1.1 libraries
with weak ciphers disabled - we've had a case on Ubuntu 20.04 for
example.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -870,6 +870,8 @@ int auth_get_config(struct tunnel *tunnel)
 	if (ret == 1)
 		return ret;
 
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
In that case the FortiGate appliance does not send nameservers in the HTTP configuration anyway. It's not really that it doesn't work, rather no information is sent.

Issue #682, opened 2020-05-02
ca md too weak


nirdothan:
Trying to connect with user cert and failing with the following error message:
`SSL_CTX_use_certificate_file: error:140AB18E:SSL routines:SSL_CTX_use_certificate:ca md too weak`
If I understand correctly, the rejection is coming from openssl, and the option is either to configure openssl with `tls-cipher=DEFAULT:@SECLEVEL=0` or to rebuild the certificate. I'm not sure how to do either one.
The certificate that I am using is originally a p12 file from which I've extracted the cert and key in PEM format.
running in ubuntu 20.04 LTS - 64-bit

DimitriPapadopoulos:
See #673 

You may be trying to connect to an obsolete FortiGate appliance This appliance seems to depend on weak algorithms that have been removed (by default or irremediably?) from the OpenSSL libraries bundled with Ubuntu 20.04. Have you tried the **`--insecure-ssl`** option?

By the way, is this PKCS #12 certificate a personal certificate you need to authenticate with the VPN appliance?

Would you happen to have some information on the model of FortiGate device and the version of FortiOS? If FortiOS is not too ancient you might find the information in the XML-formatted configuration sent by the FortiGate appliance and output by **`openfortivpn -v -v`**. It would help if you could show us the (sanitized) piece of information.

DimitriPapadopoulos:
Also please provide more information on the certificate (sanitize before posting):
```
openssl x509 -text -in <certificate_file.pem>
openssl x509 -nokeys -info -in <keystore.p12>

```

nirdothan:
Thanks @DimitriPapadopoulos 
An obsolete FortiGate appliance on the other end is quite possible. 
I'll try to find out version information from IT, hopefully by tomorrow.

`--insecure-ssl`, `--min-tls` and `--seclevel-1` didn't help

Here's the cert info that you asked for. Please let me know if I've sanitized it too far.

sanitized output of `openssl x509 -text -in <certificate_file.pem>`

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 4283 (0x10bb)
        Signature Algorithm: sha1WithRSAEncryption
        Issuer: REDACTED
        Validity
            Not Before: Jul 27 11:42:04 2017 GMT
            Not After : Jul 27 11:42:04 2037 GMT
        Subject: REDACTED
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
[...]

PR #685, opened 2020-05-03
Set the TLS security level early and on context

DimitriPapadopoulos:
The existing code does work for cipher negotiation with old (FortiOS 4?) FortiGate appliances, but not for personal certificates (SHA-1 certificates).
    
Two reasons I can see:
* SSL_set_cipher_list() was called **after** messing with certificates.
* SSL_set_cipher_list() applies only to the SSL connection, not to certificates, call SSL_CTX_set_cipher_list() on context instead.

This change addresses both of the above issues.

See [SSL(7)](https://www.openssl.org/docs/man1.1.1/man7/ssl.html#DATA-STRUCTURES) man page:
> **SSL_CTX** (SSL Context)
> This is the global context structure which is created by a server or client once per program life-time and which holds mainly default values for the **SSL** structures which are later created for the connections.
> 
> **SSL** (SSL Connection)
> This is the main SSL/TLS structure which is created by a server or client per established connection. This actually is the core structure in the SSL API. At run-time the application usually deals with this structure which has links to mostly all other structures.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -836,7 +836,7 @@ int ssl_connect(struct tunnel *tunnel)
 		return 1;
 	}
 
-	// Load the OS default CA files
+	/* Load the OS default CA files */
 	if (!SSL_CTX_set_default_verify_paths(tunnel->ssl_context))
 		log_error("Could not load OS OpenSSL files.\n");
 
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Unfortunately this does not help with #682. Will need to generate an SHA-1 certificate to investigate.

DimitriPapadopoulos:
@martinetd In order to support SHA-1 personal certificates (see #682), I have decided to call _[SSL_CTX_set_min_proto_version](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_min_proto_version.html)()_ and _[SSL_CTX_set_cipher_list](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_cipher_list.html)()_ on the **ctx** context instead of calling _[SSL_set_min_proto_version](https://www.openssl.org/docs/manmaster/man3/SSL_set_min_proto_version.html)()_ and _[SSL_set_cipher_list](https://www.openssl.org/docs/manmaster/man3/SSL_set_cipher_list.html)()_ on the **ssl** connection only. Does this sound OK to you?

DimitriPapadopoulos:
Thank you for looking into this. About TLSv1.3, see #687.

PR #693, opened 2020-05-07
SSL_OP_NO_SSLv3 is deprecated

DimitriPapadopoulos:
Option SSL_OP_NO_SSLv3 is deprecated as of  OpenSSL 1.1.0, use SSL_CTX_set_min_proto_version() instead.

[List of SSL OP Flags](https://wiki.openssl.org/index.php/List_of_SSL_OP_Flags#Protocol_Version_Options)

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -35,6 +35,7 @@
 
 #include <openssl/err.h>
 #include <openssl/engine.h>
+#include <openssl/ui.h>
 #include <openssl/x509v3.h>
 #if HAVE_SYSTEMD
 #include <systemd/sd-daemon.h>
@@ -850,21 +851,9 @@ int ssl_connect(struct tunnel *tunnel)
 		}
 	}
 
-	/* Modify default TLS security levels */
-	if (!tunnel->config->insecure_ssl) {
-		long sslctxopt = SSL_OP_NO_SSLv3 | SSL_OP_NO_COMPRESSION;
-		long checkopt;
-
-		checkopt = SSL_CTX_set_options(tunnel->ssl_context, sslctxopt);
-		if ((checkopt & sslctxopt) != sslctxopt) {
-			log_error("SSL_CTX_set_options didn't set opt: %s\n",
-			          ERR_error_string(ERR_peek_last_error(), NULL));
-			return 1;
-		}
-	}
-
-	if (!tunnel->config->insecure_ssl) {
-		if (!tunnel->config->cipher_list) {
+	/* Disable vulnerable TLS protocols and ciphers by default*/
+	if (!tunnel->config->cipher_list) {
+		if (!tunnel->config->insecure_ssl) {
 			const char *cipher_list;
 
 			if (tunnel->config->seclevel_1)
@@ -872,13 +861,7 @@ int ssl_connect(struct tunnel *tunnel)
 			else
 				cipher_list = "HIGH:!aNULL:!kRSA:!PSK:!SRP:!MD5:!RC4";
 			tunnel->config->cipher_list = strdup(cipher_list);
-		}
-	} else {
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L
-		if (tunnel->config->min_tls <= 0)
-			tunnel->config->min_tls = TLS1_VERSION;
-#endif
-		if (!tunnel->config->cipher_list && tunnel->config->seclevel_1) {
+		} else if (tunnel->config->seclevel_1) {
 			const char *cipher_list = "DEFAULT@SECLEVEL=1";
 
 			tunnel->config->cipher_list = strdup(cipher_list);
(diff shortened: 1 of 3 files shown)


DimitriPapadopoulos:
@martinetd Again I have modified some of your code, based on recent OpenSSL documentation. Could you please have a look?

Also should we add other flags as documented in [Python SSL and TLS security](https://python-security.readthedocs.io/ssl.html)? I don't know much about OpenSSL, perhaps these flags are relevant server-side only, but here is a list of flags that might be of interest at least for old OpenSSL libraries:
* **`SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS`**:
  Added in 0.9.6e to disable the fragment insertion that was added in 0.9.6d (where it was always enabled). Versions up to 0.9.6c do not include the countermeasure. 
* **`SSL_OP_SINGLE_DH_USE`**
  If set, always create a new key when using tmp_dh parameters. As of 1.0.2f single-DH key use is always on, and this option does nothing, and is retained for compatibility.
* **`SSL_OP_SINGLE_ECDH_USE `**
  If set, always create a new key when using tmp_ecdh parameters.

DimitriPapadopoulos:
It looks like compression is off by default from 1.1.0:
[SSL_CONF_cmd](https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_cmd.html#SUPPORTED-COMMAND-LINE-COMMANDS)
> **-no_comp**
>   Disables support for SSL/TLS compression, same as setting **SSL_OP_NO_COMPRESSION**. As of OpenSSL 1.1.0, compression is off by default.

Wouldn't it make sense to simply disable the code that sets options if `OPENSSL_VERSION_NUMBER >= 0x10100000L`?

martinetd:
Hmm right I got the version number wrong for default compression. I'd say it doesn't hurt either way, we can spare the call to `SSL_CTX_set_options` altogether for recent openssl ; but it's not like it hurts either, it's not deprecated yet.
Do as you like :)

Should we open a new issue for bumping min_tls up to TLS1.2?

DimitriPapadopoulos:
About TLS1.2: We only talk to FortiGate portals. I expect such portals to do the right thing and negotiate the most recent protocol available on both sides. Yet it it could be useful to make users painfully aware they are talking to obsolete gateways :-)

Do you happen to know when FortiOS started supporting TLS v1.2?

martinetd:
> About TLS1.2: We only talk to FortiGate portals. I expect such portals to do the right thing and negotiate the most recent protocol available on both sides. Yet it it could be useful to make users painfully aware they are talking to vulnerable portals :-)

Problem isn't so much that as downgrade attacks.. I just checked and FortiOS >= 6.0.6 properly set the min version on their side so it should be ok but better play safe there, it could be changed for specific servers to support older clients (I have to do this for openvpn servers I operate), so client-side min version makes sense.

> Do you happen to know when FortiOS started supporting TLS v1.2?

No idea... 5.2.9 relese notes mention it but somewhere unrelated to saying it's new; it could have been much older. I think it's old enough something not supporting v1.2 is bound to have a lot of other security issues...

DimitriPapadopoulos:
TLSv1.2 is clearly available on FortiOS 5.0, which is currently the [oldest version still documented on the Fortinet site](https://docs.fortinet.com/product/fortigate/5.0):

[FortiOS CLI Reference 5.0.9](https://help.fortinet.com/fgt/handbook/cli_html/index.html#page/FortiOS%25205.0%2520CLI/config_vpn.27.18.html)
> ### ssl settings
> 
> | Variable | Description | Default |
> | :-- | :-- | :-- |
> | `tlsv1-2 {enable \| disable}` | Enable or disable TLS 1.2 cryptographic protocol. | enable |

DimitriPapadopoulos:
Finally let's enforce a minimal TLS version with _SSL_CTX_set_options()_ too, not only when _SSL_CTX_set_min_proto_version()_ is available.

src/tunnel.c
```diff
@@ -851,34 +852,39 @@ int ssl_connect(struct tunnel *tunnel)
 	}
 
 	/* Modify default TLS security levels */
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+	if (tunnel->config->min_tls <= 0)
+		tunnel->config->min_tls = TLS1_2_VERSION;
```
martinetd, line 857: I would set `tunnel->config->min_tls` unconditionally if left to <= 0, so you don't need to check for <=0 below. This actually might belong in default config, in main.c?
And at least change the default in a separate commit for history (this might deserve something in CHANGELOG.md...)

src/tunnel.c
```diff
@@ -895,13 +878,42 @@ int ssl_connect(struct tunnel *tunnel)
 		}
 	}
 
+	if (!tunnel->config->insecure_ssl)
+		if (tunnel->config->min_tls <= 0)
+			tunnel->config->min_tls = TLS1_2_VERSION;
```
martinetd, line 883: (nitpick) you use TLS1_2_VERSION here inconditionally so checks in config.c might not make sense.
Do we still have any test with an openssl 0.9.8? The reaallly old 1.0.1 I could find already has this so it's fine as far as I would be willing to support explicitly, just wanted to point out the new ifdefs are redundant in config parsing code and below.

src/tunnel.c
```diff
@@ -895,13 +878,42 @@ int ssl_connect(struct tunnel *tunnel)
...
+			          ERR_error_string(ERR_peek_last_error(), NULL));
+			return 1;
+		}
+	}
+#else
+	if (!tunnel->config->insecure_ssl || tunnel->config->min_tls > 0) {
+		long sslctxopt = 0;
+		long checkopt;
+
+		if (!tunnel->config->insecure_ssl)
```
martinetd, line 900: we're already in such a if.

src/tunnel.c
```diff
@@ -895,13 +878,42 @@ int ssl_connect(struct tunnel *tunnel)
...
 		if (!SSL_CTX_set_min_proto_version(tunnel->ssl_context,
 		                                   tunnel->config->min_tls)) {
-			log_error("SSL_CTX_set_min_proto_version failed: %s\n",
+			log_error("Cannot set minimum protocol version (%s).\n",
+			          ERR_error_string(ERR_peek_last_error(), NULL));
+			return 1;
+		}
+	}
+#else
+	if (!tunnel->config->insecure_ssl || tunnel->config->min_tls > 0) {
```
martinetd, line 896: with insecure_ssl unset min_tls has to be >0 because we just set it

DimitriPapadopoulos:
Almost there I think.

Note that if you pass both `--min-tls` and `--insecure-ssl` the result might be strange and depend on the version of OpenSSL. Is it worth handling this inconsistency? I don't know how to handle incompatible options with `<getopt.h>`, I'd have to spend time reading about it.

src/tunnel.c
martinetd: (This comment is still valid)

DimitriPapadopoulos:
I might be wrong but I think your last comment precisely refers to the `insecure-ssl` / `min_tls` inconsistency. It is "legal" to have both `insecure-ssl` and `min_tls`, for example 1.2. In that case we do not set `SSL_OP_NO_SSLv3` (we want insecure SSL) but we do set `SSL_OP_NO_TLSv1` and `SSL_OP_NO_TLSv1_1`, which means we might disable TLSv1.1 but keep SSLv3!

martinetd:
Ah, I had read that 'or' as an 'and'.. because then if you have min_tls = tls1.2 and insecure ssl you would allow sslv3? That's weird :D 

But ok I don't mind much either way, not going to advise anyone to ever use the insecure-ssl option anyway...

DimitriPapadopoulos:
Yes, that's pretty weird! That's what I had in mind in https://github.com/adrienverge/openfortivpn/pull/693#issuecomment-625206607. Yet there is some logic in it: if user asks for insecure SSL and a minimal version of TLS at the same, so be it!
I could catch the inconsistency early on, while parsing options. That's a part of the code I don't know well, that would mean too much time best spent elsewhere, probably not worth the pain.

Issue #698, opened 2020-05-08
Build on FreeBSD broken


mrbaseman:
probably because of #681 but I haven't investigated in detail yet

mrbaseman:
```
root@freebsd:~/openfortivpn # ./autogen.sh 
+ type autoconf
+ type automake
+ type aclocal
+ aclocal
+ autoconf
+ automake --add-missing
+ echo 'now you can run ./configure && make to build openfortivpn'
now you can run ./configure && make to build openfortivpn
root@freebsd:~/openfortivpn # ./configure
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... ./install-sh -c -d
checking for gawk... no
checking for mawk... no
checking for nawk... nawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking for gcc... no
checking for cc... cc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether cc accepts -g... yes
checking for cc option to accept ISO C89... none needed
checking whether cc understands -c and -o together... yes
checking whether make supports the include directive... yes (GNU style)
checking dependency style of cc... gcc3
checking for cc option to accept ISO C99... none needed
checking for cc option to accept ISO Standard C... (cached) none needed
checking for a sed that does not truncate output... /usr/bin/sed
checking how to run the C preprocessor... cc -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking minix/config.h usability... no
checking minix/config.h presence... no
checking for minix/config.h... no
checking whether it is safe to define __EXTENSIONS__... yes
checking for pkg-config... /usr/local/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for git... no
checking whether make supports nested variables... (cached) yes
checking for OPENSSL... yes
checking for pthread_create in -lpthread... yes
checking for forkpty in -lutil... yes
[...]

PR #699, opened 2020-05-08
Revert "Use C99 standard headers"

mrbaseman:
Fix for #698
This reverts commit 77ec854eb601a1f691ab09b770158ff728aedc20.

### DIFF
--- a/src/ipv4.h
+++ b/src/ipv4.h
@@ -21,6 +21,8 @@
 #include <sys/socket.h>
 #ifdef HAVE_SYS_MUTEX_H
 /* Mac OS X and BSD wants this explicit include */
+#include <sys/param.h>
+#include <sys/lock.h>
 #include <sys/mutex.h>
 #endif
 #include <netinet/in.h>


DimitriPapadopoulos:
Instead of completely  reverting, could we perhaps add the missing headers on FreeBSD? Let me have a look at it.

DimitriPapadopoulos:
The problem is with include headers on many Unix flavors. They may not be self-sufficient, they need other headers to be included first. This is sometimes "by design" (if we can call that design). Actually POSIX is probably responsible for this "design" style. For example the [MUTEX(9)](https://www.freebsd.org/cgi/man.cgi?query=mutex&sektion=9&manpath=FreeBSD+12.1-RELEASE+and+Ports) can be interpreted to mean that `<sys/mutex;h>` requires other headers to be included first.
  ```
  #include <sys/param.h>
  #include <sys/lock.h>
  #include <sys/mutex.h>
  ```
It just happened to work until now, it doesn't mean we were including the proper headers. Let me review these includes.

DimitriPapadopoulos:
Header [src/ipv4.h](https://github.com/adrienverge/openfortivpn/blob/master/src/ipv4.h) requires:

---

#### `struct rtentry`

Does not seem to be documented in any standard. Appears to require **`<net/route.h>`** on Linux.

On FreeBSD, see [RTENTRY(9)](https://www.freebsd.org/cgi/man.cgi?query=rtentry&sektion=9):
> #### SYNOPSIS
> ```
>      #include <sys/types.h>
>      #include <sys/socket.h>
>      #include <net/route.h>
> ```

---

#### `struct sockaddr`

[POSIX.1-2017](https://pubs.opengroup.org/onlinepubs/9699919799/) defines ``struct sockaddr`` in [`<sys/socket.h>`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_socket.h.html).

On FreeBSD, see [SOCKET(9)](https://www.freebsd.org/cgi/man.cgi?query=socket&sektion=9):
> #### SYNOPSIS
> ```
>      #include <sys/socket.h>
>      #include <sys/socketvar.h>
> ```

---

#### `struct in_addr`

[POSIX.1-2017](https://pubs.opengroup.org/onlinepubs/9699919799/) defines ``struct in_addr`` in [`<netinet/in.h>`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netinet_in.h.html).

On FreeBSD, see [INET(3)](https://www.freebsd.org/cgi/man.cgi?query=inet&sektion3):
> #### SYNOPSIS
> ```
>      #include <sys/types.h>
>      #include <sys/socket.h>
>      #include <netinet/in.h>
>      #include <arpa/inet.h>
> ```

---

#### `struct sockaddr_in`

Does not seem to be documented in any standard. Appears to require **`<netinet/in.h>`** on Linux and in most cases.

On FreeBSD, see [IP(4)](https://www.freebsd.org/cgi/man.cgi?query=ip&sektion4):
> #### SYNOPSIS
> ```
>      #include <sys/types.h>
>      #include <sys/socket.h>
>      #include <netinet/in.h>
> ```

DimitriPapadopoulos:
This pull request does much more than simply reverting 77ec854eb601a1f691ab09b770158ff728aedc20 and breaks other more recent changes. I have started reviewing the included headers and in #700 I have reverted one significant change from 77ec854eb601a1f691ab09b770158ff728aedc20 and have put back in `ipv4.h`:
```
#include <sys/types.h>
```
I believe these aren't really need by `ip4.h` (perhaps by `ip4.c`?) but I have left them for now:
```
#ifdef HAVE_SYS_MUTEX_H
/* Mac OS X and BSD wants this explicit include */
#include <sys/mutex.h>
#endif
```
```
#include <net/if.h>
#ifdef HAVE_NET_IF_IF_VAR_H
/* on FreeBSD we need this additional header */
#include <net/if_var.h>
#endif
```

I currently have this problem that VirtualBox doesn't work anymore after upgrading my workstation to Ubuntu 20.04. Virtual machines get in serious trouble as soon as they start:
VirtualBox
I don't have access to a Mac OS X machine either since the confinement started. So I have been unable to test on FreeBSD and Mac OS X lately. Could you help me and give the current `master` branch a try and report any remaining issues?

DimitriPapadopoulos:
Found the problem with VirtualBox:
https://askubuntu.com/questions/1160514/vms-wont-run-on-virtualbox-guru-meditation-error-79
Hopefully I sall be able to test on FreeBSD again soon!

mrbaseman:
Me too, I was fighting with the masquerading configuration on the host where my freebsd installation is running under kvm... 

If this pull request was doing more than reverting 77ec854 I'm sorry. This was not intended. Reverting the commit directly was not possible, so I had to resolve a few conflicts manually (which looked like really minor issues).

@DimitriPapadopoulos your proposal in #700 has solved this issue.

PR #703, opened 2020-05-08
These headers are not needed, even on FreeBSD

DimitriPapadopoulos:
`<libutil.h>`   (and not `"lubutil.h"`) defines `pidfile_*` functions

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,6 +63,7 @@ net/if.h \
 netinet/in.h \
 netinet/tcp.h \
 pthread.h \
+strings.h \
 sys/ioctl.h \
 syslog.h \
 sys/socket.h \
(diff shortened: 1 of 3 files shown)


mrbaseman:
sorry, but now I get:```
src/tunnel.c:183:8: warning: implicit declaration of function 'forkpty' is
      invalid in C99 [-Wimplicit-function-declaration]
        pid = forkpty(&amaster, NULL, &termp, NULL);
              ^
1 warning generated.
```
not sure if it was caused by this commit or one before, at least the synopsis has to do something with libutil.h
https://www.freebsd.org/cgi/man.cgi?query=forkpty

sorry again for the mess this afternoon with my PR

DimitriPapadopoulos:
This one is really strange. I think I have all the headers. but it might very well depend on the other of inclusion. That's the BSD ""design" I was complnniing about. I'll get this fixed soon.

PR #708, opened 2020-05-12
Add command line option to bind to a specific network interface

aaronjg:
On multiple interface devices, it can be necessary to bind the VPN to a specific network interface. This can be necessary in situations where you have multiple interfaces and need to connect the VPN on a secondary interface. This feature is supported in openconnect, and would be useful in openfortivpn.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -244,6 +244,7 @@ AC_ARG_WITH([rt_dst],
 )
 AS_IF([test "x$with_rt_dst" = "x"], [
 AC_MSG_CHECKING(whether rtentry is available and has rt_dst)
+AC_LANG(C)
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([
 #include <sys/types.h>
 #include <sys/socket.h>
@@ -470,6 +471,23 @@ AS_IF([test "x$with_systemdsystemunitdir" = "xyes" -o "x$with_systemdsystemunitd
 AS_IF([test "x$with_systemdsystemunitdir" != "xno"],
       [AC_SUBST([systemdsystemunitdir], [$with_systemdsystemunitdir])])
 AM_CONDITIONAL([HAVE_SYSTEMD], [test "x$with_systemdsystemunitdir" != "xno"])
+
+AC_COMPILE_IFELSE([AC_LANG_SOURCE([
+#include <sys/types.h>
+#include <sys/socket.h>
+int main(int argc, char **argv){
+        int ret, handle;
+        handle = socket(AF_INET, SOCK_STREAM, 0);
+        ret = setsockopt(handle, SOL_SOCKET, SO_BINDTODEVICE,"lo",3);
+}
+])], [
+	AC_DEFINE(HAVE_SO_BINDTODEVICE, 1)
+	AC_MSG_NOTICE([HAVE_SO_BINDTODEVICE... 1])
+],[
+	AC_DEFINE(HAVE_SO_BINDTODEVICE, 0)
+	AC_MSG_NOTICE([HAVE_SO_BINDTODEVICE... 0])
+])
+
 
 # prepare to get rid of obsolete code (FortiOS 4)
 AC_ARG_ENABLE([obsolete],
--- a/doc/openfortivpn.1.in
+++ b/doc/openfortivpn.1.in
@@ -13,6 +13,7 @@ openfortivpn \- Client for PPP+SSL VPN tunnel services
 [\fB\-\-otp\-delay=\fI<delay>\fR]
 [\fB\-\-no\-ftm\-push\fR]
 [\fB\-\-realm=\fI<realm>\fR]
+[\fB\-\-ifname=\fI<interface>\fR]
 [\fB\-\-set\-routes=<bool>\fR]
 [\fB\-\-no\-routes\fR]
 [\fB\-\-set\-dns=<bool>\fR]
@@ -89,6 +90,9 @@ authentication based on OTP will be used instead.
 \fB\-\-realm=\fI<realm>\fR
 Connect to the specified authentication realm. Defaults to empty, which
 is usually what you want.
+.TP
+\fB\-\-ifname=\fI<interface>\fR
+Bind the connection to the specified network interface.
 .TP
 \fB\-\-set\-routes=\fI<bool>\fR, \fB\-\-no-routes\fR
 Set if openfortivpn should try to configure IP routes through the VPN when
--- a/src/config.c
+++ b/src/config.c
@@ -51,6 +51,7 @@ const struct vpn_config invalid_cfg = {
 	.no_ftm_push = -1,
 	.pinentry = NULL,
 	.realm = {'\0'},
+	.iface_name = {'\0'},
 	.set_routes = -1,
 	.set_dns = -1,
 	.pppd_use_peerdns = -1,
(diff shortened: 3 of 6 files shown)


DimitriPapadopoulos:
We already have an option `--pppd-ifname`, perhaps you should change the new option to `--ifname` for the sake of consistency.

DimitriPapadopoulos:
@aaronjg I changed a few details, can you have a look?

aaronjg:
Looks good to me.

DimitriPapadopoulos:
@mrbaseman The details of the patch look clean to me now, but perhaps you can comment on teh functionality?

The commits will need to be squashed before merging the pull request.

aaronjg:
I have been using it for a couple weeks, and has been working well for me on two interfaces. However, it would be great if someone could test on MacOS. 

A simple test is: 
1) Set up a device with two interfaces (e.g. wired and wireless). 
2) Run openfortivpn without the new argument and figure out which interface it is on (such as by disconnecting each interface and figuring out what goes down, usually will have been connected on wired rather than wireless).
3) Run openfortivpn with the new argument and tell it to bind to the secondary interface
4) Disconnect the primary interface
5) Confirm that you are still connected to the VPN

mrbaseman:
Sorry that I didn't find the time to review this pull request yet.
these changes look good to me so far. However, I haven't fully understood in which situations this is needed. Actually, in this part of the code we rely on the routing of the OS as it is configured before the connect. So, in order to find out the correct interface a lookup of the route to the vpn gateway should result in an entry with the correct interface, and opening the socket should bind to that interface, or am I missing something?
Later on in the code, we determine the route to the gateway and add a host route so that the communication through the tunnel is still possible, (nearly) no matter which routes are pushed by the fortigate. But if we miss something when opening the socket, we might also do something wrong in the later stage, are we?

aaronjg:
Fair question - this is needed if you can't rely on the routing of the OS to pick the correct interface. For example if you have two WANs and you want to run a forticlient on each of them to do failover between them. 

As far as I can tell, everything in the code is working fine when you want to use the interface that would be determined by the OS. This is to address situations when you want to explicitly specify the interface rather than letting the OS decide.

I wrote an openfortivpn package for OpenWRT and included this patch. I've been using it for a few months and haven't had any issues.

DimitriPapadopoulos:
@mrbaseman I'm all for merging this one, except it hasn't been tested on macOS or FreeBSD. I cannot test on macOS but perhaps you can test on FreeBSD?

mrbaseman:
@aaronjg Thank you for the explanation and also for your work packaging openfortivpn for OpenWRT

@DimitriPapadopoulos I hope that I find the time for testing next week. Until tomorrow we have a training, so not much time to do do other things in parallel

mrbaseman:
Hi,

unfortunately, on FreeBSD `SO_BINDTODEVICE` is not supported
```
src/tunnel.c:539:40: error: use of undeclared identifier 'SO_BINDTODEVICE'
                ret = setsockopt(handle, SOL_SOCKET, SO_BINDTODEVICE,
                                                     ^
1 error generated.
*** Error code 1
```
I have found this thread about that topic: 
https://busybox.busybox.narkive.com/PoCYU3I6/so-bindtodevice-is-not-supported-on-this-system 

the bottom line: obtain the ip address of the interface and bind to that IP.

aaronjg:
Thanks for testing. It's unfortunate that it doesn't work as is on FreeBSD.

On Linux, I have found binding to the IP address to be less reliable. There are situations where you can bind to an IP address and then some firewall or routing rule will end up sending the packet out of the incorrect device.

How do you want to proceed? Maybe we can find an option with compile time flags to make this enabled on Linux only? Or enable binding to either an IP Address or device? Or we can just give up and leave this out of the main OpenFortivpn distribution. The OpenWRT package of openfortivpn has this patch applied already, so for me it is okay if it is not in the upstream release...

Let me know. I'm happy to help, but unfortunately can't test on FreeBSD.

mrbaseman:
Maybe we can come up with an autoconf test e.g. using AC_TRY_COMPILE to detect if SO_BINDTODEVICE is available, and if not fallback to a variant of the code that tries to bind to the ip?

mrbaseman:
I've added an autoconf macro and tried to bind to the IP address on FreeBSD, and it compiles, but fails to obtain the IP address. 
We have two options: 
* continue working on the FreeBSD implementation and get the binding working
* use the existing autoconf macro to disable this feature completely in FreeBSD

aaronjg:
Is the problem just finding the ip address from the device? If so you could leave the code as is for linux (which should work on either a device name or ip address), use your freebsd code to bind when the user passes an IP address, and give an error if they try to bind to something that is not an ip address.

mrbaseman:
Sorry, forget it. It's much simpler: I had a typo in the interface name and therefore the `ioctl` command could not obtain an IP address. Actually, this was a perfect test if the error handling works properly ;-)

I have just tested it successfully on FreeBSD. @DimitriPapadopoulos could you test it on Mac OSX, please? My main concern is that the [autoconf macro](https://github.com/adrienverge/openfortivpn/pull/708/commits/7f4c37d8d0a352632cb6e454bc6be4b3b3337f6e#diff-67e997bcfdac55191033d57a16d1408aR475-R491) might give the wrong choice there.

And finally we need to squash and rebase the whole work.

mrbaseman:
I have rebased and squashed the commits in [my branch](https://github.com/mrbaseman/openfortivpn/tree/feature/iface-bind). Shall I force-push it here?

DimitriPapadopoulos:
I can try to build on macOS, probably not before the end of August though.
I cannot run and test on macOS though.

DimitriPapadopoulos:
Oh, I was able to find a working Apple machine to connect to remotely.

Note this _autoconf_error:
```
configure.ac:488: warning: AC_LANG_CONFTEST: no AC_LANG_SOURCE call detected in body
../../lib/autoconf/lang.m4:197: AC_LANG_CONFTEST is expanded from...
../../lib/autoconf/general.m4:2598: _AC_COMPILE_IFELSE is expanded from...
../../lib/autoconf/general.m4:2608: AC_COMPILE_IFELSE is expanded from...
configure.ac:488: the top level
```
This is because `AC_COMPILE_IFELSE` requires **`AC_LANG_PROGRAM`**. See [Autotools Mythbuster - 1. Changes in autoconf releases](https://autotools.io/forwardporting/autoconf.html#idm46482193718080).

DimitriPapadopoulos:
OK, the build then proceeds properly on macOS 10.11.6:
```
Package systemd was not found in the pkg-config search path.
Perhaps you should add the directory containing `systemd.pc'
to the PKG_CONFIG_PATH environment variable
No package 'systemd' found
configure: HAVE_SO_BINDTODEVICE... 0
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: executing depfiles commands
config.status: executing timestamp commands
```

DimitriPapadopoulos:
The resulting Makefile contains **` -DHAVE_SO_BINDTODEVICE=0`** as expected and _openfortivpn_ builds without problem. Unfortunately I won't be able to test it.

mrbaseman:
I have fixed the autoconf warnings. If it builds fine on OS X I think we can merge the code, because the new code is only executed when the new feature is actually used, so I'm confident that the change doesn't break any working use cases.

I have already rebased and squashed this as https://github.com/mrbaseman/openfortivpn/commit/fbef00b8d24bca88d018f7a9aadd99c9422a7a48 - shall I force-push it here?

nehaljwani:
I've tested this branch on macOS 10.15.6, didn't come across any issues.

aaronjg:
Great! Thanks for taking this the last mile @mrbaseman. Once this makes it into a release, I will update the openwrt package and remove the patch there.

PR #720, opened 2020-05-19
Fix password condition check

ikreb7:
Hello,

we recommend our users to let the password attribute at the config file empty and use the request `VPN account password:` if they start a connection. 

This worked only till `1.10`. This problem is fixed with this PR.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -578,7 +578,7 @@ int main(int argc, char **argv)
 		goto user_error;
 	}
 	// If username but no password given, interactively ask user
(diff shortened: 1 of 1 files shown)


mrbaseman:
Just remove the line for the password or comment it out and you are back to the previous behavior. 

An active entry in the config file with an empty password means that no password shall be sent now. This change was intentional to allow logging in with a certificate only or a smartcard and without password. 

The implementation wasn't so lucky in the retroperspective, I must admit.

Issue #678, opened 2020-04-28
Move routing and name resolution changes into an external script?


DimitriPapadopoulos:
I'm wondering whether moving the code that changes routing and name resolution parameters to an external script would help write and maintain _openfortivpn_. Typically most of `ipv4.h`/`ipv4.c` would move to that script. Additionally an option should be provided to run an alternative script.

Cons:
* Spawning an external process instead of handling that internally in the C code is bad - or so I thought.
* Providing an option to run an external script to be run with administrative rights is a security hole. This risk needs to be discussed and mitigated.

Pros:
* A script language is probably more suited to changing routing or name resolution parameters in a portable way.
* Relevant code would probably shrink at 10% of the initial C code and would be easier to maintain.
* Would help implement and maintain alternative methods of handling the name resolution subsystems (_NetworkManager_, _systemd-resolved_, _resolvconf_) as envisioned in #600.
* Linux packagers have a chance to easily adapt the name resolution stuff to their specific distribution.
* We already have a PR that suggests we add _systemd-resolved_ support as en external script: #615.
* OpenVPN have ended up doing the same, see [OpenVPN - ArchWiki](https://wiki.archlinux.org/index.php/OpenVPN#DNS).

adrienverge:
Hey Dimitri, thanks for sharing the idea!
I have taken too much distance to give valuable feedback, all I can tell is that running external shell scripts from a sudo program has always scared me. However you seem to say that it would allow running openfortivpn without root privileges, which I see as a very good move.
You and Martin have proven to master the code of this repo, so I'll trust the choice you make!

DimitriPapadopoulos:
Indeed external helper programs with SUID set may help limit the part of the code that runs as root. I think that's what _FortiClient_ does. Of course external helper programs don't have to be scripts. Like you, I feel uneasy about running SUID scripts (they actually require a C wrapper). However they're  really much easier to write, maintain and adapt to each Linux distribution.

It would also help with macOS: the DNS part is currently broken because changes to `/etc/resolv.conf` are not taken into account by recent macOS releases. Instead we have to call a specific Objective-C API - or run a command from a script.

The real obstacle is spawning `pppd` with option `noauth`. Do you know why that option requires root privileges and why it is required? Is there no way around it?

adrienverge:
I'm sorry, I have no idea...

zez3:
[...]

Issue #468, opened 2019-09-11
Add an option to trust certificate


woprandi:
Does it possible to have an option to trust interactively or automatically a certificate ? It's annoying to have to grep logs to find the sha256 certificate used and modify the configuration file.

DimitriPapadopoulos:
It looks like the certification authority (CA)  of the FortiGate appliance certificate is unknown to your computer's system (specifically _OpenSSL_). The whole point of certificates is that they are verified against trusted CAs. If the CA associated to the FortiGate appliance certificate is not trusted by the system, perhaps your computer has not been set up according to the expectations of the administrators of the FortiGate appliance. In that case you have to tell _openfortivpn_ to trust the certificate of the FortiGate explicitly. However you only need to do that once in the configuration file.

DimitriPapadopoulos:
@woprandi Does this answer your question - although it might not be the answer you want? Can we close this issue? Please feel free to discuss this issue.

DimitriPapadopoulos:
It might be more helpful to have an option (or even a separate script) to grep the logs and modify the configuration file.

woprandi:
Sorry, I'm not an expert but I will give you my opinion :
If you are an end-user without any technical knowledge and use this plugin
thought the Gnome plugin it's a blocking issue and you won't be able to
connect without help (for the first time and each time the certificate
changes). That's a problem.

DimitriPapadopoulos:
But then why use certificates in the first place?

mrbaseman:
I'd say the best solution to this problem is to get a server certificate from a CA that is commonly trusted by the operating system. The admin can export a signing request from the Fortigate and import the enrolled certificate. 
I think it's also possible to create the CSR and private key externally, pack it together with the certificate and import both into the Fortigate. If not using a  commonly trusted CA but an own one run in the company, the CA certfificate should be properly installed on the client systems (not only because _openfortivpn_ complains, but also other pieces of software that have to deal with encrypted connections).

woprandi:
> But then why use certificates in the first place? This really is a security issue. While I agree this might be a problem for some users, it doesn't mean I have to agree with the proposed solution.
> 
[...]

PR #727, opened 2020-05-29
Proposal: add trust-all-certs option.

a-andreyev:
Hello! Thank you for the project! :)

The proposal is to add an option to trust any incoming cert, what do you think?

It could be convenient for the connman and NetworkManager plugin developers not to crash the process if the incoming cert hash is not received on the client side yet.

The better approach is probably to implement the full-fledged async control interface to accept the incoming certificate digest, but I'm not sure what IPC protocol to choose then (sockets/dbus/etc).

Additional context is the connman plugin development for the mer/sailfish/aurora mobile linux distributions: https://git.sailfishos.org/mer-core/openfortivpn/merge_requests/2

Feel free to reject/criticize, would be happy to get feedback.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -73,6 +73,7 @@ const struct vpn_config invalid_cfg = {
 	.user_cert = NULL,
 	.user_key = NULL,
 	.insecure_ssl = -1,
+	.trust_all_certs = -1,
 	.cipher_list = NULL,
 	.min_tls = -1,
 	.seclevel_1 = -1,
@@ -402,6 +403,15 @@ int load_config(struct vpn_config *cfg, const char *filename)
 				continue;
 			}
 			cfg->insecure_ssl = insecure_ssl;
+		} else if (strcmp(key, "trust-all-certs") == 0) {
+			int trust_all_certs = strtob(val);
+
+			if (trust_all_certs < 0) {
+				log_warn("Bad trust-all-certs in config file: \"%s\".\n",
+				         val);
+				continue;
+			}
+			cfg->trust_all_certs = trust_all_certs;
 		} else if (strcmp(key, "cipher-list") == 0) {
 			free(cfg->cipher_list);
 			cfg->cipher_list = strdup(val);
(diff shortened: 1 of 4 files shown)


DimitriPapadopoulos:
I think it's a bad idea. In any case it has been already discussed in #468 for example.

That said I'm not sure I fully understand the use case, for example what do you mean by "the incoming cert hash is not received on the client side yet"? How does the NetworkManager-fortivpnssl process crash?

a-andreyev:
Thank you for the fast response and the previous discussion link!

You're probably right. I should take a look here   https://github.com/adrienverge/openfortivpn/issues/468#issuecomment-533963490 and https://github.com/adrienverge/openfortivpn/issues/468#issuecomment-534399623 and probably update my approach or close as a duplicate.

My initial motivation was that there's a logic in openfortivpn to get incoming sha256 cert digest. I thought it would be convenient to have an option to somehow receive it. The received digest could be used on the network manager / GUI side then. But I'm not sure is it a good security/convenience balance.

My bad in the wording also about the NM crash (sorry, not a native speaker). Not crash, but expected behavior, when the incoming cert digest is shown in the logs and the process finished.

DimitriPapadopoulos:
OK, I see:
* Indeed _openfortivpn_ receives the public certificate of the VPN gateway - and verifies it against the CA certificates it has in store.
* Indeed _openfortivpn_ will exit if the VPN gateway certificate cannot be verified. This is how TLS connections are supposed to operate.
* While NetworkManager-fortisslvpn could get the certificate from _openfortivpn_, it is (currently) not really the purpose of _openfortivpn_ to interoperate very broadly with other programs such as NetworkManager-fortisslvpn. Instead I think it makes sense for NetworkManager-fortisslvpn to download the certificate directly from the VPN gateway on the first attempt to connect to a VPN gatewey, **before** calling _openfortivpn_. It shouldn't be too complicated, see [SSL/TLS Client](https://wiki.openssl.org/index.php/SSL/TLS_Client). So on the **first** connexion to a VPN gatweay, NetworkManager-fortisslvpn could verify the certificate itself and, if verification fails, ask the user if it's OK to trust the VPN gateway despite the broken certificate.
* Finally I'm all for allowing better communication between _openfortivpn_ and other software such as NetworkManager-fortisslvpn using some IPC mechanism - see https://github.com/adrienverge/openfortivpn/issues/678#issuecomment-623094580. Which mechanism though? I'm open to suggestions.

DimitriPapadopoulos:
@a-andreyev I'm thinking of closing this pull request. Instead I suggest we continue the discussion about integration of _openfortivpn_ and GUI in existing open issues or in a new issue. Is that OK with you?

a-andreyev:
Sorry for a late reply! Yes, I guess it makes sense to continue the discussion in the existing open issues.

PR #730, opened 2020-06-15
Update checkpatch.pl from Linux kernel

DimitriPapadopoulos:
The default max line length has been changed from 80 to 100. It's about time!


DimitriPapadopoulos:
Interface with CI is broken again.
Replace with #731 in an attempt to get this fix through.

Issue #371, opened 2018-09-12
Support for adding a specified route


Ape:
I would like to add a custom route on the ppp interface after the VPN connection has been established. I can currently manually run `sudo ip route add 1.2.3.4/32 via 10.20.30.40 dev ppp0` to do this. But it would be a lot easier if openfortivpn did that for me when the interface is created and/or there is a working VPN connection. This way it wouldn't require manual intervention everytime my systems boots (and starts openfortivpn) or after VPN reconnect.

I would suggest something like `--add-route 1.2.3.4/32` which would automatically determine the correct gateway address and interface name from the VPN connection.

I would like to avoid `--set-routes` and `--half-internet-routes` because I have other networking and VPN connections on the same system.

DimitriPapadopoulos:
Duplicate of #22 as far as i can see. See [_pppd_ ip-up scripts](https://www.tldp.org/HOWTO/PPP-HOWTO/x1455.html) and option `--pppd-ipparam`.

Please do not hesitate reopen if that doesn't help.

thackel:
I have similar "problems".

As a "normal" user i don't want to modify/create any pppd/ifup/down scripts on a system global scale.
In the future it might be possible to let the Networkmanager do this like i want, but today at least the KDE version does not support any OTP dialog.

For now i created a script which starts and forks openfortivpn and then adds the routes as desired.
So a one-shot like the OP requested would be nice.

DimitriPapadopoulos:
I believe `--add-route 1.2.3.4/32` would only address part of the requests we have seen. We cannot possibly add multiple options for each different need.

Also I'm not certain it's (currently) possible to clean up the modified route from _openfortivpn_, after having forked _pppd_. Perhaps this is best left to _pppd_. Unfortunately [the privileged way to interact with _pppd_ seems to be through scripts in `/etc/ppp`](https://ppp.samba.org/pppd.html#sect13).

For the above reasons I still believe such needs would be better handled by scripts, either scripts run by _pppd_ or (at the other end) scripts that fork _openfortivpn_, as suggested above. Hell, I could even agree to a script directly run by _openfortivpn_ if there is no other easy way - but then again we cannot easily clean up when _pppd_ exits.

You are also raising the question how to help _normal_ users. Some ideas, probably not perfect:
* write documentation,
* provide sample scripts,
* have NetworkManager generate scripts,
* have _pppd_ run a script outside `/etc/ppp` (but how?),
[...]

Issue #22, opened 2015-10-16
add pppd options "call" or "file"


leo-b:
Sometimes additional pppd options are needed, e.g. to set connection specific routes when the link goes up.

If openfortivpn could pass a "call <filename>" (or "file <filename>") option to pppd, any additional pppd-options could be specified in that file.

mrbaseman:
I have just noticed this old issue. In 1.3.1 ipparam is supported. Is that useful for you? it is passed by pppd to the ip-up and ip-down scripts

DimitriPapadopoulos:
@mrbaseman Perhaps we should close this issue. I understand the new `ipparam` is an alternative solution.

@leo-b Any comments?

mrbaseman:
I agree, let's close it if there is no further reaction in the ticket in the near future.

PR #741, opened 2020-07-03
add set-default-route=[01]

oto-ledger:
Ability to disable default route for VPN with explicit routing only.
Few fixes to split routes to avoid routes with 0.0.0.0 as their destination to be unaccessible.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -59,6 +59,7 @@ const struct vpn_config invalid_cfg = {
 #endif
 	.use_syslog = -1,
 	.half_internet_routes = -1,
+	.set_default_route = -1,
 	.persistent = -1,
 #if HAVE_USR_SBIN_PPPD
 	.pppd_log = NULL,
@@ -320,6 +321,15 @@ int load_config(struct vpn_config *cfg, const char *filename)
 				continue;
 			}
 			cfg->half_internet_routes = half_internet_routes;
+		} else if (strcmp(key, "set-default-route") == 0) {
+			int set_default_route = strtob(val);
+
+			if (set_default_route < 0) {
+				log_warn("Bad set_default_route in config file: \"%s\".\n",
+				         val);
+				continue;
+			}
+			cfg->set_default_route = set_default_route;
 		} else if (strcmp(key, "persistent") == 0) {
 			unsigned long persistent = strtoul(val, NULL, 0);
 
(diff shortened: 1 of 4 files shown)


aaronjg:
What is the difference between this option and the "--set-routes" option? It looks quite similar to me, so if there is a difference it should at least be made clearer in the help text.

oto-ledger:
Well, --set-routes does not set any routes. Here we just want to set the routes sent by the server but not the default route. (Use the VPN connection to only access particular machine but not push all the traffic through it).

DimitriPapadopoulos:
This has been already discussed. See #22 and #371 for example.

oto-ledger:
Actually, it's not that I want to add particular routes, because I don't know them before connecting to the remote server. Those routes are dynamic on the server side. Therefore, I just wanted to apply them (needing some tweaks when they are specified using the default gateway), and avoid routing all the traffic through the vpn remote end.

DimitriPapadopoulos:
Yes, I understand that. Yet if the FortiGate gateway admins have set it up to request all traffic to be directed through the gateway, we do as _FortiClient_ and follow the admins' requests.

I'd rather address this as part of #678 - but not in the short unfortunately.

oto-ledger:
Oh, ok.
Also, if I'm not mistaken, even if the fortigate server does not specifies it, the client sets up the default route through the VPN, isn't it?

DimitriPapadopoulos:
Ah, I don't know, I'm not familiar with this part of the code. If so you have a case :-)

mrbaseman:
Actually the server can use either default routing mode or split tunnel mode. In default routing mode a default route will be added whereas in split tunnel mode only the routes pushed by the fortigate are added. The fortigate pushes rotes to all destinations where accept policies via ssl vpn are configured for this particular account.

An alternative to the default route are two routes with netmask /1 ("half internet routes") which can be used on the client side instead of changing the default route.

mrbaseman:
I think when the Fortigate pushes the default route, it doesn't push any other routes. 

So, if you are attempting to avoid the default route through the vpn connection,  you have to configure routing manually.

This would be --no-routes in combination with a pppd-ifup-script i.m.o.

mrbaseman:
I would close this one. Alternatives have been discussed above.

Issue #744, opened 2020-07-09
pinentry protocol parsing problem (OTP input problem when using NetworkManager)


jonas-koeritz:
I've debugged my NetworkManager-fortisslvpn problems when using the OTP option to login.
The pinentry program is expected to send the OTP in the format `D 123456\n` to the openfortivpn instance that called it. The NetworkManager plugin adds an additional `OK\n` to the response, indicating success.
I could not find any documentation on the pinentry protocol to determine wether this side or the NetworkManager side is wrong on this but the resulting OTP contains a `%0AOK` and is rejected by the gateway, preventing a successful login.
To fix this issue openfortivpn should drop the `OK` from the input or parse lines individually, the `D 123456\nOK\n` seems to be so short, and ends with `\n`, that it is treated as a single line by the current parsing code.

See also NetworkManager-fortisslvpn issue: https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn/-/issues/28

DimitriPapadopoulos:
Thank you for the report Jonas. Just to make sure I understand:
* In https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn/-/issues/28 you write:
"It works flawlessly with the same gateway and configuration when using openfortivpn to connect manually."
* Here you say that's an _openfortivpn_ issue.

I'm happy to have a look at this issue if needed, but could you clarify the above first? I t looks like _openfortivpn_ might already be taking "OK" into account:
https://github.com/adrienverge/openfortivpn/blob/713c836/src/userinput.c#L134-L142

jonas-koeritz:
My problem is that I couldn't find a formal definition of the protocol in use. openfortivpn could be more tolerant in parsing the line sent by the plugin or the plugin could send a modified line. Any of those changes would fix the problem.

jonas-koeritz:
> Thank you for the report Jonas. Just to make sure I understand:
> 
> * In https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn/-/issues/28 you write:
>   "It works flawlessly with the same gateway and configuration when using openfortivpn to connect manually."
> * Here you say that's an _openfortivpn_ issue.
> 
> I'm happy to have a look at this issue if needed, but could you clarify the above first? I t looks like _openfortivpn_ might already be taking "OK" into account:
> https://github.com/adrienverge/openfortivpn/blob/713c836/src/userinput.c#L134-L142

[...]

PR #760, opened 2020-07-25
Ensure only the pin code is sent

emelenas:
The pinentry program answers with "D PIN\nOK\n". Sometimes the "\nOK\n" gets sent to the gateway, causing errors. The change ensures only the PIN is sent, by taking the string up to, and discarding, the first "\n". See https://github.com/adrienverge/openfortivpn/issues/744

### DIFF
--- a/src/userinput.c
+++ b/src/userinput.c
@@ -97,7 +97,7 @@ static char *uri_unescape(const char *string)
 static int pinentry_read(int from, char **retstr)
 {
 	int bufsiz = 0;
-	char *buf = NULL;
+	char *buf = NULL, *saveptr = NULL;;
 	int len = 0;
 	int ret;
 
@@ -135,7 +135,7 @@ static int pinentry_read(int from, char **retstr)
 	    || strncmp(buf, "D ", 2) == 0) {
 		if (retstr) {
 			*retstr = strchr(buf, ' ');
-			*retstr = *retstr ? strtok_r(*retstr, "\n", NULL) : NULL;
+			*retstr = *retstr ? strtok_r(*retstr, "\n", &saveptr) : NULL;
 			*retstr = *retstr ? uri_unescape(*retstr + 1) : NULL;
 		}
 		free(buf);
--- a/src/userinput.c
+++ b/src/userinput.c
@@ -97,7 +97,7 @@ static char *uri_unescape(const char *string)
 static int pinentry_read(int from, char **retstr)
 {
 	int bufsiz = 0;
-	char *buf = NULL, *saveptr = NULL;
+	char *buf = NULL;
 	int len = 0;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


src/userinput.c
```diff
@@ -135,6 +135,7 @@ static int pinentry_read(int from, char **retstr)
 	    || strncmp(buf, "D ", 2) == 0) {
 		if (retstr) {
 			*retstr = strchr(buf, ' ');
+			*retstr = strtok(*retstr, "\n");
```
DimitriPapadopoulos, line 138: Use **`strtok_r`** instead of `strtok`. See #571.

emelenas:
```suggestion
			*retstr = strtok_r(*retstr, "\n");
```

emelenas: O, I was not aware

DimitriPapadopoulos: [Function `strtok_r` requires 3 arguments](https://man7.org/linux/man-pages/man3/strtok_r.3.html).

emelenas: OK, sorry. I didn't test.

```suggestion
                        char *saveptr = NULL;
			*retstr = *retstr ? strtok_r(*retstr, "\n", &saveptr) : NULL;
```

emelenas: I'll update the fork and make a new merge request.

src/userinput.c
```diff
@@ -135,6 +135,7 @@ static int pinentry_read(int from, char **retstr)
 	    || strncmp(buf, "D ", 2) == 0) {
 		if (retstr) {
 			*retstr = strchr(buf, ' ');
+			**retstr = *retstr ? strtok_r(*retstr, "\n", &saveptr) : NULL;
```
DimitriPapadopoulos, line 138: For a single call to `strtok_r` I believe you don't need `saveptr`, isn't `NULL` enough as the 3rd argument?

Also this is suspect:
```
**retstr = *retstr [...]
```
Perhaps is should be:
```
*retstr = *retstr [...]
```

emelenas:
Maybe I get it right this time? Sorry, I shouldn't be multitasking.

DimitriPapadopoulos:
Looks good to me now. Have or could you build and install this version of _openfortivpn_ and make sure it works with _NetworkManager_ and fixes the reported issue?

emelenas:
I have built it and tested OK, but I will ask the original poster to also test it

src/userinput.c
```diff
@@ -135,6 +135,7 @@ static int pinentry_read(int from, char **retstr)
 	    || strncmp(buf, "D ", 2) == 0) {
 		if (retstr) {
 			*retstr = strchr(buf, ' ');
+			*retstr = *retstr ? strtok_r(*retstr, "\n", &saveptr) : NULL;
```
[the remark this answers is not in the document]
emelenas: Sorry, it does not compile with NULL as argument:
src/userinput.c:138:24: warning: null argument where non-null required (argument 3) [-Wnonnull]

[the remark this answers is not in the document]
DimitriPapadopoulos: Ah, you're absolutely right. Besides we use `saveptr` everywhere else.

PR #763, opened 2020-07-30
properly update resolvconf if use option "use-resolvconf=0"

balagurusurendar:
1) replace_char(dns_suffix, ';', ' '); missed in while delete name-server so added it
2) dns_suffix ends with ; have ends with " " so trimmed the dns_suffix
3) strtok_r method changed the input buffer so resolv_conf updation not working properly if name-server already present

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -1060,6 +1060,79 @@ static inline char *replace_char(char *str, char find, char replace)
 	return str;
 }
 
+//string trim function it removes trim chars from first & end of string
+static inline void str_trim(const char* input, char* output,char* trim_chars){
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
To tell the the truth, I don't understand what problem you're trying to solve. Can you describe why you're providing this patch? What's with the `;` character?

DimitriPapadopoulos:
@balagurusurendar Could you explain the problem this patch is trying to solve?

DimitriPapadopoulos:
@balagurusurendar Can you perhaps comment on the issue this patch is solving?

PR #799, opened 2020-10-24
Print part of the default SSL socket options

### LINKED ISSUE #795: Add getsockopt() to debug TCP-keepalive
Copied from https://github.com/adrienverge/openfortivpn/issues/794#issuecomment-714397002:
> The observation, however, was that this didn't help at least on MacOSX. So I didn't have a clear idea how I should proceed further in #613 . When a timeout has happened to me (on linux) some weeks ago, I have seen the error messages (probably coming from pppd ?), which I understand that something below already implements some sort of keepalive mechanism.
> Maybe instead of guessing what might be happening, we should just put in some lines of debug code which print out the result of getsockopt - and see on which platforms tcp-keepalive is already enabled and where it's not...

DimitriPapadopoulos:
We want to compare their value between platforms to investigate TCP keepalive discrepancies.

Fixes #795.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -39,7 +39,6 @@
 #include <errno.h>
 #include <signal.h>
 #include <string.h>
-#include <assert.h>
 
 #if HAVE_MACH_MACH_H
 /* this is typical for mach kernel used on Mac OS X */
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
Perhaps the _getsockopt()_/_setsockopt()_ calls should occur early. For example the TCP(7) manual page mentions:
> On individual connections, the socket buffer size must be set prior to the listen(2) or connect(2) calls in order to have it take effect. See socket(7) for more information.

Should I move the setsockopt() calls from _io_loop()_ to _tcp_connect()_ - where there's already a _setsockopt()_ call by the way?

mrbaseman (approved):
This looks good to me

mrbaseman:
> Perhaps the _getsockopt()_/_setsockopt()_ calls should occur early. For example the TCP(7) manual page mentions:
> 
> > On individual connections, the socket buffer size must be set prior to the listen(2) or connect(2) calls in order to have it take effect. See socket(7) for more information.
> 
> Should I move the getsockopt() calls from _io_loop()_ to _tcp_connect()_ - where there's already a _setsockopt()_ call by the way?

I would expect that you can not change buffer sizes once the socket is already in use, but for reading the options that are currently set I would expect that it doesn't matter when the getsockopt() call is issued. 
Well, you could imagine that a library (openssl in this case) might change the settings too late (by wrong usage of the functions?) and on top of this getsockopt() might return these values while the ones that are effective are different ones... well... maybe. So, to be on the safe side, you could move the getsockopt() calls as you write.

PR #802, opened 2020-10-28
Adding --pppd-keepalive option.

FinboySlick:
In reference to  #92 where lcp-echo-interval would still be useful.  Figured if I'm going to ask for a feature, I might as well do some of the work.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -65,6 +65,7 @@ const struct vpn_config invalid_cfg = {
 	.pppd_log = NULL,
 	.pppd_plugin = NULL,
 	.pppd_ipparam = NULL,
+	.pppd_keepalive = NULL,
 	.pppd_ifname = NULL,
 	.pppd_call = NULL,
 #endif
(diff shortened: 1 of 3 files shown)


DimitriPapadopoulos:
Would we need to consider other LCP parameters? This works with `pppd` on Linux but I cannot find anything similar in `ppp` on FreeBSD. Do you have any clue?

FinboySlick:
Honestly, my initial approach would have been to enable any pppd option via a generic --pppd-opt= type of parameter, but I wasn't familiar enough with the code/intent to take into account the design and security implications (and it was very late).

As such, this patch is just to solve an immediate problem. That being said, I think there would be value in making openfortivpn 'independent' of other config files. Since the only way to do that as far as pppd is concerned is the command line, a '--pppd-opt=' type of solution is worth considering.

DimitriPapadopoulos:
I think FortiClient handles dead peer detection that at the PPP level, see for example https://gitlab.com/openconnect/openconnect/-/commit/e1eac267, but perhaps in upper layers too.

mrbaseman:
Is this still an issue? There have been changes to `pppd` (which broke some things in `openfortivpn`, too, e.g. in  v1.21.0 we have introduced a compatibility feature to deal with `pppd` versions older than 2.5.0). 

To make it clear: those changes, that I'm aware of, have no relation with keepalive, I'm just asking if this PR is still relevant for current versions of  `pppd`, simply because I did not closely follow changes there.

PR #812, opened 2020-12-10
Adding ppc64le architecture

genisysram:
Hi
Here is my contribution to your code, its working good on powersystems.

Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 os:
 - linux
 
+arch:
+    - amd64
+    - ppc64le
+
 language: c
 
 sudo: false


mrbaseman:
This patch just adds another architecture for the Travis checks. I'm not sure if this extra compute cycles are worth it, given the fact that there is no difference in the source code associated to the change.

DimitriPapadopoulos:
Hi,

Nice to learn it works for you on PPC systems.

Indeed, Travis CI merely runs lint-like tests and compiles:
https://github.com/adrienverge/openfortivpn/blob/master/tests/lint/run.sh

No PPC issues have ever been reported and I honestly don't expect any in the future. Is it really worth running the additional tests?

genisysram:
We exhaustively tested the Repo in our local ppc64le machine and ensured its working perfectly ,This also ensures that whatever versions  you are using is complete and supports all features on ppc64le. And, it is possible for go programs to be architecture dependent - if someone ever adds a dependency on the hardware, expected or not, this will ensure that the project will always build when it gets to the distros - if not, the distro porting team winds up picking up the work which can be painful to those coming in from outside the project

DimitriPapadopoulos:
We have discarded Travis CI. See #851.

Issue #801, opened 2020-10-28
Do you know something about TUN method?


rain2fog:
Hi Adrien Vergé
    It's amazing when I first time found your project.  Thank you for giving this open source project so much .... 
    I analyzed the http request about "/remote/fortisslvpn_xml ". I found vpn server side would response a special xml content which contain two entry **tunnel-method**. The first item is **ppp**. The secound item is **tun**. You chose **ppp** as the local net forwarding solution. 
    Do you have any clues about **tun** as the local net forwarding solution?

adrienverge:
Hello Hang Zhou, I have no clue about the *tun* entry. Maybe it's usable to set up a connection, but I guess it's dependent on the Fortinet appliance version.

DimitriPapadopoulos:
@raincloud Also just wondering how portable would the _tun_ entry be on macOS. Can you shed some light on the subject?

rain2fog:
we simulated the PPP negotiation procedure. There are 3 stages in PPP protocal. 
The 1th stage is LCP. client will send Magic-Number and MRU request to fortivpn server, then will ack the  server request. server will send the Magic-Number and ack the client request.
The 2th stage is IPCP. both will will send the IPCP config-request to other. client need to send the IPCP config-request Option (3) with the IP which come from the XML. then both config-request need to be ack.
The 3th stage is 0x0021. we can read and write IP packet from tun NIC. then sending data by encapsulating or decapsulating the IP packet with header 0x0021(2 Bytes) + (totolsize(2 Bytes) + 0x5050 + pppsize(2 Bytes)).
we also need send the LCP echo-request to fortinet VPN to keep the TCP  connection wouldn't close.

DimitriPapadopoulos:
@rain2fog We spawn `pppd` because when _openfortivpn_ had been written no PPP libraries were available at the time. Nowadays GPL-compatible PPP code is available: https://github.com/adrienverge/openfortivpn/issues/650#issuecomment-624511604. Do you mean we could use TUN and such PPP code instead of spawning `pppd`? I'm not sure what we should do with the above information. Do you plan on providing a patch?

rain2fog:
@DimitriPapadopoulos you are right. This is the patch for demo.  [tun.patch](http://8.210.22.94/tun.patch)

DimitriPapadopoulos:
@rain2fog Impressive work! I'll try to test it soon. Would you agree to open a pull request to discuss the patch?

[...]

PR #818, opened 2020-12-18
Support TUN tunnel instead of using `pppd`

DimitriPapadopoulos:
From https://github.com/adrienverge/openfortivpn/issues/801#issuecomment-735513189:
> we simulated the PPP negotiation procedure. There are 3 stages in PPP protocal.
> * The 1th stage is LCP. client will send Magic-Number and MRU request to fortivpn server, then will ack the server request. server will send the Magic-Number and ack the client request.
> * The 2th stage is IPCP. both will will send the IPCP config-request to other. client need to send the IPCP config-request Option (3) with the IP which come from the XML. then both config-request need to be ack.
> * The 3th stage is 0x0021. we can read and write IP packet from tun NIC. then sending data by encapsulating or decapsulating the IP packet with header 0x0021(2 Bytes) + (totolsize(2 Bytes) + 0x5050 + pppsize(2 Bytes)).
we also need send the LCP echo-request to fortinet VPN to keep the TCP connection wouldn't close.


DimitriPapadopoulos:
Re-opened as #1048.

PR #820, opened 2020-12-18
Hopefully help reach higher speeds on macOS

### LINKED ISSUE #428: Limited download speed on MacOS Mojave
When I download files from a server behind the VPN using openfortivpn my download speeds are capped around 2.5mb/sec. When I perform the same download behind the official Fortinet client, I can maximize my connection speed (around 10mb/sec). 

Any ideas what might be happening there? The problem is reproducible on different MacOS laptops when downloading the same files.

DimitriPapadopoulos:
While on Linux the _speed_ option does not seem to be taken into account, on macOS it appears to be throttling the VPN connection.

According to the [`pppd` man page](https://ppp.samba.org/pppd.html), Linux and other system "only support the commonly-used baud rates". However, `pppd` does not require specific baud rates in this use case.

Fixes #428?

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -232,7 +232,7 @@ static int pppd_run(struct tunnel *tunnel)
 		} else {
 			static const char *const v[] = {
 				ppp_path,
-				"115200", // speed
+				"230400", // speed
 				":192.0.2.1", // <local_IP_address>:<remote_IP_address>
 				"noipdefault",
 				"noaccomp",


rkirkpat (approved):
With a 40mbps (down) DSL connection, was only getting about 600kB/sec on an SCP over the VPN compared to nearly 4MB/sec with Fortinet's client. I cloned the git repo for openfortivpn, switched to the `v1.15.0` tag, applied this fix, rebuilt, and re-tested. I was now indeed getting 4MB/sec over the openfortivpn session!

mrbaseman (commented):
I believe I have tested different values in the context of #444 and I remember that on some platform one of the popular baud rates had to be chosen, otherwise ppp(d) refused to start. 
Maybe we should guard this change with an `#ifdef`for Mac OSX only?

DimitriPapadopoulos:
Yes, I too recall `pppd` not starting with exotic baud rates - but perhaps it simply wouldn't start **without** a baud rate. Indeed, `pppd` does start with `2147483647` on Linux (Ubuntu 18.04) and macOS. Perhaps it used not to work on FreeBSD or older Linux distributions. I'll double-check on older CentOS releases. I would expect FreeBSD to be similar to macOS, but then I'm not certain they ship with the same PPP software.

DimitriPapadopoulos:
Works with `2147483647` on Ubuntu 18.04, from within Ubuntu 18.04, or from the command line:
```
$ sudo pppd 2147483647 :192.0.2.1 noipdefault noaccomp noauth default-asyncmap nopcomp receive-all nodefaultroute nodetach lcp-max-configure 40 mru 1354
~�}#�!}!}!} }.}!}$}%J}%}&(��}>��~~�}#�!}!}!} }.}!}$}%J}%}&(��}>��~
```

DimitriPapadopoulos:
Works too on CentOS 6.4, from command line:
```
# pppd 2147483647 :192.0.2.1 noipdefault noaccomp noauth default-asyncmap nopcomp receive-all nodefaultroute nodetach lcp-max-configure 40 mru 1354
```

ehaupt:
Here is the FreeBSD man page of [ppp](https://www.freebsd.org/cgi/man.cgi?query=ppp&apropos=0&sektion=0&manpath=FreeBSD+13.0-current&arch=default&format=html).

On my system when running `openfortivpn` `ppp` runs like this:

```
root   44493   /usr/sbin/ppp -direct
root   44492   openfortivpn -v --no-dns --persistent=1 -c my-endpoint.conf
```

Is there anything I can run to help?

DimitriPapadopoulos:
@ehaupt Just occurred to me that we run `ppp` on FreeBSD, not `pppd`. Therefore, the _speed_ parameter does not apply to FreeBSD.

While we have you here, have you ever noticed bandwidth issues on FreeBSD, similar to the ones on macOS? Probably not, as this issue seems related to the specific `pppd` setup on macOS - but I thought I'd ask anyway.

ehaupt:
@DimitriPapadopoulos: It's hard to tell. Our endpoints (not in my control) are generally very slow even with the official Mac/Windows clients. I get about the same speed on a variety of different machines/network connections.

skycaptain:
Works for me as well. ~300KB/sec vs 3MB/sec with this patch on MacOS 10.15.7. Still not great but at least on par with the official Fortigate client. VSCode Remote SSH connection is now finally usable (had reconnection and buffer issues due to the bad connection). Had to apply #823 as well to rid of the error message reported in #822 during building.

DimitriPapadopoulos:
@skycaptain Thank you very much for testing. Can I ask for some additional information?
* What do you mean exactly by "~300KB/s vs 3MB/s"? Is that the VPN speed vs. the speed of your internet link?
* According to https://github.com/adrienverge/openfortivpn/issues/822#issuecomment-757041467, `230400` works as well as `2147483647`. Could you try with `230400`? If it works, I would rather use `230400` as it is supported on all platforms.

skycaptain:
@DimitriPapadopoulos ~300KB/s is downloading a larger file with current 1.15.0 version from my company's network. 3MB/s is after applying this patch (and #823 as written). However, the connection is still throttled as I usually download files at around 10-15MB/s.
I've tested the changes with iperf3 (connection from my Macbook --> Wifi 5 --> Home router --> FTTB --> Internet -->  Fortigate VPN with 1GBit upstream --> 40Gbit --> 25Gbit --> Server).

These are the results with 1.15.0:

```plain
[ ID] Interval           Transfer     Bitrate
[  7]   0.00-10.00  sec  2.96 MBytes  2.48 Mbits/sec                  sender
[  7]   0.00-10.00  sec  2.84 MBytes  2.38 Mbits/sec                  receiver
```

With this patch (`2147483647`) applied:

```plain
[ ID] Interval           Transfer     Bitrate
[  7]   0.00-10.00  sec  46.3 MBytes  38.8 Mbits/sec                  sender
[  7]   0.00-10.00  sec  45.4 MBytes  38.0 Mbits/sec                  receiver
```

With `230400`:

```plain
[ ID] Interval           Transfer     Bitrate
[  7]   0.00-10.00  sec  30.3 MBytes  25.4 Mbits/sec                  sender
[  7]   0.00-10.00  sec  30.3 MBytes  25.4 Mbits/sec                  receiver
```

DimitriPapadopoulos:
@skycaptain So the baud rate does seem to have an effect on the VPN speed, but I cannot find a sensible or linear relation between the baud rate and the VPN speed:
* 115200 Bd ≈ 100 kb/s ⟷ 2.48 Mb/s
* 230400 Bd ≈ 200 kb/s ⟷ 25.4 Mb/s
* 2147483647 Bd ≈ 2000 Mb/s ⟷ 38.8 Mb/s

In any case, if as you wrote the VPN speed is now on par with the official FortiClient, so I suppose we won't be able to get past 38.8 Mb/s because of a limitation elsewhere, perhaps on the FortiGate itself.

Are you perhaps able to test whether you can reach higher VPN speeds from Linux or Windows (with VPN SSL, not IPSec)? Also, could you perhaps try a few other baud rates?
* `0`
* `64424509440` (I suspect this won't work as `pppd` appears to be reading the speed argument into an `int`)

skycaptain:
@DimitriPapadopoulos There are most certainly also issues with my company's gateway as some colleagues of mine are complaining about bad speed rates even with the official client (maybe CPU limitations). Unfortunately, I can not benchmark the system separately, as it's being used by many colleagues around the clock. So please take my measurements above with care.

I've tested  `0` and `64424509440`, which both got me `WARN:   read returned 0` and no working connection.

Why not use `230400`? This is the value that the official client uses and works on all systems. It's definitely an improvement over the current setting.

DimitriPapadopoulos:
OK then, we can settle on `230400` which is supported on all systems.

Here are the well-known baud rates from `<asm-generic/termbits.h>` on Linux:
```
#define  B50	0000001
#define  B75	0000002
#define  B110	0000003
#define  B134	0000004
#define  B150	0000005
#define  B200	0000006
#define  B300	0000007
#define  B600	0000010
#define  B1200	0000011
#define  B1800	0000012
#define  B2400	0000013
#define  B4800	0000014
#define  B9600	0000015
#define  B19200	0000016
#define  B38400	0000017
```

```
#define    B57600 0010001
#define   B115200 0010002
#define   B230400 0010003
#define   B460800 0010004
#define   B500000 0010005
#define   B576000 0010006
#define   B921600 0010007
#define  B1000000 0010010
#define  B1152000 0010011
#define  B1500000 0010012
#define  B2000000 0010013
#define  B2500000 0010014
#define  B3000000 0010015
#define  B3500000 0010016
#define  B4000000 0010017
```

ah-:
Wow, I can confirm this much improves connection speed on mac os!

Thanks a lot for figuring this out.

DimitriPapadopoulos:
@ah Which change did help in your case, for the record?
* `115200` → `230400`
* `192.0.2.1` → `169.254.2.1`

ah-:
I've only pulled master and recompiled from there, not sure which of the two changes it was. I can play around a bit over the weekend.

DimitriPapadopoulos:
Oh, there's no real need. This was out of curiosity, just in case we need to revert one of the charges in the future.

PR #828, opened 2021-01-13
Revert from dynamic to automatic allocation of password

### LINKED ISSUE #826: Buffer overflow in password when many special chars
I've been facing an issue with a long password (256 char) containing many special chars.
Apparently the HTTP request is truncated when the url encoded password is too large..
Here is the debug log with a bit shorter password
```
DEBUG:  http_send:
POST /remote/logincheck HTTP/1.1
Host: XXXXX:XXXX
User-Agent: Mozilla/5.0 SV1
Accept: */*
Accept-Encoding: gzip, deflate, br
Pragma: no-cache
Cache-Control: no-store, no-cache, must-revalidate
If-Modified-Since: Sat, 1 Jan 2000 00:00:00 GMT
Content-Type: application/x-www-form-urlencoded
Cookie: 
Content-Length: 243

username=xxxxx&credential=********************************************************************************************************************Cr%60LK%25Gj%3F%24jFiO_1%23%7D%5EBgg%23%60~%3Fa%2F%298%3FaEL%5Boy%289_%23sM%22qcg%27&realm=&ajax=1
```

Here is the result with a full length password:
```
DEBUG:  http_send:
oy%289_%23sM%22qcg%27
```

Changing the password for the same length with less special characters works.
Official Forti client works with initial password / login on the webpage is also working.

DimitriPapadopoulos:
Revert #361. Replace `FIELD_SIZE` by more specific variables. Use a maximum length of 256 characters for `password`.

Fixes #826.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -292,8 +292,8 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			free(cfg->pinentry);
 			cfg->pinentry = strdup(val);
 		} else if (strcmp(key, "realm") == 0) {
-			strncpy(cfg->realm, val, FIELD_SIZE - 1);
-			cfg->realm[FIELD_SIZE] = '\0';
+			strncpy(cfg->realm, val, REALM_SIZE);
+			cfg->realm[REALM_SIZE] = '\0';
 		} else if (strcmp(key, "set-dns") == 0) {
 			int set_dns = strtob(val);
 
--- a/src/config.c
+++ b/src/config.c
@@ -268,7 +268,7 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			cfg->password[PASSWORD_SIZE] = '\0';
 			cfg->password_set = 1;
 		} else if (strcmp(key, "otp") == 0) {
-			strncpy(cfg->otp, val, FIELD_SIZE - 1);
+			strncpy(cfg->otp, val, FIELD_SIZE);
 			cfg->otp[FIELD_SIZE] = '\0';
 		} else if (strcmp(key, "otp-prompt") == 0) {
 			free(cfg->otp_prompt);
--- a/src/config.c
+++ b/src/config.c
@@ -260,8 +260,8 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			}
 			cfg->gateway_port = port;
 		} else if (strcmp(key, "username") == 0) {
-			strncpy(cfg->username, val, FIELD_SIZE - 1);
-			cfg->username[FIELD_SIZE] = '\0';
+			strncpy(cfg->username, val, USERNAME_SIZE);
+			cfg->username[USERNAME_SIZE] = '\0';
 		} else if (strcmp(key, "password") == 0) {
 			cfg->password = strdup(val);
 		} else if (strcmp(key, "otp") == 0) {
--- a/src/config.c
+++ b/src/config.c
@@ -249,8 +249,8 @@ int load_config(struct vpn_config *cfg, const char *filename)
 		}
 
 		if (strcmp(key, "host") == 0) {
-			strncpy(cfg->gateway_host, val, FIELD_SIZE);
-			cfg->gateway_host[FIELD_SIZE] = '\0';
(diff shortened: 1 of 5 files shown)


mrbaseman (commented):
I prefer static buffer sizes, too, if they are chosen large enough. So, to address the problem reported in #826 I'd go in the same direction.

However, we had a couple of issues about logging in with a certificate only or a token only, and there is a difference between not setting a password and setting it to the empty string (at least in the config). We should be careful not to break the possibility to make a distinction if the user shall be asked for a password and if it shall be sent at all. 

I have seen you have tried to handle different array sizes in a similar manner, but with a separate constant for each of them. I like this. When reading quickly through the code I had the impression that the +1 and -1 for the array arguments are not yet entirely handled the same way everywhere, but I must admit that I didn't inspect the code around these places in detail.

DimitriPapadopoulos:
The +1 and -1 issues have all been fixed. The only outlier is `IF_NAMESIZE` because it includes the terminal `'\0'`. Should I `#define IFACE_NAME_SIZE ((IF_NAMESIZE) - 1)` to clarify the intent? But then I feel the code gets more convoluted, worse instead of better.

DimitriPapadopoulos:
Yes, I do recall there is a difference between not setting the password (no `password`option) and setting an empty password (`password = `) in the config file. Yet I cannot find a place in the authentication code (typically the `http.c` file) that distinguishes between no password or empty password. We seem to be merely testing `tunnel->config->password[0] == '\0'`:
https://github.com/adrienverge/openfortivpn/blob/b123e99/src/http.c#L649-L666

I think this is only a matter of parsing the config file and merging with the command line options, in which case we need to review in more detail the relevant code. If needed II'd rather add an explicit additional variable to the configuration, instead of relying on the implicit, hidden meaning of the difference between `password == NULL` and `password[0] == '\0'`.

By the way, just caught a residual `password == NULL` test which is at the heart of the above discussion:
https://github.com/DimitriPapadopoulos/openfortivpn/blob/fde9445/src/main.c#L555
This test occurs during the command-line options / configuration file merge operation.

DimitriPapadopoulos:
With the last update, I think I have fixed the `password == NULL` test. I cannot say for sure though, as I'm not certain about the intended semantics of missing `password` and `password = ` in the configuration file. Can you help me with that?

mrbaseman:
@DimitriPapadopoulos 
> Can you help me with that?
I'll see if I can find some time during the weekend to look deeper into this

DimitriPapadopoulos:
I believe the real issue is not to ask for a password interactively when we have an "empty password" instead of "no password" - with the configuration file option **`password =`** or the command line option **`--password ""`**. We can implement that with an additional variable `empty_password` in the configuration to distinguish between these two cases.

From then on, I believe the "empty password" and "no password" cases are identical.

mrbaseman:
yes, I agree. With an empty password entry in the config file, or, when no username is specified in the config file nor on the command line, the [interactive password prompt is skipped](https://github.com/adrienverge/openfortivpn/blob/6b8f9fee1dd1c911d423b175da04b578a86f10b6/src/main.c#L595-L608).
 
If the password is still `NULL` in the config after that section, an [empty string is assigned](https://github.com/adrienverge/openfortivpn/blob/6b8f9fee1dd1c911d423b175da04b578a86f10b6/src/main.c#L617), so as you say, further on the cases of empty password and no password are identical. 

With the username we have a similar topic: If it is empty and no user certificate is configured, [an error message is shown](https://github.com/adrienverge/openfortivpn/blob/6b8f9fee1dd1c911d423b175da04b578a86f10b6/src/main.c#L590-L594).

DimitriPapadopoulos:
I believe I have addressed the password issues:
* if I pass an empty password using the command-line option `--password "`, _openfortivpn_ does not prompt for a password,
* else if I define an empty password in the configuration file (`password = `), _openfortivpn_ does not prompt for a password,
* else if the password is empty and the username is not, _openfortivpn_ will prompt for a password.

mrbaseman:
@DimitriPapadopoulos Now I'm confident that the changes don't break anything

PR #830, opened 2021-01-18
Embedded solutions support

mobrembski:
I like this project, and i wanted to use it  in open source embedded router firmware FreshTomato.
What was very surprise for me, it works very well with just a few little changes:

1. Added customization at compile time of path to resolv.conf. Don't hardcode it. In embedded systems /etc directory is sometimes placed on read-only flash partiton, which lead application to fail with modification.
2. Surround AC_CHECK_FILES with testing if we're under cross-build. Autotools doesn't support AC_CHECK_FILES in such case. i propose to fall-back to default configuration in that cases.
3. Add borrowed from Samba project libreplace ifaddrs.c, for embedded replacement of getifaddrs and freeifaddrs. My project doesn't have ifaddrs, but this replacement works very well.

Any comments are welcome.


mobrembski:
Now good question is how i should deal with coding style for ifaddrs.c?
It's borrowed component, i could fix it but then any update from upstream would be painful.

configure.ac
```diff
@@ -320,26 +320,30 @@ AS_IF([test "x$enable_proc" = "x"], [
...
+	if test "$cross_compiling" != "yes"; then
+		AC_CHECK_FILE([$PPP], [
+			AS_IF([test "x$PPP_PATH" = "x"], [
+				PPP_PATH="$PPP"
+			])
+			AS_IF([test "x$with_ppp" = "x"], [
+				with_ppp="yes"
+			])
+		],[])
+	fi
```
DimitriPapadopoulos, line 328: The idea was that you should explicitly specify the location of `pppd` when cross-compiling:
https://github.com/adrienverge/openfortivpn/blob/6b8f9fe/configure.ac#L274

But I get it, at least you don't get an obscure error message. Still, perhaps an error message is more appropriate, but we need a better one, complaining `if test "$cross_compiling" = "yes"` and `test "x$PPP_PATH" = "x"`.

DimitriPapadopoulos:
> Now good question is how i should deal with coding style for ifaddrs.c?

Just skip the file when testing, see [`run.sh`](https://github.com/adrienverge/openfortivpn/blob/master/tests/lint/run.sh).

mobrembski (commented):
Good :) done :)

mobrembski:
It seems that somehow GitHub actions is pulling old tests/lint/run.sh.
Style check fails on ifaddrs.c file, which is (and should be) ignored.
I've added an rule but github actions is still using old one.

DimitriPapadopoulos:
Ah, after changing from Travis CI to GitHub Actions, we do not use `tests/lint/run.sh` any more, see [`.github/workflows/openfortivpn.yml`](https://github.com/adrienverge/openfortivpn/blob/master/.github/workflows/openfortivpn.yml).

Thinking about it, it's a confusing regression: manual checks and GitHub Actions are triggerred differently. On the other hand we get better reports from GitHub Actions.

Issue #650, opened 2020-04-18
Revisit running openfortivpn as root?


DimitriPapadopoulos:
After #373 _openfortivpn_ must be run with root privilege. There are multiples reasons for that. It would be worth re-investigating whether there are ways around that, or at least whether dropping root privilege after initial setup is a possibility, for example after spawning _pppd_.

### Spawning _pppd_

Members of the [`dip`](https://wiki.ubuntu.com/Security/Privileges#Connect_to_the_Internet_using_a_modem) group may run `pppd` on Linux distributions such as Debian or Ubuntu:
```sh
$ ls -l /usr/sbin/pppd
-rwsr-xr-- 1 root dip 395144 Feb 11 16:03 /usr/sbin/pppd
$ 
```
Yet _openfortivpn_ requires root privilege because option [**`noauth`**](https://ppp.samba.org/pppd.html) is privileged:
```sh
$ pppd noauth
pppd: using the noauth option requires root privilege
$ 
```
Not sure how to work around this is an a generic way - apart from complex solutions such as splitting _openfortivpn_ into multiple pieces of software with only the one spawning _pppd_ running with root privileges.

### Setting routes

The [`CAP_NET_ADMIN`](http://man7.org/linux/man-pages/man7/capabilities.7.html) capability might be enough for _ipv4_set_route_() / _ioctl_():
* I don't know how easy it is to manage capabilities (probably apply [_setcap_](http://man7.org/linux/man-pages/man8/setcap.8.html) to _openfortivpn_) and whether Linux distributions are willing to allow/manage capabilities.
* In any case we could check either for _root_ `geteuid() == 0` **or** the current process capabilities with something like [`prctl(PR_CAPBSET_READ, CAP_NET_ADMIN)`](http://man7.org/linux/man-pages/man2/prctl.2.html).

Alternatively routes might be handled outside of _openfortivpn_:
* Routing can be handled by the calling framework, for example [_NetworkManager_](https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn). Use option `--set-routes=0`/`--no-routes`.
* Routing could be handled by _openfortivpn_ call-back scripts. Such scripts would require specific sudo privileges. See [Wrapper for _ip_](https://community.openvpn.net/openvpn/wiki/UnprivilegedUser#Wrapperforip).

### Name resolution

DNS servers and search domains might be handled outside of _openfortivpn_:
[...]

PR #833, opened 2021-01-21
Assign an IPv4 address from the link-local subnet

### LINKED ISSUE #832: 192.0.2.1 → 169.254.2.1
Modify non-routable address used by PPP as suggested by @ultrafiber in https://github.com/adrienverge/openfortivpn/issues/822#issuecomment-763202546

DimitriPapadopoulos:
While the 192.0.2.0/24 TEST-NET-1 block would avoid collisions,
the 169.254.0.0/16 subnet described in RFC3927 has precisely been
specified precisely for such cases.

We should probably select "an address using a pseudo-random number
generator with a uniform distribution in the range from 169.254.1.0
to 169.254.254.255 inclusive", as described in RFC3927. However,
that would of course require more code and I'm lazy, but more
importantly it had been reported that FortiClient uses 169.254.2.1.
I'd rather use the same address as FortiClient.

Fixes #832.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -232,7 +232,7 @@ static int pppd_run(struct tunnel *tunnel)
 			static const char *const v[] = {
 				ppp_path,
 				"230400", // speed
-				":192.0.2.1", // <local_IP_address>:<remote_IP_address>
+				":169.254.2.1", // <local_IP_address>:<remote_IP_address>
 				"noipdefault",
 				"noaccomp",
 				"noauth",


mrbaseman (commented):
Back in #280 we had a discussion if this address does really matter. We came to the conclusion that an arbitrary address could be used here for building up the communication over ppp. It's send as a proposal and the remote side uses it until dhcp has assigned a proper address. 

The only problem I see is that a host route to this address is added via the ppp interface. The link local addresses can coexist on different links and they don't need explicit routing as far as I know. 

I'm not sure what happens if this particular address is used on several interfaces and an explicit host route via one particular interface is added. Does the communication on the other interfaces still work, even if the routing table would suggest to send the packets over the ppp link?

DimitriPapadopoulos:
Not certain about the issues you raise either. I don't want to modify something that works, it's just that this IP address might have an effect on the VPN speed on macOS Mojave.

DimitriPapadopoulos:
@ultrafiber Do we really need to change `192.0.2.1` → `169.254.2.1` in _openfortivpn_? Does the value of this IP address really have an effect on _openfortivpn_ speed on macOS?

mrbaseman:
I have given this change a try and here is the resulting routing table on Linux:
```
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         192.168.XYZ.254 0.0.0.0         UG    100    0        0 enp2s0
169.254.0.0     0.0.0.0         255.255.0.0     U     1000   0        0 enp2s0
169.254.2.1     0.0.0.0         255.255.255.255 UH    0      0        0 ppp0
192.168.XYZ.0   0.0.0.0         255.255.255.0   U     100    0        0 enp2s0
ABC.DEF.GHI.0   0.0.0.0         255.255.255.0   U     0      0        0 ppp0
ABC.DEF.OPQ.LMN 192.168.XYZ.254 255.255.255.255 UGH   0      0        0 enp2s0
ABC.DEF.OPQ.0   0.0.0.0         255.255.255.0   U     0      0        0 ppp0
```
While there is already a class B network route on the ethernet device for the link local network, _openfortivpn_ adds a host route over the ppp interface. 

So I must correct myself in this point:

> The link local addresses can coexist on different links and they don't need explicit routing as far as I know.

There is a network route, and it would still work even if we add a host route to a single address of this network on a different interface.

My expectation is that nothing gets broken unless something else (another instance of _openfortivpn_?) uses `169.254.2.1` on a different interface. If there are any problems, the effect would probably be the same with the `192.0.2.1` address used so far - just for a different one. [RFC 3927](https://tools.ietf.org/html/rfc3927) discusses the topic _Conflict detection and defense_ in Section 2.5 and _Considerations for Multiple Interfaces_ in Section 3, but I'm not sure to which extend all this applies to the ppp link that we have in our case.

DimitriPapadopoulos:
@mrbaseman If I understand you correctly, the situation won't be worse with `169.254.2.1` than with `192.0.2.1`, will it? If so I suggest we apply this patch.

Then we could address _Conflict detection and defense_ in Section 2.5 and _Considerations for Multiple Interfaces_ in Section 3 later on. How soes this sound?

mrbaseman:
Right, it's just that the `169.254.0.0/16` network might be in use more widely and collisions might happen more often.

But, given the fact that I'm connected with two instances of _openfortivpn_  right now, which both use the `192.0.2.1`, I agree that the situation won't be worse with `169.254.2.1`.

dlenski:
The address you're discussing is the address of _the remote end_ of a _tunnel_ (that is, a point-to-point connection). It's utterly meaningless and a correct implementation of a point-to-point connection (such as a VPN tunnel) shouldn't attempt to route the "address of the remote end of the tunnel" in any way. :man_shrugging:

`169.254.0.0/16` are link-local-addresses anyway, which means that they're [unforwardable](https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml#note1). Most likely Forti and other VPN providers have chosen `169.254.0.0/16` addresses as a crutch for their buggy implementations, because operating systems won't attempt to traffic to these addresses.

In all the protocols supported by OpenConnect, including in the in-development support for Fortinet… we just ignore "remote end of a tunnel" addresses that we receive, and it works just fine.

DimitriPapadopoulos:
@dlenski Yet `pppd` requires a remote IP address, and it has been reported `192.0.2.1` → `169.254.2.1` does fix macOS slow VPN tunnel speeds. I am not 100% sure about that, as I have not reproduced this myself it has been reported only once. I don't know whether it's a `pppd`issue or a problem with FortiOS at the remote end.

dlenski:
> Yet it looks like pppd requires a remote IP address

If you don't give pppd a remote address, it will attempt to negotiate one using IPCP:

```
       <local_IP_address>:<remote_IP_address>
              Set the local and/or remote interface IP addresses.  Either one may be omitted.  The IP addresses can be specified with a host  name  or  in
              decimal dot notation (e.g. 150.234.56.78).  The default local address is the (first) IP address of the system (unless the noipdefault option
              is given).  The remote address will be obtained from the peer if not specified in any option.  Thus, in simple cases,  this  option  is  not
              required.   If  a  local and/or remote IP address is specified with this option, pppd will not accept a different value from the peer in the
              IPCP negotiation, unless the ipcp-accept-local and/or ipcp-accept-remote options are given, respectively.
```

I've verified that at least one Fortinet server is willing to do this negotiation (using [`ppp_protocols` branch](https://gitlab.com/openconnect/openconnect/-/commits/ppp_protocols) of OpenConnect).

> … and it has been reported `192.0.2.1` → `169.254.2.1` does fix macOS slow VPN speeds.

Yeah, this is probably because <something> is incorrectly trying to set up a route the peer's address. So setting it to a non-forwardable address definitely works as a bandaid. Reusing the same peer address `169.254.2.1` for multiple connections _should not matter_, because no conforming OS should be trying to send any IP packets to that address, anyway :-)

DimitriPapadopoulos:
I have verified that omitting the remote IP address (`:169.254.2.1` → `:`) does work.

Yet if I understand you correctly, the proper way to fix this is to stop forking `pppd` and embed minimal PPP code in _openfortivpn_ instead (#650).

Issue #791, opened 2020-10-15
dissociate authentication - 


DimitriPapadopoulos:
We should probably dissociate these two steps:
1. Authentication, users send credentials to the portal, receive a cookie.
2. The above cookie can be used to restart the tunnel multiple times.

Step 1 should not need root privileges as far as I can see but often requires user interaction.
Step 2 can run in the background or even be implemented as a daemon. Root privileges are required to modify routing and DNS parameters, unless perhaps interfaced with systemd and/or NetworkManager.

mrbaseman:
Sounds good to me.
Maybe we should start one or more separate branches for larger changeslike this.
Moving the routing to an external script (which might be a basis for much easier implementation of ipv6 support) would be another piece which I would see for a 2.0 release. 
The separation of authentication and tunnel establishment might not be that big change, but it may go well along with the other more fundamental changes.

Issue #594, opened 2020-03-23
run in background ?


txsastre:
hi there, thanks for this marvellous tool

I'm trying to do a lightweight distro for our users, and this is going to be very helpful.

but I'm trying to start a tunnel, and it works, but it does not "leave" the console it stays in foreground saying 
INFO:   Tunnel is up and running.

but I would like to be able to start and RDP connection, but it does not start because it is "stopped" by the openfortivpn 

may be I'm missing something but I don't find it !

thanks

DimitriPapadopoulos:
Indeed _openfortivpn_ is not a "daemon" or service that you start and forget. It's a program launched from the command line and you want it running in a console so that you can stop it with Ctrl+C. That's by design. As an alternative you may use _NetworkManager-fortisslvpn_ available on many Linux distributions.

Not sure what you mean about the RDP connection being "stopped" by _openfortivpn_. Are you writing  a script and do you call `openfortivpn` and the some other command so start RDP but that command is not reached because the script gets stock on the `openfortivpn` command until it exits? If so that's indeed teh case. It's up to you write a more complex script to handle that.

txsastre:
hi there.

well, the idea was to create a debian distro (with no desktop) for users who have problem with their computers or they are to old, to stablish an vpn connection and then open an RDP conection against our servers.

the concept is :
- start debian with no desktop
- execute bash
------- openfortivpn -c my_config 
------- rdesktop 192.168.1.1

but when executing "openfortivpn -c my_config " the bash script does not continue to the rdesktop connection, because I can see on the screen a successful vpn.

hope I have explained properly my problem.

Thanks

txsastre:
hi again
I've used this solution : 
sudo openfortivpn -c my_config **> program.log 2>&1**

DimitriPapadopoulos:
Yes, as long as you don't need to enter a password or other identifiers interactively, you may launch `openfortivpn` in the background yourself. By the way, you may have forgotten an extra **`&`**:
* `sudo openfortivpn -c my_config` starts openfortivpn with all parameters in a configuration file
* `> program.log 2>&1` redirects _stdout_ and _stderr_ to `program.log` (perhaps `openfortivpn.log` is better by the way)
* you probably need a final `&` to start in the background.

That said, how do you stop _openfortivpn_ (never?) and how do you restart it if needed (reboot?)?

txsastre:
hi, I have some issues making it work in background but finally I get it working.
[...]

Issue #93, opened 2017-01-23
Run as system daemon


nidico:
What's the best practice in order to run openfortivpn as a system daemon, e.g. as a systemd unit? I'm particularly concerned about automatic reconnection in case of connection loss.

adrienverge:
Hi @nidico,

You need to create a systemd config file (either in `/etc/systemd/system` or `/usr/lib/systemd/system`), for example `openfortivpn.service`. I'm not a systemd expert but I guess you would need:
```systemd
[Install]
After=NetworkManager.service
```

Once done:
```
sudo systemctl daemon-reload
sudo systemctl start openfortivpn
sudo systemctl enable openfortivpn
```

nidico:
For others looking at this: This isn't working (i.e. automatically restarting) for me as it is (though a similar systemd unit in general is). I haven't spend more time since and setup a cron job which restarts the systemd service if the connection is down.

r3tr0g4m3r:
This is my service running in CentOS 7.4:
/usr/lib/systemd/system/openfortivpn.service

``
Description = OpenFortiVPN
After=network.target
Documentation=man:openfortivpn(1)

[Service]
Type=simple
ExecStart = /usr/local/bin/openfortivpn -c /etc/openfortivpn/config
KillSignal=SIGTERM

[Install]
WantedBy=multi-user.target

deevictor:
> /etc/openfortivpn/config

Type=forking , otherwise systemd will show active status even if VPN connection is not established.

DimitriPapadopoulos:
Type=forking` doesn't look right here, because openfortivpn does not fork. Doesn't `Type=simple` work for you?

deevictor:
my bad, the service is not start with forking or simple type, only with Type=idle as mentioned r3tr0g4m3r.

DimitriPapadopoulos:
Type `notify` might be useful too as _openfortivpn_ calls sd_notify() to notify systems when the tunnel is up.

iKlsR:
@DimitriPapadopoulos I was setting up this and when I switch to `notify` it gets stuck in this state until it fails, going back to `simple` works.

```
ploi@data-server:~/data.redacted.com/itch$ systemctl status openfortivpn.service
● openfortivpn.service - Forti VPN Api Service
   Loaded: loaded (/etc/systemd/system/openfortivpn.service; enabled; vendor preset: enabled)
[...]

PR #835, opened 2021-01-22
Added daemon run mode

mobrembski:
As title says, adding daemon mode to run tunnel in background, like most servers do.


DimitriPapadopoulos:
There have already been discussion about this specific issue: #93, #594, #791. Does this pull request add anything new or address #791?

mobrembski:
> There have already been discussion about this specific issue: #93, #594, #791. Does this pull request add anything new or address #791?

I haven't checked issues. But it looks like it addresses most from #791 ;)

DimitriPapadopoulos:
Why not use _systemd_ instead?

mobrembski:
> Why not use _systemd_ instead?

Because our system is pure embedded one without systemd and with read-only fs ;)

adrienverge:
Hello @mobrembski, why not writing a small wrapper on your system, to daemonize openfortivpn?

    nohup openfortivpn 0<&- &>/dev/null &

Almost all systems have their services/daemons manager (systemd is an example), so a deamon mode in openfortivpn wouldn't be used much, and would probably too basic for most uses (auto-restart on crash, etc.)

mobrembski:
Well, maybe it's a good idea :) i will have a look :) thanks!

DimitriPapadopoulos:
@mobrembski Indeed we'd rather not add functionality in _openfortivpn_ if the same can be achieved by a small script. That said, if you feel changes in _openfortivpn_ are required to obtain extra functionality, we're happy to hear about it.

charlietomo:
This is a feature that I would find very useful and it is included in other command line VPN tools we use. That said, we would be open to using a "small script" if that was straightforward. However, across all the tickets mentioned, I have struggled to understand and get any working as expected.

@adrienverge are you able to expand on the `nohup openfortivpn 0<&- &>/dev/null &` command you have posted? For example, where do I reference my .conf file?

Other implementations I have seen have for example a `-b` command to send to background. It is simple to use and works as we expect.

DimitriPapadopoulos:
@charlietomo The conf file question seems orthogonal to sending _openfortivn_ to the background. You need to read about the command line options.

charlietomo:
> @charlietomo The conf file question seems orthogonal to sending _openfortivn_ to the background. You need to read about the command line options.

I have read the documentation and prefer to use a `.conf` file as, in my setup, it is more secure than having all the credentials in the command. I am aware I may be missing something, but have not found that missing information from my searching and reading so far. For example, in the `nohup openfortivpn 0<&- &>/dev/null &` command I am not aware of what the `0` does.

It seems I am not the only one expecting this feature and struggling to make it work through other means, although I understand it is not a feature that wants to be included. If you have any time to write a brief explanation of your command that would be very much appreciated.

DimitriPapadopoulos:
You still need to read the command line options to learn how to reference your conf file.

As for what `0` means, this is specific to the shell, not to _openfortivpn_. If it works, you don't need to be fully aware of what it does. If you do want to be aware of what it does, you need to read about the shell. Yet I agree it would be great to document how to run _openfortivpn_ in the background in more detail, for example in the wiki or even as an example in the man page. Speaking of wiki, you will find a [starting _openfortivpn_ as a system service](https://github.com/adrienverge/openfortivpn/wiki#starting-openfortivpn-as-a-system-service) paragraph there.

adrienverge:
@charlietomo I'm not sure what the `0` does in `nohup openfortivpn 0<&- &>/dev/null &`. I found this command on a well-known search engine, while looking for a solution for you. Maybe you could look it out?

charlietomo:
Apologies I think I may not have worded my question well and misunderstood @DimitriPapadopoulos reply re "The conf file question seems orthogonal to sending openfortivn to the background."

I have read the documentation and have multiple successfully working config files. What I do not have is a successful "send to background" mode, which is what this thread was started about (not by me).

@DimitriPapadopoulos thank you for the link to the wiki. That may be worth linking to from the `README.md` file. It may also be worth noting in the README.md to check the man page for additional options. I was unaware of the two additional (wiki and man) pages - despite reading the README.md

@adrienverge I'm not the original poster and only commented after you detailed the command, which I assumed you may have used or had knowledge of, which you might have been able to share. I have searched and used the various comments in other threads, but to date have not been successful. I will go back to searching and experimenting.

mrbaseman:
> ```
> nohup openfortivpn 0<&- &>/dev/null &
> ```
the `0<&-` closes the input file descriptor (see `man bash` for instance).

to supply a config file just put the usual `-c configfile` between `openfortivpn` and these redirects. 

For completeness: `&>/dev/null` redirects both stdout and stderr to /dev/null, but if you omit that, the `nohup` command will redirect them to a nohup.out file in the current directory ;)

charlietomo:
> > ```
> > nohup openfortivpn 0<&- &>/dev/null &
> > ```
> 
> 
> the `0<&-` closes the input file descriptor (see `man bash` for instance).
> 
> 
> to supply a config file just put the usual `-c configfile` between `openfortivpn` and these redirects (see `man openfortivpn`)
> 
> 
> For completeness: `&>/dev/null` redirects both stdout and stderr to /dev/null, but if you omit that, the `nohup` command will redirect them to a nohup.out file in the current directory ;)

Many thanks. I will experiment with this whilst reading further, but your explanation was very useful.

decke:
nohup and systemd are not a replacement for background mode in some cases. As soon as 2FA is enabled it needs to stay in foreground until the user has entered the token and then it can daemonize.

This is also how --daemon in openvpn works.

DimitriPapadopoulos:
True, but see https://github.com/adrienverge/openfortivpn/issues/791. Meanwhile 2FA is not implemented.

decke:
I have a connection where I either get the token via EMail or SMS so 2FA is definitely working - I use it all the time.

DimitriPapadopoulos:
Yes, SAML is not working, sorry. Still it's best implemented together with a patch for #791.

debMan:
Any update on this PR?

PR #841, opened 2021-02-05
Add a space at the end of the OTP prompt

DimitriPapadopoulos:
Other prompts do have that final space:
```
	"VPN account password: "
	"Two-factor authentication token: "
```
@mrbaseman I cannot test myself, but this makes sense, doesn't it?


mrbaseman (approved):

mrbaseman:
just a thought: some users who automatize the tunnel using `expect` for instance might have to adapt their scripts... but well... the format of such prompts may change from time to time :)

PR #842, opened 2021-02-07
Query for the PEM pass phrase only once

DimitriPapadopoulos:
OpenSSL queries for the PEM pass phrase whenever we call ssl_connect().
Instead, openfortivpn will query for the PEM password only once, if needed, from within an OpenSSL callback.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -136,6 +136,7 @@ PPPD_USAGE \
 "  --user-cert=pkcs11:           Use smartcard. Takes also partial or full PKCS11-URI.\n" \
 "  --user-key=<file>             Use specified PEM-encoded key if the server requires\n" \
 "                                authentication with a certificate.\n" \
+"  --pem-passphrase=<pass>       Pass phrase for the PEM-encoded key.\n" \
 "  --use-syslog                  Log to syslog instead of terminal.\n" \
 "  --trusted-cert=<digest>       Trust a given gateway. If classical SSL\n" \
 "                                certificate validation fails, the gateway\n" \
@@ -194,6 +195,7 @@ int main(int argc, char **argv)
 		.gateway_port = 443,
 		.username = {'\0'},
 		.password = {'\0'},
+		.password_set = 0,
 		.otp = {'\0'},
 		.otp_prompt = NULL,
 		.otp_delay = 0,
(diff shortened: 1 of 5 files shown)


DimitriPapadopoulos:
@mrbaseman @alfem Should work now. I had forgotten to update the OpenSSL buffer in addition to the new `cfg->user_key_passphrase`.

Still need to add support for the user key pass phrase in the config file and command line options. WIll do that once querying interactively for the password works.

alfem:
It works!

DimitriPapadopoulos:
Then I just need to add the relevant command line and configuration file options. Stay tuned.

src/main.c
```diff
@@ -392,6 +397,15 @@ int main(int argc, char **argv)
 				cli_cfg.user_key = strdup(optarg);
 				break;
 			}
+			if (strcmp(long_options[option_index].name,
+			           "user-key-passphrase") == 0) {
+				cli_cfg.user_key_passphrase = strdup(optarg);
```
mrbaseman, line 402: line 402 shouldn't be there. Without that it works. line 403 is the correct one

mrbaseman: I think the intention of `strdup()` on L. 402 is to allocate the buffer that is passed to `strncpy()` on L. 403. 
Unfortunately, the code doesn't compile 
```
src/main.c: In function ‘main’:
src/main.c:402:33: error: assignment to expression with array type
     cli_cfg.user_key_passphrase = strdup(optarg);
                                 ^
Makefile:620: recipe for target 'src/openfortivpn-main.o' failed
make: *** [src/openfortivpn-main.o] Error 1
```
vpn_config uses static arrays, so allocation of a fresh buffer isn't needed. We can just copy the string into cli_cfg.user_key_passphrase

DimitriPapadopoulos:
@mrbaseman What about the variable names? Is `user_key_passphrase` OK? Should I change to `pem_passphrase` for example? And what about the current option name `user-key-passphrase`?

src/main.c
mrbaseman: the current version works now, including the command line option

mrbaseman:
Maybe `pem_passphrase` is better because it is shorter.

mrbaseman:
Hmm... Travis CI seems to be stuck. But I have just tested this again (on linux). It works nicely. 
Thank you very much, @DimitriPapadopoulos for implementing this feature. 
I have been working around this for years, and somehow didn't get the kick to look deeper into the problem. 
It's a real benefit for _openfortivpn_ having this fixed finally. Thank you.

alfem:
I would like to join in congratulating @DimitriPapadopoulos for the development of this long-awaited feature. 

Thanks a lot!

adrienverge:
Travis seems blocked, I'll merge this manually!

PR #849, opened 2021-02-17
Prefer SSL_CTX_use_certificate_chain_file

goug71:
To honor bundle of chained client certificates, prefering  SSL_CTX_use_certificate_chain_file over SSL_CTX_use_certificate_file could come handy, as stated in https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_use_certificate_file.html.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -1176,10 +1176,9 @@ int ssl_connect(struct tunnel *tunnel)
 #endif
 
 		if (tunnel->config->user_cert) {
-			if (!SSL_CTX_use_certificate_file(
-			            tunnel->ssl_context, tunnel->config->user_cert,
-			            SSL_FILETYPE_PEM)) {
-				log_error("SSL_CTX_use_certificate_file: %s\n",
+			if (!SSL_CTX_use_certificate_chain_file(
+			            tunnel->ssl_context, tunnel->config->user_cert)) {
+				log_error("SSL_CTX_use_certificate_chain_file: %s\n",
 				          ERR_error_string(ERR_peek_last_error(), NULL));
 				return 1;
 			}


src/tunnel.c
```diff
@@ -1176,9 +1176,8 @@ int ssl_connect(struct tunnel *tunnel)
 #endif
 
 		if (tunnel->config->user_cert) {
-			if (!SSL_CTX_use_certificate_file(
-			            tunnel->ssl_context, tunnel->config->user_cert,
-			            SSL_FILETYPE_PEM)) {
+			if (!SSL_CTX_use_certificate_chain_file(
+			            tunnel->ssl_context, tunnel->config->user_cert)) {
 				log_error("SSL_CTX_use_certificate_file: %s\n",
```
DimitriPapadopoulos, line 1181: Please modify the error message athe same way.

DimitriPapadopoulos (commented):
Indeed, the man page reads:
> See the NOTES section on why SSL_CTX_use_certificate_chain_file() should be preferred.

> SSL_CTX_use_certificate_chain_file() adds the first certificate found in the file to the certificate store. The other certificates are added to the store of chain certificates using [SSL_CTX_add1_chain_cert(3)](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_add1_chain_cert.html). Note: versions of OpenSSL before 1.0.2 only had a single certificate chain store for all certificate types, OpenSSL 1.0.2 and later have a separate chain store for each type. SSL_CTX_use_certificate_chain_file() should be used instead of the SSL_CTX_use_certificate_file() function in order to allow the use of complete certificate chains even when no trusted CA storage is used or when the CA issuing the certificate shall not be added to the trusted CA storage.

I am just a bit worried by this part:
> SSL_CTX_use_certificate_chain_file() loads a certificate chain from **file** into **ctx**. The certificates must be in PEM format and must be sorted starting with the subject's certificate (actual client or server certificate), followed by intermediate CA certificates if applicable, and ending at the highest level (root) CA. SSL_use_certificate_chain_file() is similar except it loads the certificate chain into **ssl**.

It's not clear whether the certificates must always "end at the highest level (root) CA" or should "end at the highest level (root) CA" only when applicable. I guess the latter is correct, isn't it?

src/tunnel.c
goug71: Commited as req

goug71:
> Indeed, the man page reads:
> 
> > See the NOTES section on why SSL_CTX_use_certificate_chain_file() should be preferred.
> 
> > SSL_CTX_use_certificate_chain_file() adds the first certificate found in the file to the certificate store. The other certificates are added to the store of chain certificates using [SSL_CTX_add1_chain_cert(3)](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_add1_chain_cert.html). Note: versions of OpenSSL before 1.0.2 only had a single certificate chain store for all certificate types, OpenSSL 1.0.2 and later have a separate chain store for each type. SSL_CTX_use_certificate_chain_file() should be used instead of the SSL_CTX_use_certificate_file() function in order to allow the use of complete certificate chains even when no trusted CA storage is used or when the CA issuing the certificate shall not be added to the trusted CA storage.
> 
> I am just a bit worried by this part:
> 
> > SSL_CTX_use_certificate_chain_file() loads a certificate chain from **file** into **ctx**. The certificates must be in PEM format and must be sorted starting with the subject's certificate (actual client or server certificate), followed by intermediate CA certificates if applicable, and ending at the highest level (root) CA. SSL_use_certificate_chain_file() is similar except it loads the certificate chain into **ssl**.
> 
> It's not clear whether the certificates must always "end at the highest level (root) CA" or should "end at the highest level (root) CA" only when applicable. I guess the latter is correct, isn't it?

Well, as far as I could test (many many times with many other products), you just need to append the intermediate CAs in between leaf Cert and the party's trusted root, which somehow makes sense and is just the way it is with https. The 'proper' way to do this would be to split CA chain in another file, just like apache or openvpn do, but this 'bundled' format is known to be working most of the time...

goug71:
Just tested the trivial case with this patch (single client certificate in user-cert file, which is issued by gateway's trusted CA), and it works just the same.

PR #851, opened 2021-02-18
Travis CI → GitHub Actions

DimitriPapadopoulos:
Travis CI has kept failing in the last few years, hopefully GitHub Actions will prove more robust.

Here is how the checks look like under GitHub Actions:
github_actions
I have put each check in its own job because jobs run in parallel.

Please help me improve the workflow and job names: _Legacy_ and the rest.


.github/workflows/openfortivpn.yml
```diff
@@ -0,0 +1,75 @@
+---
+name: Legacy
+
+on:
+  push:
+    branches-ignore: [master]
+    # Remove the line above to run when pushing to master
```
adrienverge, line 7: Why not running tests when pushing commits to `master`? It can be a good information to know whether tests pass or not, after adding a commit manually.

.github/workflows/openfortivpn.yml
```diff
@@ -0,0 +1,75 @@
...
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Install OpenSSL
+        run: ./tests/ci/install_openssl.sh $HOME/.openfortivpn-deps
+
+      - name: Build openfortivpn
+        run: |-
```
adrienverge, line 71: You can use the standard form:
```suggestion
        run: |
```
The only difference is that `|-` will remove `\n` after the last line (`make\n`), which isn't useful in my opinion.

adrienverge (commented):
Nice! Thanks a lot @DimitriPapadopoulos :+1:

> I have put each check in its own job because jobs run in parallel.

OK! Personnally I would have merged `astyle`, `eol-at-eof` and `line_length` (+ maybe `checkpatch`?) into a `style` or `lint` job, to have a less resource-comsuming and more straightforward feedback on GitHub. But I get the "parallel" point, so as you prefer.

> Please help me improve the workflow and job names: Legacy and the rest.

What about using "Tests" instead of "Legacy"?

.github/workflows/openfortivpn.yml
DimitriPapadopoulos: I don't know, just re-used Github's Super-Linter file. Perhaps because we're not supposed to push directly to master in the first place?

adrienverge: We're not *supposed* to, but it could happen on exceptional circumstances :)
In such situations, why not doing the default (run tests) like for other branches?

DimitriPapadopoulos:
> > I have put each check in its own job because jobs run in parallel.
> 
> OK! Personnally I would have merged `astyle`, `eol-at-eof` and `line_length` (+ maybe `checkpatch`?) into a `style` or `lint` job, to have a less resource-comsuming and more straightforward feedback on GitHub. But I get the "parallel" point, so as you prefer.

Since the `astyle`, `eol-at-eof`, `line_length` and `checkpatch` checks are much faster than the build task, it makes sense to merge them into a single job:
* Parallelism doesn't help in any way.
* The YAML file is much lighter.
* The output is more straightforward, I agree.

.github/workflows/openfortivpn.yml
DimitriPapadopoulos: My guess is that force pushing to master happens only when we want to bypass the tests - as is currently the case with Travis CI.

DimitriPapadopoulos:
Sceen capture after updates:
github_actions

DimitriPapadopoulos:
I think it's ready to merge. If you're happy with it, please force merge into master.

mrbaseman (commented):
@DimitriPapadopoulos thank you for taking the time to work on this.

About the names of the checks, maybe expand a bit here: 
"Linux Kernel checkpatch.pl" -> "Linux Kernel Coding Style checkpatch.pl"

Apart from that, if it helps fixing the trouble with the automated checks, it's a good move forward.
If it doesn't help, well, it's at least worth trying. We will see if github workflows run more reliably. I hope so.

adrienverge:
I agree with @mrbaseman, it's a good move. Thanks for taking care of this!

Since all remarks are addressed, I'll merge and try to enable GitHub Actions on the repo.

.github/workflows/openfortivpn.yml
DimitriPapadopoulos: @adrienverge Now I know why we should ignore the `master` branch: we end up running the tests twice on every pull request, when submitting the pull requests and when merging the pull request into the `master` branch. That's a waste of resources.

I cannot find a way to distinguish between pushing a pull request to `master` and an exceptional direct push to `master`. I plan on restoring `branches-ignore: [master]`.

PR #855, opened 2021-02-18
Support Coverity Scan from GitHub Actions

DimitriPapadopoulos:
Inspired by:
https://github.com/ruby/actions-coverity-scan

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -1,0 +1,43 @@
+---
+name: Synopsys
+
+on:
+  push:
+    branches: [coverity_scan]
+
+jobs:
+  coverity-scan:
+    name: Coverity Scan
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Download the Coverity Scan Build Tool
+        run: |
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
It works on my own [`coverity_scan`](https://github.com/DimitriPapadopoulos/openfortivpn/tree/coverity_scan) branch.

We can test against the main repository once the secret token `COVERITY_SCAN_TOKEN` has been set.

.github/workflows/coverity-scan.yml
```diff
@@ -0,0 +1,43 @@
+---
+name: Synopsys
+
+on:
+  push:
+    branches: [coverity_scan]
```
adrienverge, line 6: Is it a leftover?

adrienverge (commented):
`COVERITY_SCAN_TOKEN` has been set! :slightly_smiling_face:

.github/workflows/coverity-scan.yml
DimitriPapadopoulos: We are limited to a few scans per week/day:
[What is the frequency for build submissions to Coverity Scan?](https://scan.coverity.com/faq#frequency)

We have been checking only the `coverity_scan` branch, which we update manually.

adrienverge: Ah OK, I missed that. Perfect then!

PR #858, opened 2021-02-19
Ignore pushes to master branch

DimitriPapadopoulos:
Otherwise, we end up running tests twice with each pull request, first when submitting the pull request, then when pushing the pull request into master.

### DIFF
--- a/.github/workflows/openfortivpn.yml
+++ b/.github/workflows/openfortivpn.yml
@@ -3,6 +3,8 @@ name: Tests
 
 on:
   push:
+    branches-ignore: [master]
+    # Remove the line above to run when pushing to master
 
   pull_request:
     branches: [master]


adrienverge:
I don't agree with this one (for reference, [discussion started here](https://github.com/adrienverge/openfortivpn/pull/851#discussion_r578231245)): in my opinion it's useful to have tests rerun on each commit pushed to `master`, even if a previous test was run for that commit on a preceding pull request. The only drawback is that it uses more resources. Advantages are: tests are run when pushing commits without a pull request; it allows checking that everything is OK on `master`; it allows comparing PRs status with `master` status (e.g. if Coverity tests start failing: is it due to something on `master`, or to what the PR changes?). Moreover, most repositories do the same: run tests on `master` (this is probably why it's the default behavior in GitHub Actions).
What do you think?

DimitriPapadopoulos:
OK, I'll revert that, there are indeed many benefits for a single drawback.

For the record, here is the initial rationale for this PR:
* [GitHub's own Super-Linter action suggests](https://github.com/github/super-linter#example-connecting-github-action-workflow):
  ```yaml
  #############################
  # Start the job on all push #
  #############################
  on:
    push:
      branches-ignore: [master]
      # Remove the line above to run when pushing to master
    pull_request:
      branches: [master]
  ```
* I cannot find a default behaviour for GitHub Actions. The **`on`** element is required, there are no defaults as far as I can see:
  [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on)

adrienverge:
> I cannot find a default behaviour for GitHub Actions. The on element is required, there are no defaults as far as I can see

OK, I didn't know that. This means that removing the key `push` (that is empty) would prevent all builds, on all branches, to be executed when pushing a commit?

DimitriPapadopoulos:
Yes, I think so. I must admit I haven't tried a lone **`on:`** in the workflow. But the documentation states:
> **Required**. The name of the GitHub event that triggers the workflow. You can provide a single event `string`, `array` of events, `array` of event `types`, or an event configuration `map` that schedules a workflow or restricts the execution of a workflow to specific files, tags, or branch changes.

PR #860, opened 2021-02-19
Consistency in CFLAGS / CPPFLAGS

DimitriPapadopoulos:


DimitriPapadopoulos:
While I can see `OPENSSL_CFLAGS` and `LIBSYSTEMD_CFLAGS` might contain an **`-I`** option to specify a directory to be searched for header files during preprocessing, I still believe `*_CFLAGS` should go into `CFLAGS`, not `CPPFLAGS`, for reasons of consistency. Besides the preprocessor is never run without the compiler in most builds.

mrbaseman (approved):
sounds logical to me

PR #864, opened 2021-03-03
Add repology link

ehaupt:
Add repology badges to display the packaging status in all known repositories. It's nice to compare versions and provides easy browseable access to each repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,10 @@ or [install MacPorts](https://www.macports.org/install.php) then install openfor
 sudo port install openfortivpn
 ```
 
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
The list is quite long. Perhaps it would be better to have this information on a different page? For example in the GitHub wiki, with a link from `README.md`?

mrbaseman:
...or we have the link in `README.md` but without the SVG image (which we can include on the wiki imo.)

ehaupt:
Perhaps linking to https://repology.org/project/openfortivpn/versions

PR #866, opened 2021-03-11
SSLv23_client_method() → TLS_client_method()

DimitriPapadopoulos:
From the man page:
	SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()
	Use of these functions is deprecated. They have been replaced with
	the above TLS_method(), TLS_server_method() and TLS_client_method()
	respectively. New code should use those functions instead.


mrbaseman (approved):
this looks good to me

PR #874, opened 2021-03-26
Additional sources of information

DimitriPapadopoulos:


mrbaseman (approved):
yes, some issues (like the possibilities of pppd ip-up scripts) were solved by pointing the users to the wiki. A hint in the man page can't hurt.

PR #875, opened 2021-03-26
CONFIG FILES → CONFIGURATION

DimitriPapadopoulos:
According to man-pages(7), conventional man page section names include CONFIGURATION but not CONFIG FILES. On the other hand CONFIGURATION is `Normally only in Section 4` _Special files (devices)_. Still, I prefer standard names.


adrienverge (approved):
Makes sense. Let's use the standards. Thanks @DimitriPapadopoulos!

PR #879, opened 2021-03-30
Remove unused functions from AC_CHECK_FUNCS

DimitriPapadopoulos:
* While the _malloc_() and _realloc_() functions are being called, the _AC_FUNC_MALLOC_ and _AC_FUNC_REALLOC_ macros have already been checking if these functions are available **and** compatible with the GNU C library functions.
* Also, _optarg_ and _optind_ are not functions - they are documented to be external variables.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -120,7 +120,6 @@ fileno \
 fopen \
 forkpty \
 fprintf \
-fputc \
 fputs \
 fread \
 free \
@@ -142,7 +141,6 @@ ioctl \
 isatty \
 isdigit \
 isspace \
-malloc \
 memcmp \
 memcpy \
 memmem \
@@ -151,8 +149,6 @@ memset \
 ntohs \
 open \
 openlog \
-optarg \
-optind \
 pclose \
 popen \
 printf \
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Note that the _sem__* functions are not used by the mac0S implementation in `src/io.c`:
```c
#if HAVE_MACH_MACH_H
/* this is typical for mach kernel used on Mac OS X */
#include <mach/mach.h>

/* Mac OS X defines sem_init but actually does not implement them */
typedef semaphore_t os_semaphore_t;

#define SEM_INIT(sem, x, value)	semaphore_create(mach_task_self(), sem, \
								SYNC_POLICY_FIFO, value)
#define SEM_WAIT(sem)			semaphore_wait(*(sem))
#define SEM_POST(sem)			semaphore_signal(*(sem))
#define SEM_DESTROY(sem)		semaphore_destroy(mach_task_self(), *(sem))

#else

#include <semaphore.h>

typedef sem_t os_semaphore_t;

#define SEM_INIT(sem, x, value)	sem_init(sem, x, value)
#define SEM_WAIT(sem)			sem_wait(sem)
#define SEM_POST(sem)			sem_post(sem)
#define SEM_DESTROY(sem)		sem_destroy(sem)

#endif
```
Therefore, we should remove them from the _AC_CHECK_FUNCS_ list, at least conditionally on the platform. It would be great to have off-the-shelf specialized macros for _sem__* functions, but I haven't found any.

Anyway, I'll create a specific issue for the _sem__* functions.

PR #890, opened 2021-04-26
Different hint for each different connection

### LINKED ISSUE #886: Hint given to `pinentry_read_password` is too generic thus over writing secrets
In `userinput.c` the function `read_password` takes a `hint` parameter which is only used by `pinentry_read_password` to distinguish between secrets.

Unfortunately this `hint` is too generic.

`read_password` is called with a `hint` value of `otp`, `2fa` and `password`, and these are used by `pinentry` to store the secret securely in e.g. Keychain on macOS and similar on other OS'es.

Pinentry can only be used if a username is specified, and using different usernames for different vpn endpoints will either over write the stored secret or use a wrong secret each time you call with a different username.

I suggest that hint is changed to include the username, so the hint becomes `<username>-otp`, `<username>-2fa` and `<username>-password` (where `<username>` is the actual username specified by the user).

In addition to this there appears be bug in the way `SETKEYINFO` is called as the account value in my macOS Keychain is `ssword` and not `password` as the hint specifies.

DimitriPapadopoulos:
Fixes #886.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -613,8 +613,12 @@ int main(int argc, char **argv)
 		}
 	// If username but no password given, interactively ask user
 	if (!cfg.password_set && cfg.username[0] != '\0') {
-		read_password(cfg.pinentry,
-		              "password", "VPN account password: ",
+		char hint[USERNAME_SIZE + 1 + REALM_SIZE + 1 + GATEWAY_HOST_SIZE + 10];
+
+		sprintf(hint, "%s_%s_%s_password",
+		        cfg.username, cfg.realm, cfg.gateway_host);
+		read_password(cfg.pinentry, hint,
+		              "VPN account password: ",
 		              cfg.password, PASSWORD_SIZE);
 	}
 	log_debug("Configuration host = \"%s\"\n", cfg.gateway_host);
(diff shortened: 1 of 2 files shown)


src/main.c
```diff
@@ -613,8 +613,12 @@ int main(int argc, char **argv)
 		}
 	// If username but no password given, interactively ask user
 	if (!cfg.password_set && cfg.username[0] != '\0') {
-		read_password(cfg.pinentry,
-		              "password", "VPN account password: ",
+		char hint[USERNAME_SIZE + 1 + REALM_SIZE + 1 + GATEWAY_HOST_SIZE + 12];
```
mrbaseman, line 616: should be 10 here if I'm not mistaken

mrbaseman (approved):
I assume that entries already stored via pinentry loose their connection to openfortivpn, but well, be it... maybe this is something worth to mention in the changelog for the next release

appart from that I think this change makes sense and adds some value, especially for users who work with several vpns

src/main.c
DimitriPapadopoulos: Indeed, `"_password\0"` is 1 + 8 + 1 = 10 characters long.

DimitriPapadopoulos:
@mrbaseman Yes, existing entries are lost, users need to re-enter the password.

I don't see how to fix that without side effects: I would need to try any existing entry, possibly from a different server, on the server we connect to. This may leak information. What do you think?

mrbaseman:
It's probably not worth the effort to add extra code for the transistion (which wouldn't be used anymore after the first login after the update), but we should remember to mention it in the changelog

PR #901, opened 2021-06-03
Release allocated SSL resources

### LINKED ISSUE #462: socket: Too many open files
Hi,

Not sure if this is a local computer issue or something the forticlient. Every so often we see this in the log over and over, the only thing that helps is to kill it and restart:
Aug 19 21:14:11 n.e.org openfortivpn[947]: ERROR:  socket: Too many open files
Aug 19 21:14:11 n.e.org openfortivpn[947]: INFO:   Could not log out.
Aug 19 21:15:10 n.e.org openfortivpn[947]: ERROR:  socket: Too many open files
Aug 19 21:15:10 n.e.org openfortivpn[947]: INFO:   Closed connection to gateway.
Aug 19 21:15:10 n.e.org openfortivpn[947]: ERROR:  socket: Too many open files
Aug 19 21:15:10 n.e.org openfortivpn[947]: INFO:   Could not log out.
Aug 19 21:16:09 n.e.org openfortivpn[947]: ERROR:  socket: Too many open files
Aug 19 21:16:09 n.e.org openfortivpn[947]: INFO:   Closed connection to gateway.
Aug 19 21:16:09 n.e.org openfortivpn[947]: ERROR:  socket: Too many open files
Aug 19 21:16:09 n.e.org openfortivpn[947]: INFO:   Could not log out.

DimitriPapadopoulos:
The following resources seem to  always be allocated in this order:
```
	ssl_socket
	ssl_context
	ssl_handle
```
Release them in the reverse order:
```
	ssl_handle
	ssl_context
	ssl_socket
```

Hopefully helps to fix #462.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -327,6 +327,7 @@ static int pppd_run(struct tunnel *tunnel)
 
 		if (close(tunnel->ssl_socket))
 			log_warn("Could not close ssl socket (%s).\n", strerror(errno));
+		tunnel->ssl_socket = -1;
 		execv(pppd_args.data[0], (char *const *)pppd_args.data);
 		free(pppd_args.data);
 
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Unfortunately, _ssl_connect_() is called twice _run_tunnel_()! Because the rationale is not documented, I'm not certain this patch does the right thing. Are these resources supposed to have different lifetimes?

DimitriPapadopoulos:
Doesn't work for me. The lifetime of these SSL resources is intricate and had never been documented. Unfortunately, it appears to be incompatible with `--persistent`. The analysis of their undocumented lifetime after the fact is probably going to be too difficult to fix, at least for me with the limited time I can get into it.

DimitriPapadopoulos:
Another attempt at it. This new patch is less ambitious:
- We reset variables to `-1`/`NULL` after releasing the associated resources.
- We do release resources in case of error in _ssl_connect_().

The remaining anomaly is that _ppd_run_() releases `tunnel->ssl_socket` without releasing `tunnel->ssl_context` and `tunnel->ssl_handle`.

Not perfect, but certainly a step in the right direction.

PR #986, opened 2022-04-28
Add hooks for when link goes up and down

Joibel:
When the link has been established optionally run a supplied 'up-hook'
and when it goes down run 'down-hook'.

Allows simple interaction with other subsystems, in my case
reconfiguring proxies.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -442,6 +442,12 @@ int load_config(struct vpn_config *cfg, const char *filename)
 		} else if (strcmp(key, "check-virtual-desktop") == 0) {
 			free(cfg->check_virtual_desktop);
 			cfg->check_virtual_desktop = strdup(val);
+		} else if (strcmp(key, "up-hook") == 0) {
+			free(cfg->up_hook);
+			cfg->up_hook = strdup(val);
+		} else if (strcmp(key, "down-hook") == 0) {
+			free(cfg->down_hook);
+			cfg->down_hook = strdup(val);
 		} else {
 			log_warn("Bad key in configuration file: \"%s\".\n", key);
 			goto err_free;
@@ -600,4 +606,12 @@ void merge_config(struct vpn_config *dst, struct vpn_config *src)
 		dst->hostcheck = src->hostcheck;
 	if (src->check_virtual_desktop != invalid_cfg.check_virtual_desktop)
 		dst->check_virtual_desktop = src->check_virtual_desktop;
+	if (src->up_hook) {
+		free(dst->up_hook);
+		dst->up_hook = src->up_hook;
+	}
+	if (src->down_hook) {
+		free(dst->down_hook);
+		dst->down_hook = src->down_hook;
+	}
 }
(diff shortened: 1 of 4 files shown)


Tatsh:
>I'm using ppp's if-up and if-down scripts for this purpose.
How do you do that?

DimitriPapadopoulos:
> How do you do that?

https://github.com/adrienverge/openfortivpn/wiki#using-pppd-ip-up--ip-down-scripts

DimitriPapadopoulos:
That said, I think it might be worth setting routing and DNS in a separate script called by openfortivpn, ideally [vpnc-scripts](https://gitlab.com/openconnect/vpnc-scripts). When needed, it can be replaced by alternate scripts such as [vpn-slice](https://github.com/dlenski/vpn-slice) or a custom script. We would get IPv6 support and better DNS as a bonus.

I just don't have the time to change openfortivpn myself these days.

Joibel:
Yeah, if-up and if-down do this, so lets abandon this.

pschichtel:
@DimitriPapadopoulos I've read a bunch of the related issue discussion yesterday and those left me asking the same question. if there was a way for ppp ip-{up,down}.d scripts to fetch information about the tunnel (e.g. DNS and IP routes pushed by the server, could be as simple as having a file), then I don't think there would be any benefit to a solution in openfortivpn directly. I'm doing routing and DNS like this for a while now because I don't agree with a few things the server pushes. It would be need if I could filter over what the server tries to push instead of replicating the list.

currently I do this:

```bash
#!/bin/bash

routes=(...)
dns_servers=(...)
search_domains=(...)

resolvectl dns "$1" "${dns_servers[@]}"
resolvectl domain "$1" "${search_domains[@]}"

for route in "${routes[@]}"
do
    ip route add "$route" dev "$1"
done
```

DimitriPapadopoulos:
That's exactly what [vpnc-scripts](https://gitlab.com/openconnect/vpnc-scripts) does: get information DNS and IP routes pushed by the server from the calling program. There are multiple ways to achieve that:
* a file,
* IPC,
* environment variables as `vpnc-script` expects.

If I were to choose, I would choose the latter - to avoid re-inventing the wheel.

pschichtel:
@DimitriPapadopoulos I moved the comment to #678

Tatsh:
@pschichtel omitting the details, can you share your scripts that set up DNS with the pppd hooks?

pschichtel:
@Tatsh 

```bash
#!/bin/bash

routes=(...)
dns_servers=(...)
search_domains=(...)

resolvectl dns "$1" "${dns_servers[@]}"
resolvectl domain "$1" "${search_domains[@]}"

for route in "${routes[@]}"
do
    ip route add "$route" dev "$1"
done
```

Tatsh:
> @Tatsh that's my /etc/ppp/ip-up.d/* script
> 
> ```bash
> #!/bin/bash
> 
> routes=(...)
> dns_servers=(...)
> search_domains=(...)
> 
> resolvectl dns "$1" "${dns_servers[@]}"
> resolvectl domain "$1" "${search_domains[@]}"
> 
> for route in "${routes[@]}"
> do
>     ip route add "$route" dev "$1"
> done
> ```

So all your routes and DNS etc are static and known beforehand?

I would like to make the same kind of thing but I want to get it from the VPN.

PR #1009, opened 2022-08-29
fix: only try to set unique routes

EriksRemess:
Hi,

Currently **openfortivpn** tries to set duplicate routes and fails with "WARN:   Route to gateway exists already.".
These changes tries to filter out duplicates and only set unique routes.

### DIFF
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -888,10 +888,20 @@ static int ipv4_set_split_routes(struct tunnel *tunnel)
 	int i;
 
 	for (i = 0; i < tunnel->ipv4.split_routes; i++) {
-		struct rtentry *route;
-		int ret;
+		int j, ret;
+		struct rtentry *route = &tunnel->ipv4.split_rt[i];
 
-		route = &tunnel->ipv4.split_rt[i];
+		for (j = 0; j < i ; j++) {
+			struct rtentry *other_route = &tunnel->ipv4.split_rt[j];
+
+			if (route_dest(route).s_addr == route_dest(other_route).s_addr)
+				break;
+		}
+
+		// skip duplicate routes
+		if (i != j)
+			continue;
+
 		// check if the route to be added is not the one to the gateway itself
 		if (route_dest(route).s_addr == route_dest(&tunnel->ipv4.gtw_rt).s_addr) {
 			log_debug("Skipping route to tunnel gateway (%s).\n",


EriksRemess:
@adrienverge DimitriPapadopoulos can you have a look?

DimitriPapadopoulos:
Part of the code has been reformatted (tabs changed into spaces, spaces removed). Please follow the Linux kernel style.

About the rest, I haven't looked into the routing code for months (or even years) so I am unable to review in detail, but I think I get the general idea and the general idea looks good to me.

EriksRemess:
Made changes to the formatting.

EriksRemess:
Made changes to the formatting. Thanks for taking a look, @DimitriPapadopoulos!

DimitriPapadopoulos:
You can run `./tests/lint/run.sh` in your working tree.

EriksRemess:
yeah, done that and fixed also new issues with line length caused by the new indentation

src/ipv4.c
```diff
@@ -885,36 +885,46 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
...
-			route->rt_flags &= ~RTF_GATEWAY;
-		if (route_gtw(route).s_addr != 0)
-			route->rt_flags |= RTF_GATEWAY;
-		ret = ipv4_set_route(route);
-		if (ret == ERR_IPV4_SEE_ERRNO && errno == EEXIST)
-			log_warn("Route to gateway exists already.\n");
-		else if (ret != 0)
-			log_warn("Could not set route to tunnel gateway (%s).\n",
-			         err_ipv4_str(ret));
+		if (i == j) {
```
DimitriPapadopoulos, line 901: That would result in less code modifications and light be more readable:
```c
		// skip duplicate routes
		if (j != i)
			continue
```

EriksRemess: Good point! Fixed!

mrbaseman:
Looks good to me. This catches the case when the fortigate pushes duplicate routing entries. I haven't seen that yet, but there might be configurations that produce such duplicates. But there might be other cases, e.g. `pppd` with a `ifup`-script that adds routes which still could collide. Anyhow, it's a good move to handle duplicate entries in a better way.

src/ipv4.c
```diff
@@ -885,13 +885,23 @@ int ipv4_add_split_vpn_route(struct tunnel *tunnel, char *dest, char *mask,
 
 static int ipv4_set_split_routes(struct tunnel *tunnel)
 {
-	int i;
+	int i, j;
```
DimitriPapadopoulos, line 888: Perhaps move this declaration of `j` inside the `i` loop.

EriksRemess: done.

DimitriPapadopoulos:
@mrbaseman Looks good to me. Can I merge?

EriksRemess:
New release any time soon?

DimitriPapadopoulos:
I will revert this PR. Actually, openfortivpn does **not** fail as far as I can see. It just emits a warning. What is wrong with the warning?

Issue #362, opened 2018-08-22
ERROR: read: Input/output error


borisovg:
For some reason I started seeing this error yesterday when trying to connect:

```
DEBUG:  Got Address: 10.18.2.1
DEBUG:  if_config: not ready yet...
ERROR:  read: Input/output error
INFO:   Cancelling threads...
DEBUG:  Waiting for pppd to exit...
DEBUG:  waitpid: pppd exit status code 16
INFO:   pppd: The link was terminated by the modem hanging up.
INFO:   Terminated pppd.
INFO:   Closed connection to gateway.
```

Debian Sid, openfortivpn 1.7.1

Updated some packages yesterday so possible something came in and broke stuff.

Looks different to #154 so opened this new issue.

DimitriPapadopoulos:
This message does look strange:
```DEBUG:  if_config: not ready yet...```
The relevant code is:
https://github.com/adrienverge/openfortivpn/blob/a5ab08a/src/io.c#L540-L563
Clearly the PPP interface isn't up.

Now why PPP isn't up? You have provided the PPP log which is extremely helpful. Often users don't provide it and I myself forget about it. We should write a wiki page to document how to troubleshoot _openfortivpn_!
This looks suspect:
```Peer refused to agree to our IP address```
has anything changed on the Fortigate device?

borisovg:
Thanks, I will ask the team that looks after the device.

borisovg:
OK, error went away with me doing nothing - must have been something on the server side. Thanks for your help. :)

ageric:
I experienced the same nonsense when I forgot to use sudo.

Personally, I think openfortivpn should have "Your'e not root. It's frequently needed" as a printout in its path to an error exit (assuming it's true). For me, it would have saved about 40 minutes of post-afterwork-pondering.

DimitriPapadopoulos:
@ageric That looks unrelated. Please open a new ticket.

DimitriPapadopoulos:
@ageric Also _openfortivpn_ prints:
_WARN:   This process was not spawned with root privileges, this will probably not work._
How can we improve on this? We're open to suggestions.

DimitriPapadopoulos:
Opened #373 myself to exit immediately if `geteuid() != 0`.

mrbaseman:
in the recently released version 1.8.0 _openfortivpn_ exits immediately when it is not started with enough privileges (i.e. when it is not started via sudo or direictly launched by root or systemd)

PR #1018, opened 2022-09-22
add option to permit running as non-root

monoidk:
On a configured system, running as a non-privileged user is possible.

Keep bailing out when running as non-root for those who have not configured their systems, so they don't report #362,
but add an option to permit non-privileged use for those who have configured their systems.

In particular, one would need permission to run pppd (or equivalent), which is suid root on typical systems,
then setup pppd such that it handles ip, route, dns configuration. No need to split openfortivpn as proposed in #373, #650,
at least not for the functionality I have tested.

```
cat > /etc/ppp/peers/myconnection << 'EOF'
38400
:192.0.2.1
noipdefault
noaccomp
noauth
default-asyncmap
nopcomp
receive-all
nodefaultroute
nodetach
lcp-max-configure 40
mru 1354
ipparam myconnection
EOF

cat > /etc/ppp/ip-up.local << 'EOF'
case "$PPP_IPPARAM" in
    myconnection)
        # setup networking, dns, etc - run via ppp as root
        ip route add 1.2.3.4/24 dev "$PPP_IFACE"
        ;;
esac 2>&1 | logger -p daemon.debug -i -t "$0"
true
EOF
chmod a+x /etc/ppp/ip-up.local

for f in ip-down ipv6-down ipv6-up; do
    script="/etc/ppp/${f}.local"
    echo '#!/bin/sh' > "$script"
    chmod a+x "$script"
done
```

Then one could start the tunnel, unprivileged with a config like
```
cat > ~/zssk.conf << 'EOFCONF'
host = 1.2.3.4
port = 443
username = myuser
password = mypass
trusted-cert = 1234567890123456789345678456789034567890456789034567891234567890
set-routes = 0
set-dns = 0
pppd-call = myconnection
allow-nonroot = 1
EOFCONF
```


DimitriPapadopoulos:
We can get rid of `pppd` (see #1048), but for now `pppd` must be started by root because of `reauth` (see #650).

If we get rid of `pppd`; setting routes and DNS parameters will require root privileges.

monoidk:
While I understand your plan of removing use of `pppd` and relying on internal encapsulation/decapsulation (seems required for DTLS), I would like to note for anyone interested, that the patch:

 * enables running `openfortivpn` as an unprivileged non-root user (#650), albeit for a narrow use case - requires `pppd`, so no DTLS, `pppd` itself must be suitably configured by an administrator and invoking user does require privilege to invoke pppd (e.g. belong to `dip` group).
 * enables one to configure the system such that only root code that runs is a system supported component with vendor updates and a long track record (`pppd`) instead of an entirety of an evolving code base of `openfortivpn`

As such, it may be more suitable for some users, particularly system administrators who would not shy away from configuring `pppd`.

`Openfortivpn` is implemented in a way that one could expect vulnerabilities to be present (memory-unsafe language, no privilege separation (I think?), evolving code base, no security audits (I think)), whereas `pppd` is at least stable and has had lots of time and eyes on it.

DimitriPapadopoulos:
What about `noauth`?

monoidk:
I'm not really sure what you mean by `noauth`, but I'm using the code as specified in the merge request, with pppd running as root and openfortivpn as a separate unprivileged user (unprivileged except permission to run pppd).

I see `noauth` specified in configuration - I believe pppd does not handle authentication, hence why I may have specified it - it has been a long time since I have configured `openfortivpn` and `pppd`.

I understand `pppd` is written in C, and sadly it has had a few vulnerabilities that have been discovered, but realistically - yes - it seems more widespread, has had a test of time and has automatic updates from the linux distro. It is also a form of privilege separation, as we are only using a part of `pppd` and it is not even exposed to the internet, only to the authenticated server (though I do not have implicit trust for the server).

DimitriPapadopoulos:
`openfortivpn` is not DIRECTLY exposed to internet either, is it?

monoidk:
Yes - it requires using pppd-call and configuring the connection under `/etc/ppp/peers`, which in turn requires root, but only for configuration.

I have not added it to the documentation, though it's listed at the top comment of the merge request (`/etc/ppp/peers/myconnection` example).

DimitriPapadopoulos:
I am asking because I had added myself to `dip`, but `pppd` would nevertheless refuse to be started because of `noauth`. I must have missed something.  See #650:
```sh
$ pppd noauth
pppd: using the noauth option requires root privilege
$
```

So, what you are saying is that after adding `noauth` in `/etc/ppp/peers/myconnection` for example, it works?

monoidk:
Yes, noauth is specified in `/etc/ppp/peers/myconnection`, but not in the command line when using `--pppd-call`, and that is permitted by `pppd`, at least on my system.

DimitriPapadopoulos:
I get it now.

I feel `Requires system/ppp configuration.` is not sufficient. At least more information should be added to the man page – which should be updated any way. Cross-reference `--pppd-call` and `--allow-nonroot` in the man page?

Perhaps we could even let `openfortivpn` run as non-root any way if `--pppd-call` is used, in which case `--allow-nonroot` can be discarded?

monoidk:
I agree it requires better documentation :)

I guess non-root is only useful for `--pppd-call`, so yes, the switch would be redundant, though then the you're-not-root error message should probably mention `--pppd-call`, and documentation for `--pppd-call` should mention it allows use as non-root. Perhaps with a separate section for a sample configuration?

Sorry, I've got to go now. I will check back tomorrow.

PR #1021, opened 2022-09-29
Fix "Peer refused to agree to our IP address"

### LINKED ISSUE #920: Peer refused to agree to our IP address
I wasn't able to connect this morning. Part of the log:
```
INFO:   Negotiation complete.
Peer refused to agree to our IP address
Connect time 0.4 minutes.
Sent 1225 bytes, received 1219 bytes.
Connection terminated.
Modem hangup
INFO:   Cancelling threads...
INFO:   Cleanup, joining threads...
ERROR:  pppd: The link was terminated by the modem hanging up.
INFO:   Terminated pppd.
INFO:   Logged out.
```

Looks similar to #362.

DimitriPapadopoulos:
I do not understand `ipcp-accept-local` would be required or have any effect here:
> With this option, pppd will accept the peer's idea of our local IP address, even if the local IP address was specified in an option.

Indeed, we do not specify the local IP address:
> `":169.254.2.1", // <local_IP_address>:<remote_IP_address>`

Instead, we explicitly ask the peer to provide the local IP address, using option `noipdefault`:
> With this option, the peer will have to supply the local IP address during IPCP negotiation (unless it specified explicitly on the command line or in an options file).

It seems to be working in practice. So let's test it.

Fixes #920.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -238,6 +238,7 @@ static int pppd_run(struct tunnel *tunnel)
 				"230400", // speed
 				":169.254.2.1", // <local_IP_address>:<remote_IP_address>
 				"noipdefault",
+				"ipcp-accept-local",
 				"noaccomp",
 				"noauth",
 				"default-asyncmap",


DimitriPapadopoulos:
I guess it might help when a local IP address is defined in **`/etc/ppp/options`**.

nomeata:
I am observing this with openfortivpn 1.20.3 in NixOS and pppd 2.5.0. Is that a combination that should work?

DimitriPapadopoulos:
Doesn't it work?

DimitriPapadopoulos:
Well, what's the problem, and does this new fix the problem?

nomeata:
Oh, my bad, for some reason I assumed that 1.10.3 contained all patches already! Will try `master` later when near the devices. Sorry for the noise.

akha666:
I have fixed it https://github.com/adrienverge/openfortivpn/issues/1076#issuecomment-1727748285

Neustradamus:
To follow this PR.

PR #1026, opened 2022-10-14
Security hardening for GitHub Actions

DimitriPapadopoulos:
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

The idea is that the software supply chain relies on 3rd party actions that could be compromised. Mitigate this risk by giving these actions minimal rights to the repository. Here read-only access is good enough.


adrienverge:
Thanks!

PR #1034, opened 2022-10-28
WIP: SAML/SSO/2FA support

### LINKED ISSUE #867: Support login SAML
Hi,
I see that openfortivpn supported 2FA but no supports SAML. Can you implement this feature?

Best regards,
Severus

LorenzISR:
This is a WIP. Fixes #867 

I have implemented SAML/SSO using firefox/geckodriver. It is already working, but missing a lot of features and the code needs better documentation. I am opening a pull request here to get feedback from you. It would be great if you could test this and give me feedback if it is working for you (reacting with a 👍). If you get the Error `ERROR:  Failed to create the webdriver session.` please try again few times. This is a known issue. See below for a technical explanation. Please report any other errors that you encounter trying to run it multiple times.

## How to use
You need to compile my branch (https://github.com/LorenzISR/openfortivpnSAML.git). Except from the normal dependencies, you will need geckodriver and firefox (snap version does not work). 

Please run openfortivpn with `--saml=firefox` (Firefox is the only supported browser at the moment). You don't need to set your username or password. For example,
```
openfortivpn vpn.host.tld:443 --saml=firefox
```
should work if you don't need any other flags to connect (Please see below in TODO because these features might not be implemented yet.). 

## Technical explanation
I am creating a custom Firefox profile with custom properties in `~/.mozilla/firefox/openfortivpn/`. The custom properties include disabling the first launch screen, experiments, tab/url bar, etc. 

When the saml setting is enabled, we checking if the profile exists. If the profile exists, it will continue, if not, it will create the profile with the prefs.js and userChrome.css file and then continue. Then, openfortivpn launches geckodriver (See here for what geckodriver does: https://firefox-source-docs.mozilla.org/testing/geckodriver/index.html) as the user given by the `SUDO_UID` enviroment variable (In a child process). This is needed because geckodriver will launch firefox and firefox cannot easily run as root. If `SUDO_UID` is not set, you need to set it manually to the user that should run the browser.

Then we make a request to localhost:4444 (default geckodriver port) to create a new webdriver session (See the webdriver sepc here: https://w3c.github.io/webdriver). This will launch the browser and return a session id. With this session id, you can tell the just launched browser what to do. With a **POST** `/session/<id>/url {"url": "<domain>/remote/saml/start"}` we tell the browser to go to the saml website. 

What we do now is just looping and checking every 1/4 second if the SVPNCOOKIE has been set with **GET** `/session/<id>/cookie/SVPNCOOKIE {}`. If we are looping more than `COOKIE_MAX_TIMEOUT_SECS` (set in `saml.c:22` to 5 minutes) we are breaking the loop and exiting with an error to avoid wasting processing power when the PC is not used. When we got the cookie, we destroy the webdriver session using **DELETE** `/session/<id> {}` and waiting until it has been destroyed. After that, we kill geckodriver. We set the cookie that we got to the `cfg.cookie` variable and continue running normally.

## TODO:
- [x] Support for authentication realms
- [ ] Investigate if WebKit2GTK is actually a better solution
- [ ] Use the specified port
- [ ] Support for chromium
- [ ] Better code documentation
- [ ] Support for other platforms (*BSD, MacOS)
- [ ] Add to the README
- [ ] Add to --help
- [ ] Connection to geckodriver is failing sometimes, fix this.
(This is not a complete list, more to come)

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -238,6 +238,7 @@ static int pppd_run(struct tunnel *tunnel)
 				"230400", // speed
 				":169.254.2.1", // <local_IP_address>:<remote_IP_address>
 				"noipdefault",
+				"ipcp-accept-local",
 				"noaccomp",
 				"noauth",
 				"default-asyncmap",


Newbytee:
Have you considered using e.g. WebKit2GTK instead of this Firefox setup? It's built to be embedded in applications like this. Don't get me wrong, I love Firefox, but I think WebDriver is intended more as an interface for testing than something for end users.

With that said, thanks a bunch for this! Just in time for my university sadly starting to require FortiClient VPN for access to certain resources.

LorenzISR:
Oh, I haven't thought of WebKitGTK, that's actually a pretty good idea! I will have a look, thanks very much for the suggestion!

DimitriPapadopoulos:
Perhaps the core openfortivpn shouldn't be changed into a GUI program instead of a command line program. It would be great if a library was available to mimic a browser from within a command line program, and that could be used to interact with HTML or JavaScript forms pages. I don't know of such a library.

Newbytee:
> It would be great if a library was available to mimic a browser from within a command line program, and that could be used to interact with HTML or JavaScript forms pages. I don't know of such a library.

There are programs like that, namely elinks and w3m. However, neither of them are particularly well-supporting of modern web standards and would not be sufficient to authenticate SAML/SSO/2FA in most if not all cases. Additionally, I don't think they have an API like WebKit2GTK.

Newbytee:
Why did you delete all the code? This was still useful even if not in a mergeable state.

LorenzISR:
The commits where a mess, so I deleted them. The new pull request is #1042 
Can you test if this is working for you?

PR #1035, opened 2022-10-30
Prepend "SVPNCOOKIE=" to the given cookie if missing

gm-vm:
It is maybe not obvious that currently the cookie is only accepted if it has "SVPNCOOKIE=", so add it when it is missing.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -186,37 +186,6 @@ PPPD_USAGE \
 "      trusted-cert = othercertificatedigest6631bf...\n" \
 "  For a full-featured configuration see man openfortivpn(1).\n"
 
-/**
- * This function returns 0 if it could add the "SVPNCOOKIE=" prefix to the given
- * cookie or if the given cookie already had the prefix. If cookie_with_prefix
- * is NULL, then the given cookie already had the prefix, otherwise it will be
- * a dynamically allocated string containing the cookie with the prefix.
- *
- * If the return value is not 0, then the given cookie did not have the prefix
- * and it was not possible to create the string with the prefix.
- * cookie_with_prefix will be NULL in this case.
- */
-static int get_cookie_with_prefix(const char *cookie, char **cookie_with_prefix)
-{
-	char *buf;
-	int len;
-
-	*cookie_with_prefix = NULL;
-
-	if (strstr(cookie, "SVPNCOOKIE=") != NULL)
-		return 0;
-
-	len = strlen(cookie) + strlen("SVPNCOOKIE=") + 1;
-	buf = malloc(len);
-	if (buf == NULL)
-		return 1;
-
-	snprintf(buf, len, "%s%s", "SVPNCOOKIE=", cookie);
-	*cookie_with_prefix = buf;
-
-	return 0;
-}
-
 int main(int argc, char **argv)
 {
 	int ret = EXIT_FAILURE;
@@ -550,16 +519,13 @@ int main(int argc, char **argv)
 			           "cookie") == 0) {
 				char *cookie_with_prefix;
 
-				if (get_cookie_with_prefix(optarg, &cookie_with_prefix)) {
+				cookie_with_prefix = strdup_with_prefix(optarg, "SVPNCOOKIE=");
+				if (cookie_with_prefix == NULL) {
 					log_error("Could not prepend \"SVPNCOOKIE=\" to the cookie.\n");
 					goto exit;
 				}
-
 				free(cli_cfg.cookie);
-				if (cookie_with_prefix != NULL)
-					cli_cfg.cookie = cookie_with_prefix;
-				else
-					cli_cfg.cookie = strdup(optarg);
+				cli_cfg.cookie = cookie_with_prefix;
 				break;
 			}
 			if (strcmp(long_options[option_index].name,
@@ -569,22 +535,19 @@ int main(int argc, char **argv)
 
 				cookie = read_from_stdin(COOKIE_SIZE);
 				if (cookie == NULL) {
-					log_warn("Could not read the cookie from stdin");
-					break;
+					log_error("Could not read the cookie from stdin\n");
+					goto exit;
 				}
 
-				if (get_cookie_with_prefix(cookie, &cookie_with_prefix)) {
+				cookie_with_prefix = strdup_with_prefix(cookie, "SVPNCOOKIE=");
+				free(cookie);
+				if (cookie_with_prefix == NULL) {
 					log_error("Could not prepend \"SVPNCOOKIE=\" to the cookie.\n");
 					goto exit;
 				}
 
 				free(cli_cfg.cookie);
-				if (cookie_with_prefix != NULL) {
-					free(cookie);
-					cli_cfg.cookie = cookie_with_prefix;
-				} else {
-					cli_cfg.cookie = cookie;
-				}
+				cli_cfg.cookie = cookie_with_prefix;
 				break;
 			}
 			goto user_error;
--- a/src/main.c
+++ b/src/main.c
@@ -545,37 +545,19 @@ int main(int argc, char **argv)
 			}
 			if (strcmp(long_options[option_index].name,
 			           "cookie") == 0) {
-				char *cookie_with_prefix;
-
-				cookie_with_prefix = strdup_with_prefix(optarg, "SVPNCOOKIE=");
-				if (cookie_with_prefix == NULL) {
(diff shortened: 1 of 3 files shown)


gm-vm:
Sorry for force-pushing this a couple of times, I couldn't make my mind on how to handle the errors of `get_cookie_with_prefix` and didn't want to add useless commits.

I think that printing an error message and continuing is the easiest way to handle this, the authentication will simply fail and the reason should be clear.

gm-vm:
Exiting immediately is not that more complicated than ignoring (the very unlikely) error, so I just did that.

I will squash the changes if this looks OK.

gm-vm:
By the way, the last fix probably needs to be replicated for all the `strdup(optarg)`, but I think that needs to be done in a separate PR.

DimitriPapadopoulos:
> By the way, the last fix probably needs to be replicated for all the `strdup(optarg)`, but I think that needs to be done in a separate PR.

@gm-vm Is #1041 what you had in mind?

gm-vm:
Pretty much, yeah. I also considered adding something like the following and use it in both `config.c` and `main.c`:

```C
void set_string_argument(char **destination, const char *argument)
{
	free(*destination);
	*destination = strdup(argument);
	if (*destination == NULL)
		log_error("Could not set argument: %s\n", argument);
}
```

src/main.c
```diff
@@ -517,19 +517,37 @@ int main(int argc, char **argv)
...
-				cli_cfg.cookie = strdup(optarg);
+				char *cookie_with_prefix;
+
+				cookie_with_prefix = strdup_with_prefix(optarg, "SVPNCOOKIE=");
+				if (cookie_with_prefix == NULL) {
+					log_error("Could not prepend \"SVPNCOOKIE=\" to the cookie.\n");
+					goto exit;
+				}
+				free(cli_cfg.cookie);
+				cli_cfg.cookie = cookie_with_prefix;
```
DimitriPapadopoulos, line 528: I would rather keep this PR focused on prepending SVPNCOOKIE, without adding checks:
````c
				free(cli_cfg.cookie);
				cli_cfg.cookie = strdup_with_prefix(optarg, "SVPNCOOKIE=");
````
I'm not saying these checks are bad. Just keep them for a different PR.

File: src/main.c [hunk 2]
```diff
@@ -517,19 +517,37 @@ int main(int argc, char **argv)
...
+				cookie_with_prefix = strdup_with_prefix(cookie, "SVPNCOOKIE=");
+				free(cookie);
+				if (cookie_with_prefix == NULL) {
+					log_error("Could not prepend \"SVPNCOOKIE=\" to the cookie.\n");
+					goto exit;
+				}
+
 				free(cli_cfg.cookie);
-				cli_cfg.cookie = cookie;
+				cli_cfg.cookie = cookie_with_prefix;
```
DimitriPapadopoulos, line 550: I would keep it simple as well, without checks, focusing on adding support for `SVPNCOOKIE=`:
```c
				char *cookie = read_from_stdin(COOKIE_SIZE);

				if (cookie == NULL) {
					log_warn("Could not read the cookie from stdin");
					break;
				}

				free(cli_cfg.cookie);
				cli_cfg.cookie = strdup_with_prefix(cookie, "SVPNCOOKIE=");
				free(cookie);
```

src/main.c
```diff
@@ -517,19 +517,37 @@ int main(int argc, char **argv)
...
 			           "cookie-on-stdin") == 0) {
-				char *cookie = read_from_stdin(COOKIE_SIZE);
+				char *cookie;
+				char *cookie_with_prefix;
 
+				cookie = read_from_stdin(COOKIE_SIZE);
 				if (cookie == NULL) {
-					log_warn("Could not read the cookie from stdin");
-					break;
+					log_error("Could not read the cookie from stdin\n");
```
DimitriPapadopoulos, line 538: The extra `\n` seems wrong.

Why `log_warn` → `log_error`? I would keep that for a different PR if needed.

src/userinput.c
```diff
@@ -363,3 +363,31 @@ char *read_from_stdin(size_t count)
...
+	buf = malloc(input_len + prefix_len);
+	if (buf == NULL)
+		return NULL;
+
+	memcpy(buf, prefix, prefix_len);
+	memcpy(buf + prefix_len, input, input_len);
+	buf[input_len + prefix_len] = '\0';
+
+	return buf;
+}
```
DimitriPapadopoulos, line 393: I would rather add this as a static function in `src/main.c`. Or otherwise `src/config`.

src/main.c
gm-vm: `\n` is actually correct, the `log_*` functions do not add it.

Anyway, to me erroring out seems the best way to handle errors here, reason why I changed it from a warning to an error.

Agreed on keeping this out, given that I assume all of this is going to be squashed. I would have kept these changes separate, but a different PR probably makes sense.

src/userinput.c
gm-vm: I turned it back into a static function, since I doubt it'll be needed elsewhere.

src/main.c
```diff
@@ -517,19 +517,37 @@ int main(int argc, char **argv)
...
+					goto exit;
 				}
+
+				cookie_with_prefix = strdup_with_prefix(cookie, "SVPNCOOKIE=");
+				free(cookie);
```
DimitriPapadopoulos: I understand now why the initial function was a bit complex: to avoid double allocation if not needed (already starts with `SVPNCOOKIE=`). Perhaps you can check whether cookie starts with  `SVPNCOOKIE=` before re-allocating.

src/main.c
```diff
@@ -529,7 +557,7 @@ int main(int argc, char **argv)
 					break;
 				}
 				free(cli_cfg.cookie);
-				cli_cfg.cookie = cookie;
+				cli_cfg.cookie = strdup_with_prefix(cookie, "SVPNCOOKIE=");
```
DimitriPapadopoulos, line 560: Just need to `free(cookie).

Perhaps better, with a test:
```c
				free(cli_cfg.cookie);
				if (strncmp(cookie, "SVPNCOOKIE=", strlen("SVPNCOOKIE=")) == 0) {
					cli_cfg.cookie = strdup_with_prefix(cookie, "SVPNCOOKIE=");
					free(cookie);
				} else  {
					cli_cfg.cookie = cookie;
				}
```
Not sure it's worth it though. Opinion?

src/main.c
DimitriPapadopoulos: You're right of course, we can fix all error messages in a different PR.

src/main.c
DimitriPapadopoulos: OK, I understand we can add `free(cookie)` in #1041, right? Perhaps in this case you can add it, because all the surrounding code is changed. Besides, for consistency, I think you should add `free(cookie);` and remove `free(cli_cfg.cookie);`.

gm-vm: Right, I removed too much stuff. I kept `free(cli_cfg.cookie);` simply because it was there already, but I can remove it to make the code consistent.

src/main.c
gm-vm: Not having the checks make the code look a bit simpler and it avoids a tiny bit of duplication, given that it needs to be done for both `cookie` and `cookie-on-stdin`. And what about `strdup_with_prefix`? Should that blindly add the prefix and rely on callers?

In any case, I don't have strong opinions on this, so I'll let you decide. I don't think doing a tiny `malloc()` here for a short-lived string should really matter too much.

src/main.c
DimitriPapadopoulos: Keep it if it was already there. Looks good.

src/main.c
```diff
@@ -186,6 +186,27 @@ PPPD_USAGE \
...
+ * Returns the given "input" prefixed with "prefix" in a dynamically
+ * allocated string. This behaves exactly like "strdup" if "input" already
+ * starts with "prefix".
+ */
+static char *strdup_with_prefix(const char *input, const char *prefix)
+{
+	size_t prefix_len = strlen(prefix);
+	char *output;
+
+	if (memcmp(prefix, input, prefix_len) == 0)
```
gm-vm, line 199: Technically this allows `memcpy` to read `prefix_len` bytes starting from `input`, which is wrong in case `input` is shorter than that. This is why I was comparing the lengths of the strings. Maybe you wanted to use `strncmp`? I used the `mem*` functions simply because I needed to compute the length of everything already.

DimitriPapadopoulos: Right, `strcmp()` it is.

PR #1039, opened 2022-11-18
fixed ubuntu package link in Readme

zdebra:
The original link to ubuntu packages doesn't work. Updated with working link.


EriksRemess:
https://packages.ubuntu.com/search?keywords=openfortivpn does work and is proper one for all the currently supported Ubuntu vesions.

PR #1042, opened 2022-11-29
SAML/SSO/2FA support

LorenzISR:
Hi,

I just deleted the code that I initially wrote and the pull request was closed aswell (Not on purpose). As sugessted by @Newbytee, I've rewritten this with WebKit2Gtk.

It probably needs a compile time option aswell? I need some help with that.

Regarding your comment, @DimitriPapadopoulos "Perhaps the core openfortivpn shouldn't be changed into a GUI program instead of a command line program. It would be great if a library was available to mimic a browser from within a command line program, and that could be used to interact with HTML or JavaScript forms pages. I don't know of such a library."

I don't think adding webkit2gtk to authenticate makes it a gui application. I would also like to avoid browser stuff but unfortunately there is no other way for doing SAML at the moment (None that I know of). I already explained why this should be a feature in #867 

What do you think? Btw, I don't want to sound unkind, its just that my english isn't that great.


LorenzISR:
Btw, thank you for your sugesstion to use webkit2gtk instead, @NewBytee. I feel stupid that I didn't do this initially :laughing:

MrTomRod:
Thanks for your brilliant work!

I have a few issues and suggestions:

# Error on Wayland

I installed this repo on Fedora 37 (Wayland). I get this error:

```bash
(root)$ openfortivpn <myserver.org>:443 --saml
INFO:   Connected to gateway.
Authorization required, but no authorization protocol specified

(openfortivpn:26675): Gtk-WARNING **: 11:37:16.388: cannot open display: :0
ERROR:  Could not authenticate to gateway. Please check the password, client certificate, etc.
INFO:   Logged out.
```

The reason is that root cannot open GTK windows on Wayland I think. What is the recommended way of running `openfortivpn` in this situation?

# Better documentation

It works fine like this: `openfortivpn univpn.unibe.ch:443 --cookie-on-stdin`, but it may not be obvious to users that they must type `SVPNCOOKIE=<cookie>`, then return.

Somewhere, it should be mentioned that the the cookie can be obtained via this page: https://<myserver.org>:443/remote/saml/start . One solution could be to print this URL and brief instructions. 

Moreover, imo, `SVPNCOOKIE=` should be added automatically if missing.

# Update build dependencies

I had to install two more dependencies to build. Please mention that in README.md:

- Fedora: `gtk3-devel` and `webkit2gtk4.0-devel`
- Debian (not tested): `libgtk-3-dev` and `libwebkitgtk-3.0-dev`

MrTomRod:
It works beautifully! Thanks!

mikedehaan:
I think this is brilliant and I love where this is headed, but I'm having an issue. The SAML login does not appear to be following redirects. For example, my login process starts with https://azvpn.mydomain.com:4433/remote/saml/start, but the actual authentication happens at login.microsoftonline.com.

I don't really know what I'm talking about, but I hope this helps.

Below is an excerpt from the official forticlient logs:
`
Fortitray.desktop[5650]: Saml - handleRedirect url=https://azvpn.mydomain.com:4433/remote/saml/start this.saml={"url":"https://azvpn.mydomain.com:4433/remote/saml/start","authTimeout":"360","ignoreCert":false,"type":1} 1
Fortitray.desktop[5650]: 16:28:18.329 › Saml - handleRedirect url=https://azvpn.mydomain.com:4433/remote/saml/start this.saml={"url":"https://azvpn.mydomain.com:4433/remote/saml/start","authTimeout":"360","ignoreCert":false,"type":1}
Fortitray.desktop[5650]: Saml - 'ready-to-show' 1
Fortitray.desktop[5650]: 16:28:18.337 › Saml - 'ready-to-show'
Fortitray.desktop[5650]: SAML - 'ready-to-show'- authTimeout = 360 1
Fortitray.desktop[5650]: 16:28:18.337 › SAML - 'ready-to-show'- authTimeout = 360
Fortitray.desktop[5650]: Saml - 'did-finish-load url=https://login.microsoftonline.com/4444232112-1234-476c-9968-123452323121f/saml2?SomeReallyLongMicrosoftURL
`

LorenzISR:
The problem is probably that the SAML login wasn't using the specified port. Should be fixed now. Could you try if it works for you?

MrTomRod:
I tried to install the latest branch, but get an error on `make`:

```
src/tunnel.c: In function ‘run_tunnel’:
src/tunnel.c:1291:39: error: type of formal parameter 1 is incomplete
 1291 |                 saml_get_cookie(config->gateway_host, config->gateway_port,
      |                                 ~~~~~~^~~~~~~~~~~~~~
src/tunnel.c:1291:61: error: type of formal parameter 2 is incomplete
 1291 |                 saml_get_cookie(config->gateway_host, config->gateway_port,
      |                                                       ~~~~~~^~~~~~~~~~~~~~
make: *** [Makefile:629: src/openfortivpn-tunnel.o] Error 1
```

LorenzISR:
Sorry, I accidentally named to variables the same in a header file. Should work now.

MrTomRod:
Now it work again for me!

LorenzISR:
Thats great! Does it work for you now, @mikedehaan?

MrTomRod:
I noticed an issue:

```C
	// Don't allow other users to read the cookies.
	chmod("/home/lorenz/.openfortivpncookies", 0600);
```

https://github.com/LorenzISR/openfortivpnSAML/blob/3c4bb21d92c70cf05d6bed172139c123ee72ade4/src/saml.c#L155

MrTomRod:
> Running openfortivpn will now require a graphical session, X or Wayland.

Only for 2FA authentications, and only if the user wants to use `--saml` option. The necessary 2FA cookie can also be provided using `--cookie` or `--cookie-on-stdin`. (Maybe instructions should be added to the readme?) I think there is no better way to implement this for 2FA.

> You add a dependency on [WebKitGTK](https://webkitgtk.org/)

Is it possible to make WebKitGTK optional during build? If openfortivpn is built without WebKitGTK, only the `--saml` option should be disabled...

LorenzISR:
>     * You add a dependency on [WebKitGTK](https://webkitgtk.org/), which is a GUI library that may be missing, for exemple from headless servers.

I agree that this might be a unnecessary dependency, but I don't know of any other way to avoid this on most distros. Maby creating a openfortivpn-headless package? I am currently working on implementing a compile time option to disable saml login.
But I don't think many people are using openfortivpn for headless servers?

As @MrTomRod already said in his comment, you only need a graphical session (X/Wayland) if you want to use saml. Everything else does work without it.

mrbaseman:
There's a [package for openwrt](https://openwrt.org/packages/pkgdata/openfortivpn) available for instance, but as far as I have seen there is no X server or webkitgtk on that platform

LorenzISR:
Openwrt could just compile the package without saml support.

DimitriPapadopoulos:
@mrbaseman @adrienverge Would a dependency on [WebKitGTK](https://webkitgtk.org/) be acceptable if a build-time option was available to drop SAML support and the [WebKitGTK](https://webkitgtk.org/) dependency?

Again, note that a graphical session is required only for SAML. Otherwise `openfortivpn` runs as usual in text mode.

DimitriPapadopoulos:
@LorenzISR Can you fix tests?

DimitriPapadopoulos:
Also note that openfortivpn is often called from Networkmanager-fortisslvpn. I am afraid this will break Networkmanager-fortisslvpn.

adrienverge:
> @mrbaseman @adrienverge Would a dependency on [WebKitGTK](https://webkitgtk.org/) be acceptable if a build-time option was available to drop SAML support and the [WebKitGTK](https://webkitgtk.org/) dependency?

I'd say it's not desirable because it will make openfortivpn harder to compile on multiple platforms (even if it's possible to disable it, it complexifies the process for newcomers), and we'll have to follow GTK updates / deprecations.

Isn't there a way to avoid relying on graphical sessions?

MrTomRod:
# Suggestion for extending the docs

* Connect heedlessly using saml:
  ```
  openfortivpn vpn-gateway:8443 --cookie <cookie>
  openfortivpn vpn-gateway:8443 --cookie-on-stdin
  ```

<details>
<summary>Click here to learn how to obtain the cookie</summary>

1.  Open the SAML login page in a web browser. It is ususally `https://vpn-gateway:443/remote/saml/start`
2.  Log in
3.  Open the developer tools and navigate to Storage (Firefox: `Shift`+`F9`, Chrome: `F12`)
4.  Copy the value of `SVNCOOKIE`

openfortivpn instructions


</details>

DimitriPapadopoulos:
> I'd say it's not desirable because it will make openfortivpn harder to compile on multiple platforms (even if it's possible to disable it, it complexifies the process for newcomers), and we'll have to follow GTK updates / deprecations.

Perhaps if SAML support was not built in **by default**?

> Isn't there a way to avoid relying on graphical sessions?

@adrienverge We need a way to interact with complex authentication pages: javascript, redirections, etc. A headless browser ideal would be ideal, with the possibility to find fields asking for passwords in pages, ask for the password in text mode from the command line, then feed back the password to the pages. I don't know of such a library, except perhaps the two obsolete libraries in https://github.com/adrienverge/openfortivpn/pull/1034#issuecomment-1330264550.

I haven't found any C library in the list of [Headless Browsers](https://github.com/dhamaniasad/HeadlessBrowsers), but there are Python modules. The [Selenium WebDriver](https://www.selenium.dev/documentation/webdriver/) supports Python and other programming languages, but not C.

DimitriPapadopoulos:
In short, the authentication web page sequence is too unpredictable to automate the processing of these pages. Yet OpenConnect  appears to be able to do that, not only for Fortinet but also other SSL VPN protocols, not without difficulty I must admit.

Any way, the real problem here is integration with NetworkManager. How would that work?

DimitriPapadopoulos:
I am not sure that's pragmatic, I believe most users use openfortivpn through NetworkManager.

Changing openfortivpn into a library, with a lightweight main() passing callbacks for user interaction to the library could be a solution. Of course, NetworkManager would need to be modified, and implement its own graphical callbacks.

MrTomRod:
> I am not sure that's pragmatic

It would make it possible to use openfortivpn with 2FA _now_. My university recently enforeced 2FA and everyone had to switch to the crappy official client. Waiting for some NetworkManager workaround could take many months.

> most users use openfortivpn through NetworkManager

Googling around (terms: "openfortivpn" university), I find that most instructions mention either only the CLI or both.

mikedehaan:
> The problem is probably that the SAML login wasn't using the specified port. Should be fixed now. Could you try if it works for you?

Yes!!!! That fixed my SAML issue. Works perfectly now!

mrbaseman:
Let me come back to the discussion about a compile option to disable saml. What about a configure check which automatically enables saml support only if webgtk is available? Perhaps this is a simple check for gtk devel headers or so...?

DimitriPapadopoulos:
@mrbaseman I think we should be able to disable SAML support at build-time, even when GTK is available on the build machine.

adrienverge:
> Perhaps if SAML support was not built in **by default**? In nay case, autotools would detect the lack of WebKitGTK and disable SAML support in that case.
> @mrbaseman I think we should be able to disable SAML support at build-time, even when GTK is available on the build machine.

+1, it would need to be disablable, and why not disabled by default (even when GTK is available) for easier compilation and maintainance for package maintainers.

Like @DimitriPapadopoulos, I believe many people use openfortivpn through NetworkManager, so the problem would remain for them.

LorenzISR:
As far as I can tell, the biggest problem seems to be the dependency on Webkit-/GTK (which are huge packages) for people who don't need it.

I don't know if most people are using openfortivpn through NetworkManager but if this is the case its probably a better idea to just let NetworkManager (or alternatives) do this if used for most people. 

If its disabled by default at build time, there is little sense in merging this here, right?

> Like @DimitriPapadopoulos, I believe many people use openfortivpn through NetworkManager, so the problem would remain for them.

NetworkManager could just pass the saml flag to openfortivpn.

In the end you need to decide if the extra dependency is worth it or not. I think its more easy to use openfortivpn from the command line rather than from NetworkManager but I don't know how many people are on my side.

DimitriPapadopoulos:
> NetworkManager could just pass the saml flag to openfortivpn.

That's interesting:
* As far as I can see, without `---saml`, the SAML code is **not** executed. So, the only notable change when running `openfortivpn` with default parameters is indeed the dependency on GTK.
* Users who need SAML support must pass `---saml`.
* My concern with NetworkManager is that it should take over all GUI interaction. Passing `---saml` does **not** seem a good idea in this case, NetworkManager should handle all SAML internally. Unfortunately, openfortivpn lacks the machinery to allow NetworkManager to take over the GUI part of SAML authentication.

With all that said, I vote for applying this patch.

@adrienverge I realize this leaves openfortivpn in a less clean state, but then major changes are needed to address new needs in the cleanest possible way:
* switch from `pppd` to a TUN device and internal PPP code (see #1048),
* perhaps most importantly, turn openfortivpn into a library that implements the core Fortinet VPN SSL protocol, and leave the rest to callbacks (interaction with authentication forms, routing and DNS),
* ...

DimitriPapadopoulos:
@theinvisible As a maintainer of  [openfortigui](https://github.com/theinvisible/openfortigui/commits/master) what is your opinion on adding SAML support based on WebKit/GTK to openfortivpn?

er1:
New to the scene here but this feature is of interest to me.

From what I gather, we don't really need WebKit/GTK here so much as need a browser to be involved. Could we instead set up a local HTTP server for the SAML process and then have the browser POST back to that local server any token relevant tokens?

It would require that we depend on something that would provide HTTP which I imagine would be might lighter and we can depend on the user's local browser regardless of their OS and start that up with `xdg-open` (or just `open` on macOS) if that is installed and present falling back to just printing the URL to the screen for the user to open their browser manually.

adrienverge:
> As far as I can tell, the biggest problem seems to be the dependency on Webkit-/GTK (which are huge packages) for people who don't need it.

That's true, these huge packages + the extra complexification of compilation.

> I realize this leaves openfortivpn in a less clean state, but then major changes are needed to address new needs in the cleanest possible way:

@DimitriPapadopoulos I follow you on the choice you think is the best.

Inputs from @theinvisible and @er1 could be fruitful.

src/saml.c
```diff
@@ -0,0 +1,222 @@
...
+		char *home_dir = get_under_home_dir("");
+		char *xdg_runtime_dir = malloc(strlen("/run/user/XXXXXXXXXX") + 1);
+		sprintf(xdg_runtime_dir, "/run/user/%d", browser_uid);
+
+		clearenv();
+		setenv("HOME", home_dir, 1);
+		setenv("DISPLAY", ":0", 1);
+
+		// Needed for wayland
+		setenv("XDG_RUNTIME_DIR", xdg_runtime_dir, 1);
```
mvf, line 200: These hard-coded envvars assume the the user's homedir is under `/home`, their X11 session is on local display 0 and their `XDG_RUNTIME_DIR` is at `/run/user/$UID`. `HOME`, `DISPLAY` and `XDG_RUNTIME_DIR` might not suffice either. What about `DBUS_SESSION_BUS_ADDRESS`, `XMODIFIERS`, `GTK_IM_MODULE`, etc.?

src/saml.c
```diff
@@ -0,0 +1,222 @@
...
+
+		setuid(browser_uid);
```
mvf, line 202: If `setuid` fails, the browser will run as root. If it succeeds, it will run as `browser_uid` but still have access to file descriptors that were opened as root.

mvf (changes requested):
I don't believe `openfortivpn` can cleanly and reliably start a browser (see `saml_get_cookie` comments). But why is that even necessary?

Maybe I'm missing something obvious, but IMO #974 implemented all the SAML/2FA support that belongs into core `openfortivpn`. Network Manager & friends should now be able to implement auth on top of `--cookie[-on-stdin]` ([issue link](https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn/-/issues/57)).

For CLI, there's [`openfortivpn-webview`](https://github.com/gm-vm/openfortivpn-webview), a browser that prints the VPN cookie to `stdout`. This hacky one-liner is what I've been using for months:

```sh
$ openfortivpn-webview my.fortigate.com | sudo openfortivpn my.fortigate.com --cookie-on-stdin
```

Linking WebKitGTK *significantly* increases footprint, even when the browser isn't used.

RSS of `openfortivpn` at `__libc_start_main`, i.e. before doing anything:
```
# 1.19.0
$ grep ^VmRSS /proc/21337/status
VmRSS:      1936 kB

# This PR
$ grep ^VmRSS /proc/21581/status
VmRSS:     30420 kB
```

Startup also feels noticeably slower, but I didn't measure that.

The footprint issues are fixable, but IMO the whole "impersonate a user and start a browser" approach is both infeasible and unnecessary.

DimitriPapadopoulos:
@er1 This all started with ticket #867 which ended up with multiple implementations:
* #1034 did rely on an external browser, but it was closed because WebKitGTK looked like a better, more integrated approach. 
* #1042 (this ticket) took over with an integrated WebKitGTK approach.
* [**gm-vm/openfortivpn-webview**](https://github.com/gm-vm/openfortivpn-webview) is an external program based on the `--cookie` option.

Does your proposal add something radically different?

DimitriPapadopoulos:
@gm-vm I believe a drawback of [**gm-vm/openfortivpn-webview**](https://github.com/gm-vm/openfortivpn-webview) is that installation might appear difficult for many users:
* Use the [Qt](https://www.qt.io) or [Electron](https://www.electronjs.org/) version?
* How hard is it to port from Qt 5 to Qt 6?
* I had read something about deprecation and Electron. But I guess this was about [Atom](https://github.blog/2022-06-08-sunsetting-atom/), not Electron, wasn't it?
* RPM or DEB packages are not available (yet). How to install?
* Are all of Linux/FreeBSD/macOS supported?

DimitriPapadopoulos:
@LorenzISR I believe @mvf raises thought-provoking concerns:

> I don't believe `openfortivpn` can cleanly and reliably start a browser (see `saml_get_cookie` comments). But why is that even necessary?

Indeed, what does the WebKitGTK integration offer that an external program such as [**gm-vm/openfortivpn-webview**](https://github.com/gm-vm/openfortivpn-webview) doesn't?

Wouldn't WebKitGTK-based `external` software be the best solution for Linux/BSD platforms? A challenge would be to rapidly provide RPM/DEB packages bundled with Linux distributions.

Or instead of re-inventing the wheel, extending [OpenFortiGUI](https://github.com/theinvisible/openfortigui) by @theinvisible to support SAML could maybe do the trick.

gm-vm:
@DimitriPapadopoulos

> - Use the Qt or Electron version?

It shouldn't matter, but I created the Electron version simply because the Qt one stopped working with my SAML provider for some unknown reason. I had no issues with Electron and so I ended up using it. I also remember doing the authentication from a minimal python+Gtk+WebKit browser.

The main advantage of the Qt version is that it uses shared libraries for pretty much everything and so the binary is really small.

> - How hard is it to port from Qt 5 to Qt 6?

Already supported. I should update the README.

> - I had read something about deprecation and Electron. But I guess this was about Atom, not Electron, wasn't it?

Yep, that was Atom.

> - RPM or DEB packages are not available (yet). How to install?

The Qt version needs to be built from source, while there are portable binaries for Linux (x64) of the electron variant here: https://github.com/gm-vm/openfortivpn-webview/releases

The AppImage can be executed as is, the tar.gz needs to be extracted and inside of it there will be an executable (and a bunch of files).

> - Are all of Linux/FreeBSD/macOS supported?

Linux for sure, but I haven't tried it on anything else. I know someone else has been using the Electron application with macOS. No idea about FreeBSD, but it's mostly a matter of Qt or Electron being supported there.

er1:
> @er1 This all started with ticket #867 which ended up with multiple implementations:
> 
>     * [WIP: SAML/SSO/2FA support #1034](https://github.com/adrienverge/openfortivpn/pull/1034) did rely on an external browser, but it was closed because WebKitGTK looked like a better, more integrated approach.
> 
>     * [SAML/SSO/2FA support #1042](https://github.com/adrienverge/openfortivpn/pull/1042) (this ticket) took over with an integrated WebKitGTK approach.
> 
>     * [**gm-vm/openfortivpn-webview**](https://github.com/gm-vm/openfortivpn-webview) is an external program based on the `--cookie` option.
> 
> 
> Does your proposal add something radically different?

No, debian stable is currently on 1.15.0 where --cookie does not exist, in looking for SAML support, this ticket came up first which led me to (wrongly) believe SAML support was not implemented yet but progress was being made toward it. I believe openfortivpn-webview will fit my use case.

LorenzISR:
I am pretty much out of arguments at this point why this should be merged. I think that doing this with an external program using `--cookie-on-stdin` is probably better for everyone. The only problem that I see with this approach is that this program needs to be packed for the different distributions which will probably be quite a challenge but this can be solved. https://github.com/gm-vm/openfortivpn-webview looks promising.

DimitriPapadopoulos:
@gm-vm Do you have an idea how best to pack [openfortivpn-webview](https://github.com/gm-vm/openfortivpn-webview) for different distributions? While I am not the maintainer of openfortivpn and cannot speak for him, I wonder whether you would agree to merge the two projects or maybe pack the sources together, using Git subtrees or submodules for example. Ideas?

gm-vm:
@DimitriPapadopoulos I'm using `electron-builder` to make the `AppImage` and package the application as `tar.xz`, but those two are not the only options: https://www.electron.build/configuration/linux

I'm only building those two because other than being (almost) universal, I need to adjust few things for the other formats (I remember for example that `electron-builder` did not like that there was no maintainer specified).

The Qt app needs work to be packaged properly, right now the Makefile only generates an executable.

My hope is that some day someone will add the support for this directly in NetworkManager. openfortivpn-webview was just a way to conveniently use the VPN without first studying how to patch NetworkManager.

I'm not against including it as submodule, but is that desired?

While the Qt application requires almost no maintenance, the electron app needs to be updated constantly due to the fact that it bundles electron. Luckily there's dependabot and so far I think it notified me about only two vulnerabilities that were affecting electron itself, all the other vulnerabilities only had impact on stuff used at build time. I still had to go through the changes and check they were OK.

What I'm saying is that there is some maintenance burden and providing packages for popular distributions only make things worse.

Not to mention the fact that there are two seemingly equivalent applications that makes everything confusing, I'd probably have to sort that out.

I understand the desire of providing an easy way to use SAML with `openfortivpn`, but I'd think this through.

Alternatively there could even simply be a section in the wiki (maybe linked in the README, because I personally always miss the wiki of Github projects), explaining how to do the authentication with SAML providers using external applications, even standard browsers. This would also allow to share simple scripts to do the authentication headlessly where doable.

Predrag:
Hello,

Is it possible to clone your branch yet ? I am using your SAML solution to connect  to work vpn. Without your solution I will be not able to connect to vpn in future.

MrTomRod:
@Predrag

This branch still works on my system. You could compile it yourself and use it until it stops working.

Otherwise, you could install the latest version of openfortivpn (Fedora: 1.17.0-5 or later / Ubuntu: 1.19.0-2 or later) and perform 2FA yourself using [following these instructions](#issuecomment-1344211491).

If I understood correctly, no GUI will be added to this library and you have to wait for things like [openfortigui](https://github.com/theinvisible/openfortigui/issues/186) to make this more convenient.

Neustradamus:
@LorenzISR: Why this PR has been closed?

BLaDZer:
@Neustradamus if you still interested in this PR you may want to check build in my [fork](https://github.com/BLaDZer/openfortivpn/releases) with a few small patches for my fortivpn provider

Neustradamus:
@BLaDZer: Thanks for your comment!

It will be better to have in official build...

I do not know why there is not an official support...

Predrag:
I agree

Neustradamus:
Dear all,

It is possible to create a new PR linked to this original?

Thanks in advance.

LorenzISR:
since openfortivpn runs as root, it does not make any sense to run the browser as root too. this whole PR is a complete hack, as the review @mvf also shows. it's a much better idea to use an external program that fetches the cookie and pipes it into the stdin.

PR #1074, opened 2023-03-09
add Dockerfile and notes

inceabdullah:
Hello,

I generally use openfortivpn with docker. There are some fortivpnclient docker images but not `openfortivpn`. So I would like to add Dockerfle to use it everywhere.

in Docker needed two param ` --privileged --cap-add=NET_ADMIN` and `--net host`

There two options to run it

`docker run --privileged --cap-add=NET_ADMIN --net host -d -v config:/etc/openfortivpn/config openforticlient`

or could be run with other params like `--set-dns=0 --pppd-use-peerdns=1`

`docker run --privileged --cap-add=NET_ADMIN --net host -d -v config:/etc/openfortivpn/config openforticlient -c /etc/openfortivpn/config --set-dns=0 --pppd-use-peerdns=1`

Thank you.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM alpine
+
+WORKDIR /opt
+
+RUN apk update
+
+RUN apk add --update-cache \
+    git autoconf automake pkgconfig gcc g++ libressl-dev make ppp-pppoe
+
+COPY . ./
+
+RUN ./autogen.sh && \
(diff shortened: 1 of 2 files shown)


adrienverge:
Hello,

Having a Docker image for openfortivpn is a good idea, but like for distribution packages (rpm, deb, flatpak, etc.) I believe it's better not to maintain them here.

PR #1082, opened 2023-04-07
limit number of retries

DonThorntonJr:
limit number of retries for persistent connections


src/main.c
```diff
@@ -85,6 +85,7 @@ PPPD_USAGE \
 "                    [--user-cert=<file>] [--user-key=<file>]\n" \
 "                    [--use-syslog] [--trusted-cert=<digest>]\n" \
 "                    [--persistent=<interval>] [-c <file>] [-v|-q]\n" \
```
mrbaseman, line 87: Shouldn't line 88 replace line 87 here? But maybe the line becomes too long then and should be split.

mrbaseman (changes requested):
The man page should be updated to reflect this new option.

Overall I'm not sure if it is really that useful. On the other hand it doesn't add too much code, so if it helps some users in their setups, why not?

The code should be squashed and rebased, because some other changes have landed on master in the meantime.

PR #1092, opened 2023-04-28
 Add support for systemd-resolved

DimitriPapadopoulos:
Re-open #615. Not sure how it got closed.


Tatsh:
I've had the same behaviour before. I honestly don't think GitHub documents the logic of why a PR should close without warning. It sees '0 changes' somehow so this signals the system to automatically close the PR.

Tatsh:
I no longer have access to a host to test this against (new job), but I was using a variant of a C++ version of the script: https://github.com/Tatsh/ifup-systemd-resolved. Would really appreciate if anyone using this patch could try it as I've made some minor updates.

PR #1093, opened 2023-05-04
Improve openfortivpn@.service

DimitriPapadopoulos:
Based on equivalent OpenVPN file:
`distro/systemd/openvpn-server@.service.in`

See also OpenVPN service:
https://community.openvpn.net/openvpn/wiki/Systemd


adrienverge:
Nice :+1:

PR #1111, opened 2023-05-30
Pass ipcp-accept-remote to pppd

floppym:
Since https://github.com/ppp-project/ppp/commit/9fe8923419a954fedf8b6d1a6cc07b45f165c1ab, pppd refuses to accept a different remote IP if we explictly pass one on the command line. This results in an error like:

pppd: Peer refused to agree to his IP address

Passing ipcp-accept-remote disables this behavior.

Bug: https://bugs.gentoo.org/907404

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -236,6 +236,7 @@ static int pppd_run(struct tunnel *tunnel)
 				":169.254.2.1", // <local_IP_address>:<remote_IP_address>
 				"noipdefault",
 				"ipcp-accept-local",
+				"ipcp-accept-remote",
 				"noaccomp",
 				"noauth",
 				"default-asyncmap",


DimitriPapadopoulos:
Interesting. I had added `ipcp-accept-local` in https://github.com/adrienverge/openfortivpn/pull/1021 in an attempt to get rid of this error message. Are both needed?

floppym:
> Interesting. I had added `ipcp-accept-local` in #1021 in an attempt to get rid of this error message. Are both needed? Perhaps to tackle different problems?

Yes, the problems are similar, but ultimately different.

The change in #1021 seems to work around a pppd bug that occurs when the local IP is unspecified.

The change from this PR disables enforcement of the remote IP when one is explicitly configured. It is unlikely that the VPN server would ever accept "169.254.2.1" as its address.

DimitriPapadopoulos:
@floppym Unfortunately, this change breaks macOS (#1118) and NetworkManager on Linux (#1120).

I have therefore made it optional in #1119. Opinions? Better alternatives?

floppym:
@DimitriPapadopoulos I don't really understand how pppd-accept-remote would cause either of those issues. I feel like this revert is not the right solution.

At the very least, I think the pppd-accept-remote should default to enabled to allow ppp-2.5.0 to function properly by default. Legacy users should disable it.

floppym:
In any case, I don't use this package personally, and I will probably suggest that users migrate to OpenConnect instead.

DimitriPapadopoulos:
> @DimitriPapadopoulos I don't really understand how ipcp-accept-remote would cause either of those issues.

I don't understand either, it doesn't make any sense. It has to be a bug elsewhere, probably in the routing changes. I am only reverting to the previous situation to give some time to investigate and get to the bottom of this issue.

> At the very least, I think the ipcp-accept-remote should default to enabled to allow ppp-2.5.0 to function properly by default. Legacy users should disable it.

Most users are legacy users and they cannot disable the option because they openfortivpn through [NetworkManager-fortisslvpn](https://gitlab.gnome.org/GNOME/NetworkManager-fortisslvpn).

> In any case, I don't use this package personally, and I will probably suggest that users migrate to OpenConnect instead.

Sure, but transitioning takes some time.

uffejakobsen:
> 
> > In any case, I don't use this package personally, and I will probably suggest that users migrate to OpenConnect instead.
> 
> Sure, but transitioning takes some time.

Just my few cents on this topic

I think that the transitioning will take a long time - mostly because of one reason:

OpenConnect is a good solution - but the NetworkManager GUI dialogs for OpenConnect Fortinet SSLVPN are just weird and messy with OTP/2FA

The NetworkManager GUI dialogs for openfortivpn are much easier to operate in day-to-day operation...

woprandi:
I just discovered OpenConnect as an alternative client since openfortivpn + pppd 2.5.0 seems broken on Fedora 39.
Does OpenConnect should be preferred ?

Neustradamus:
To follow this PR.

PR #1122, opened 2023-06-22
openfortivpn 1.20.5

### LINKED ISSUE #1121: openfortivpn version 1.20.5
Unfortunately, #1111 fixes the issue with PPP 2.5.0, but breaks NetworkManager on Linux, and macOS massively report issues.

I am making the code in #1111 optional, but that is suboptimal. We should have a solution that works everywhere. But for now, I have nothing better to suggest. I would like a new version with this code made optional.

Thoughts?

DimitriPapadopoulos:
Fixes #1121.


adrienverge (approved):
Thanks :+1:

PR #1139, opened 2023-09-25
Merge 'tun' config value if set in source configuration

DblD:
This commit modifies the `merge_config` function to include the 'tun' value from the source configuration if it is set. Previously, this value was not being merged, leading to inconsistencies in the final configuration.

The change checks if 'tun' in the source configuration is different from its default value before merging it into the destination configuration.

This ensures that the 'tun' setting is correctly propagated when configurations are merged, allowing for more flexible and accurate VPN setups.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -523,6 +523,9 @@ void merge_config(struct vpn_config *dst, struct vpn_config *src)
 		free(dst->pinentry);
 		dst->pinentry = src->pinentry;
 	}
+	if (src->tun != invalid_cfg.tun) {
+		dst->tun = src->tun;
+	}
 	if (src->realm[0])
 		strcpy(dst->realm, src->realm);
 	if (src->iface_name[0])


src/config.c
```diff
@@ -523,7 +523,10 @@ void merge_config(struct vpn_config *dst, struct vpn_config *src)
 		free(dst->pinentry);
 		dst->pinentry = src->pinentry;
 	}
-	if (src->realm[0])
+        if (src->tun != invalid_cfg.tun) {
+                dst->tun = src->tun;
+        }
+        if (src->realm[0])
```
DimitriPapadopoulos, line 529: Looks good, please fix indentation.

DblD: ups what did i miss, thought i had them all on 8 spaces as required ?

DimitriPapadopoulos: It's tabs, not spaces.

DimitriPapadopoulos: Looks good now!

DimitriPapadopoulos:
Strange. CI checks have not been triggered.

PR #1144, opened 2023-10-14
Use bash and python3 from env

Luz:
Hi
According to [Stackoverflow](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang#10383546) and [Stackoverflow](https://stackoverflow.com/questions/7670303/purpose-of-usr-bin-python3-shebang) it looks like the portability of the scripts could be improved by having this new style for the shebangs.

### DIFF
--- a/etc/ppp/ip-down.local.example
+++ b/etc/ppp/ip-down.local.example
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 case "$PPP_IPPARAM" in
     openfortivpn*)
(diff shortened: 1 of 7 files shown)


adrienverge:
@DimitriPapadopoulos I will stand behind your decision, but it could be useful to change this under `tests/` too? In case someone on an exotic *nix system wants to run them?

DimitriPapadopoulos:
@adrienverge You're right, test scripts are not run in CI jobs only, they are run by end-users too.

PR #1152, opened 2023-11-05
Assorted small fixes

DimitriPapadopoulos:
These need not appear in the change log.


DimitriPapadopoulos:
Partially reverted by #1155.

PR #1154, opened 2023-11-05
Fix line length linter

DimitriPapadopoulos:
* Write to `sys.stderr` in the hope the error message will show up in the CI job output.
* Start enumerate at 1 instead of 0, to show the correct line number.


adrienverge:
Nice, thanks.

PR #1158, opened 2023-11-06
openfortivpn 1.21.0

### LINKED ISSUE #1157: openfortivpn version 1.21
This version should enable passing option `ipcp-accept-remote` to `pppd` by default, unless built with `--enable-legacy-pppd`. We had briefly tried to do that in 1.20.4, and had to revert in 1.20.5 as the change would break pppd < 2.5.0. Hopefully this new version can benefit both with  pppd ≥ 2.5.0 but also users with  pppd < 2.5.0.

The `ipcp-accept-remote` option is required by pppd ≥ 2.5.0 that some Linux distributions started shipping quite a few months ago.

Packagers that target platforms with pppd < 2.5.0 should build with `--enable-legacy-pppd`.

Packagers that build for platforms that ship either pppd ≥ 2.5.0 or pppd < 2.5.0 should decide which default fits most of their end-users. Other end-users may have to use options `--pppd-accept-remote=0` or `--pppd-accept-remote=1` if they need something different than the packagers chose as the default.

DimitriPapadopoulos:
Fixes #1157.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ Releases
 This high level changelog is usually updated when a release is tagged.
 On the master branch there may be changes that are not (yet) described here.
 
+### 1.21.0
+
+* [~] fix "Peer refused to agree to his IP address" message, again
+* [~] deprecate option --plugin
+* [-] better masking of password in logs
+* [-] break on reading 0 from ppp pty, for non-Linux systems
+
 ### 1.20.5
 
 * [-] revert previous fix from 1.20.4, make it optional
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,13 @@ Releases
 This high level changelog is usually updated when a release is tagged.
 On the master branch there may be changes that are not (yet) described here.
 
+### 1.21
```
adrienverge, line 17: Could you use `1.21.0` here, and in the commit title? For consistency with all previous releases, and to easily differentiate with future `1.21.x` versions.

DimitriPapadopoulos: Fixed.

adrienverge: :pray:

adrienverge (approved):
Thank you @DimitriPapadopoulos!

PR #1160, opened 2023-11-06
SSL → TLS

DimitriPapadopoulos:
* I have changed most user-visible strings, documentation, and code comments.
* I have also changed macro constants, keeping old definitions and marking them as deprecated.
* I have not changed function and variable names, because that would break compatibility.

The most debatable change would be:
```
openfortivpn - Client for PPP+TLS VPN tunnel services
```
instead of:
```
openfortivpn - Client for PPP+SSL VPN tunnel services
```
Indeed, "VPN SSL" remains much more used than "VPN TLS". I have therefore kept that as is.

@adrienverge Comments?

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -37,7 +37,7 @@ jobs:
             --form email=DimitriPapadopoulos@users.noreply.github.com \
             --form file=@openfortivpn.xz \
             --form version=coverity_scan \
-            --form description="Client for PPP+SSL VPN tunnel services" \
+            --form description="Client for PPP+TLS VPN tunnel services" \
             https://scan.coverity.com/builds?project=adrienverge%2Fopenfortivpn
         env:
           TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,7 +98,7 @@ On the master branch there may be changes that are not (yet) described here.
 * [-] improve tunnel speed on macOS
 * [-] modify memory allocation in the tunnel configuration structure
 * [+] openfortivpn returns the PPP exit status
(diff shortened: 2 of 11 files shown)


adrienverge:
I am myself a supporter of the more recent word "TLS", but I remember writing "SSL" when creating openfortivpn mostly because the main proprietary client was called "FortiSSL". I believe they changed this name to "FortiClient VPN", so I don't see any reason to keep "SSL" now. Except from the fact that you point out: "VPN SSL" remains more used currently. But we could/should show example and use the new word.

A compromise could be to use "TLS/SSL" inside documentation strings for 1 ou 2 years. What do you think?

DimitriPapadopoulos:
I feel "TLS/SSL" is too complex. The FortiClient download page still refers to "SSL VPN and IPSecVPN":

https://www.fortinet.com/support/product-downloads#vpn

adrienverge:
OK, understood.

If it was just me, I would dare to use "TLS" even in the title "Client for PPP+SSL VPN tunnel services" (`-h`, man page, README...)
But I'm perfectly fine with your choice too, so I let you the final word :slightly_smiling_face:

DimitriPapadopoulos:
OK let's innovate and use TLS everywhere.

adrienverge:
OK!

I just changed the GitHub page description too:
GitHub page

mrbaseman:
The change has been merged already, but since you have requested my review: 
I fully agree. TLS has replaced the original SSL protocol, but just the name has been used allover (perhaps most of the SSL/TLS implementations still carry ssl in their name?). Even for TLS the first standardized  minor versions are considered less secure nowadays, not speaking about SSLv3 and less. Therefore, moving from SSL to TLS throughout the code seems to emphasize the right message to me (and it's easier for us to do this move, because we don't carry ssl in the name of the project).

PR #1200, opened 2024-02-23
Accept cookie in the config file

sino1641:
Hi,
I would like to request if it is possible to accept cookie in the config file.
I tried to copy everything from .otp and yes the cookie is accepted by the program.
But there are still 3 TODOs which I have no idea:

- [ ] Refactor prepend `SVPNCOOKIE` function for char array.
- [ ] Refactor cli_cfg.cookie and cookie from stdin. I tried overnight but all returns segment fault. It seems manipulating pointers is out of my knowledge as a beginner.
- [ ] Which cookie should be used when more than one cookies are imported to the program.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,8 @@
 aclocal.m4
 autom4te.cache
 compile
-config.log
-config.status
+config.*
+config
 configure
 cscope.*
 depcomp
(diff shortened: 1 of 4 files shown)


DimitriPapadopoulos:
Why? The cookie is temporary. Therefore it doesn't belong to a config file, does it?

PR #1213, opened 2024-04-22
Let know we don't support HTTP compression

### LINKED ISSUE #1212: Wrong value in the 'Accept-Encoding' header
When the client sends an HTTP request to a server, it includes the **Accept-Encoding** header to express its preference for content compression, but it seems compression is not supported. Some versions/installations of FortiNet server might support compression and start sending compressed content and cause errors (e.g. `auth_get_config`).

In **src/http.c** we have the following:
`"Accept-Encoding: gzip, deflate, br\r\n"`

To fix this, the client should set **Accept-Encoding** to **identity**.
`"Accept-Encoding: identity\r\n"`

DimitriPapadopoulos:
It looks like newer Fortinet gateways will send compressed HTTP if we notify them we support it - but we don't.

Fixes #1212.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -275,7 +275,7 @@ static int do_http_request(struct tunnel *tunnel,
 	                        "Host: %s:%d\r\n"
 	                        "User-Agent: %s\r\n"
 	                        "Accept: */*\r\n"
-	                        "Accept-Encoding: gzip, deflate, br\r\n"
+	                        "Accept-Encoding: identity\r\n"
 	                        "Pragma: no-cache\r\n"
 	                        "Cache-Control: no-store, no-cache, must-revalidate\r\n"
 	                        "If-Modified-Since: Sat, 1 Jan 2000 00:00:00 GMT\r\n"


rodrigoaguilera:
@DimitriPapadopoulos I'm unable to connect since the 1.22.0 update. Might this be related?
I get

```
INFO:   Got addresses: [172.16.232.11], ns [172.16.249.101, 172.16.49.101]
INFO:   Negotiation complete.      
INFO:   Negotiation complete.       
Peer refused to agree to his IP address
Connect time 0.1 minutes.                                                            
Sent 1101 bytes, received 1081 bytes.
ERROR:  Timed out waiting for the ppp interface to be UP.
```

DimitriPapadopoulos:
No, I don't see how it could be related.

Please open an issue, and provide information on the OS and the version that worked before the upgrade as opposed to 1.22.0.

DimitriPapadopoulos:
On platforms with PPP < 2.5.0 **and** if openfortivpn has not been properly built (MacPorts and perhaps even Homebrew despite https://github.com/Homebrew/homebrew-core/commit/d88b7a153a82827379aa729360fd71a4e1c47e52), use the `pppd-accept-remote` option:
```
openfortivpn --pppd-accept-remote=0
```

PR #1217, opened 2024-04-29
Saml external browser

filippor:
add parameter --ext-browser-saml[=<listen port>]

- start listening on listen port (default)
- print an address to manually open in a browser 
- ( the gateway server will send browser a javascrip to call `localhost:8020/?id=<auth-id>` )
- receive the http request and extract the auth-d
- use auth-id to autenticate on address /remote/saml/auth_id?id=<auth-id>

there is also parameter --auth-id=<id>  if retrieved externally 

NEED HELP to review printed messages and update man pages

the ext-browser-saml  can be configured configuration file but in this case the port is required 
es
```
host = vpn.sample.com
port = 10443
ext-browser-saml = 8020
```


LorenzISR:
as i've mentioned in #867, imo this is not a good idea. external programms such as https://github.com/gm-vm/openfortivpn-webview are a cleaner approach to handling this -- maby look into packaging it for your distro instead?

filippor:
The advantage is that this solution could use the default browser that keep the session and if it is valid doesn't need to reautenticate. And i think it is easy to integrate in external gui or NetworkManager

exzombie:
I hope to be able to test this in the coming weeks.

I'm a regular user of code from an [unfinished MR in NetworkManager-fortisslvpn](https://gitlab.gnome.org/GNOME/yy/-/merge_requests/28) which uses the same approach as openfortipvn-webview, with the benefits of NetworkManager integration. While it's a life saver in my situation, I have to say that the approach taken in this PR would be preferable. The VPN is not very stable, and I am often on a flaky uplink, necessitating many reconnections. Typing the password and dealing with 2FA can be quite annoying.

We can debate whether this functionality belongs into openfortivpn itself instead of an external tool. I think having it in openfortivpn has the benefit of making it available to all users. For comparison, having openfortivpn-webview available doesn't do much for those of us who need NetworkManager, and while the Gnome version of the NM user interface now has an unfinished reimplementation of what openfortivpn-webview offers, other interfaces (like KDE Plasma) don't have it. In other words, everybody would need to integrate a webview, which is not trivial. On the other hand, if openfortivpn can handle this by itself, adding SAML to NM interfaces is one new checkbox away.

The way I see it, this is just another way for openfortivpn to get the cookie, complementing the stdin method. Authentication is delegated to an external tool (i.e., the default browser), same as always.

robdevops:
Working well on Ubuntu. I do strongly prefer this method because it is a much simpler command that does not require additional packages, nor the awkward pipe into sudo, nor re-authentication because it uses my existing browser session.

This is how convoluted my launch command needs to be with openfortivpn-webview:
`$ sudo whoami # cache sudo credentials because you can't type your password into a pipe`
`$ ~/openfortivpn-webview-1.2.0.AppImage --trusted-cert='sha256/+NbF2hzIunRbOynladSfCD9gj9HEKYHc4Y7h9dlGqQg=' vpn.example.com:10443 | sudo ./openfortivpn --cookie-on-stdin  --trusted-cert f8d6c5da1cc8ba755b3b29e569d49f082f608fd1c42981dce18ee1f5d946a908 vpn.example.com:10443
`

Vs this branch:
`$ sudo ./openfortivpn --ext-browser-saml vpn.example.com:10443 --trusted-cert f8d6c5da1cc8ba755b3b29e569d49f082f608fd1c42981dce18ee1f5d946a908
`

It is better in every way: quicker to install, easier to use, and saves a good 40 secs by using existing session.

robdevops:
Hi @LorenzISR 
> a bunch of code for functionality that openfortipvn-webview can already do, but better. if it gets packaged for distros that is actually a much nicer user experience

Expanding on my previous comment, the user experience is superior with this PR for the following reasons:

- Even if openfortipvn-webview were packaged, in order to continue supporting non-GUI systems, it can never be pulled in as a dependency of the openfortipvn package, so on any system with a working browser, the openfortipvn-webview method will always be more complex to install.
- The openfortipvn-webview launch command is convoluted with openfortipvn-webview, requiring shell tricks like pipes or stdin redirects, so it will always be harder to use, especially when combined with sudo and password input.
- Both methods use an external browser. With this PR, it is guaranteed the user will have browser package options that have an active maintainer, as long as they are on a supported OS.
- This PR supports hardware auth methods like YubiKey, FIDO/U2F, smart cards, and Passkeys. It is a more universal solution.
- The reuse of existing browser session cookies can significantly reduce the number of user interactions and average time to connect.

I think the last two items in particular are high value for user experience. I respectfully request you reconsider use cases where this PR adds real value.

LorenzISR:
pipes are not "tricks", they are normal features of a shell, and they have always been. you will need sudo anyways, regardless if you use webview or not. i am not using `sudo` but `doas` instead but using pipes works perfectly fine for me. openfortivpn-webview can safe the credentials (i think?), you'll just have to click the button for your SSO provider to safe them. i don't understand why openfortivpn-webview would not work with hardware auth methods?

that said, i am not a maintainer but just a random person on the internet who's worked on this problem before and this is just my opinion.

robdevops:
> pipes are not "tricks", they are normal features of a shell, and they have always been. you will need sudo anyways, regardless if you use webview or not. i am not using `sudo` but `doas` instead but using pipes works perfectly fine for me.

I'll clarify. If `sudo` is at the beginning of the command, it can cleanly prompt for keyboard input before `openfortivpn` is executed. In the `openfortivpn-webview` case where `sudo` is after a pipe, you have `sudo` writing to stdout, `openfortivpn-webview` also writing to stdout, `sudo` reading from stdin, and finally `openfortivpn` reading from stdin. The output of `sudo` and `openfortivpn-webview` gets mixed together and the `sudo` password prompt is buried above line feeds from `openfortivpn-webview`, with `openfortivpn-webview` printing to the part of the screen where the users password input indicator is supposed to be printed. It's clunky. There are ways to make this smoother, but my point is it's an inferior user experience out of the box. If further steps are required to make this smoother, then the existence of those steps themselves make for a more complicated user experience.

> i don't understand why openfortivpn-webview would not work with hardware auth methods?

I thought I read on another discussion that it didn't. Apologies if I am wrong.


> openfortivpn-webview can safe the credentials (i think?), you'll just have to click the button for your SSO provider to safe them. i don't understand why openfortivpn-webview would not work with hardware auth methods?

`openfortivpn-webview` does cache, but you may be overlooking that SAML is commonly used for single sign-on, where web apps can use the VPN initiated cookie or vice versa. That is, the browser may already be signed in before the VPN is ever used, and likewise, signing into the VPN may also sign into various web apps. Using `openfortivpn-webview` segregates the cache of the VPN from the cache of web apps, breaking the single sign-on model by resulting in more consecutive sign on interactions.

LorenzISR:
sudo _does not_ read the password from stdin.

i'd argue that re-using the sign-in from the browser might even be a security risk for some and not desired; having to sign in once for your VPN is fine imo and this is also how it works on windows.

gm-vm:
This seems to do what it is described [here](https://docs.fortinet.com/document/fortigate/7.0.0/new-features/364443/use-a-browser-as-an-external-user-agent-for-saml-authentication-in-an-ssl-vpn-connection-7-0-1).

That's neat, I don't remember having this option in my official client (and I no longer have access to the VPN to check it now), but this definitely removes the need of using a custom made browser.

@LorenzISR this seems to be a feature of the gateway that lets clients use any browser for the authentication. It is different from launching the browser from within openfortivpn.

robdevops:
@gm-vm that's correct. I use this option in the official mac client.

filippor:
If needed I can split this PR 
1 for adding the authentication method with ID and another to listen for id on port. 
But I think the changes are not so big.

herpus-derpus:
I am of the opinion this PR would vastly improve all linux user experience for everyone I work with.

 My default browser session requires me to SAML auth with a second factor each day. Currently I have to do this twice each morning since `openfortivpn-webview` requires it separately.
                                                                                
If the `openfortivpn` client could make use of my already authenticated session in my default browser, this would significantly improve the SSO experience for myself and all other linux desktop users where I work. They would be over the moon. I don't see a decrease in security for us, since we need to SAML auth daily anyway as policy means our session expire each day.

filippor:
> Presumably I could build openfortivpn from this filippor:saml-external-take2 branch ? Is that easy ?
It is not complicated to compile the instruction is in the [readmi ](https://github.com/adrienverge/openfortivpn/blob/master/README.md) on this repository

or to retrieve the token you could use this project https://github.com/filippor/XdgOpenSaml/tree/main it also use the external browser a binary that not require javacompiled for linux-x86-64  is [here](https://github.com/filippor/XdgOpenSaml/raw/main/XdgOpenSaml)

dsl400:
Without knowing this even exited and driven only by the need of a specific feature I written a [solution](https://github.com/adrienverge/openfortivpn/pull/1219) that kinda solves the same problem,  there are few extra features like specifying routes, dns entries and domain prefixes in the config file to override the configuration coming from the server. 
In my case the server uses microsoft services with 2FA. While I do undertand the need for a SAML sign in method on systems without gui I am not really sure that [openfortivpn-webview](https://github.com/gm-vm/openfortivpn-webview) is a good idea and here are some reasons:

- bloat
- security
- overcomplicated 

 I really would not push a "home made browser" over a mainstream browser. 

and .... 
> that said, i am not a maintainer but just a random person on the internet who's worked on this problem before and this is just my opinion.

itss0076:
Rather than trying to put the browser spawning and web server listening bits into the (run as root) openfortivpn process, why not use the privilege-separated interface already provided for openfortvpn-webview?

Make a new process (run unprivileged) that does the SAML dance with the VPN appliance, by shelling out to the user's browser and listening for a response on port 8020, and then outputs the "SVPNCOOKIE=..." string for piping into "sudo openfortivpn --cookie-on-stdin ...".

filippor:
The implementation as external program already exist at this repo https://github.com/filippor/XdgOpenSaml.
The idea of this pr is to integrate all the process in this software.

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+	}
+	log_info("Authenticate at %s\n", url);
+}
+
+// Function to parse HTTP request and extract parameter "id"
+static char* parse_request(const char *request) {
+	char *id_param;
+	char *query_start = strchr(request, '?');
+	if (query_start != NULL) {
+		id_param = strstr(query_start, "id=");
```
Rainer-Keller, line 47: I case there is a more complex query string in the requested url, this will find possibly also other `id=` string like

    GET /?myid=foo&id=bar

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+		if (id_param != NULL) {
+			id_param += 3; // Length of "id="
+			char *id_end = strchr(id_param, '&');
+			if (id_end == NULL) {
```
Rainer-Keller, line 51: Might `strsep` do what you want here?

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+			return strdup(id_param);
+
+		}
+	}
+	return NULL;
+}
+
+void send_response(int sockfd, const char *message) {
+	char response[MAX_REQUEST_SIZE];
+	sprintf(response, "HTTP/1.1 200 OK\r\n"
```
Rainer-Keller, line 70: Even though the `message` is coming from internal, maybe not everyone in future will be aware that there is a size restriction. At least the `sprintf` should be restricted to write out of bounds. Better would be to do a size check.

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+			"Content-Type: text/html\r\n\r\n"
+			"%s", strlen(message), message);
+	write(sockfd, response, strlen(response));
+}
+
+char* listen_for_id(struct vpn_config *cfg) {
+	int sockfd = socket(AF_INET, SOCK_STREAM, 0);
+	if (sockfd < 0) {
+		log_error("Error opening socket");
+		exit(1);
```
Rainer-Keller, line 81: This sounds a bit hard ;)
Why not `return -1` ?

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+	
+	serv_addr.sin_family = AF_INET;
+	serv_addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+	serv_addr.sin_port = htons(cfg->listen_port);
+
+	int opt = 1;
+	if (setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &opt, sizeof(opt)) < 0) {
+		log_error("Error setting SO_REUSEADDR");
+	}
+	if (setsockopt(sockfd, SOL_SOCKET, SO_REUSEPORT, &opt, sizeof(int)) < 0) {
```
Rainer-Keller, line 94: I guess this was meant to be `sizeof(opt)` like above.

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if (bind(sockfd, (struct sockaddr*) &serv_addr, sizeof(serv_addr)) < 0) {
+		perror("Error on binding");
+		exit(1);
+	}
+
+
+	
+	listen(sockfd, 1);
```
Rainer-Keller, line 105: Error check missing

src/idlistener.h
```diff
@@ -0,0 +1,15 @@
+/*
+ * cookieRetriever.h
```
Rainer-Keller, line 2: Is there a name mistmatch?

Also the file name and function name are every generic. It would improve understandability if it whould somehow mention SAML somewhere.

src/main.c
```diff
@@ -604,6 +613,24 @@ int main(int argc, char *argv[])
 				free(cookie);
 				break;
 			}
+			if (strcmp(long_options[option_index].name,
+					"ext-browser-saml") == 0) {
+				long port = 8020;
+				if (optarg != NULL) {
+					port = strtol(optarg, NULL, 0);
+					if (port < 1 || port > 65535) {
+						log_error("Specify a valid listen port or omit for parameter ext-browser-saml\n");
```
Rainer-Keller, line 622: Suggestion:

`Specify a valid listen port for parameter ext-browser-saml or omit it to use the default.`

src/main.c
```diff
@@ -604,6 +613,24 @@ int main(int argc, char *argv[])
...
+						goto user_error;
+					}
+				}
+				cli_cfg.listen_port = (uint16_t)port;
+				break;
+			}
+			if (strcmp(long_options[option_index].name, "auth-id") == 0) {
+				free(cli_cfg.auth_id);
+				cli_cfg.auth_id = strdup(optarg);
```
Rainer-Keller, line 631: I find it handy to have at least a zero length check to catch shell mistakes like empty variables when calling as early as possible.

src/main.c
```diff
@@ -732,6 +759,15 @@ int main(int argc, char *argv[])
 	if (cfg.otp[0] != '\0')
 		log_debug("One-time password = \"%s\"\n", cfg.otp);
 
+	if (cfg.listen_port != 0){
+	    if(cfg.auth_id){
+	    	log_warn("auth-id will be ignored conflict with --ext-browser-saml");
+	    }
+		log_debug("Will listen on port \"%d\" for authentication id \n", cfg.listen_port);
```
Rainer-Keller, line 766: There's about the same message in `idlistener.c:listen_for_id`

src/main.c
```diff
@@ -732,6 +759,15 @@ int main(int argc, char *argv[])
...
+		free(cfg.auth_id);
+		cfg.auth_id = listen_for_id(&cfg);
+	}
+	
 	if (geteuid() != 0) {
 		log_error("This process was not spawned with root privileges, which are required.\n");
 		ret = EXIT_FAILURE;
```
Rainer-Keller, line 773: (Sorry, for putting it here, but I didn't find out how to comment on files that there not changed.)

This MR is missing changes to `README.md` and `doc/openfortivpn.1.in`

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+	}
+	close(sockfd);
+
+	// Read HTTP request from client
+	char buffer[MAX_REQUEST_SIZE];
+	bzero(buffer, MAX_REQUEST_SIZE);
+	read(newsockfd, buffer, MAX_REQUEST_SIZE - 1);
+	log_debug("Received HTTP request:\n%s\n", buffer);
+
+	char *id = parse_request(buffer);
```
Rainer-Keller, line 127: I'd like to have a check if the id string has zero length, just in case something went wrong. I won't make sense to continue in that case.

src/idlistener.c
```diff
@@ -0,0 +1,142 @@
...
+
+static void print_url(struct vpn_config *cfg) {
+	char url[512];
+	snprintf(url, sizeof(url), "https://%s:%d/remote/saml/start?redirect=1",
+			cfg->gateway_host, cfg->gateway_port);
+
+	if (cfg->realm[0] != '\0') {
+		strncat(url, "&realm=", sizeof(url) - 1);
+		char *dt = url + strlen(url);
+		url_encode(dt, cfg->realm);
```
Rainer-Keller, line 37: In the documentation of `url_encode` it is stated that the buffer size must be at least 3x the size of `cfg->realm` +1. I can't find that reflected in the code here.

mrbaseman:
a similar PR #1241 has been merged now, anyhow, thanks for the contribution. It seems the other work was at least in parts inspired by this one

PR #1219, opened 2024-05-19
Listen for SAML login requests

dsl400:
When started with `--saml-login` the program creates a http server that accepts SAML login requests 
The server will remain open and refresh the connection whenever a request is made

The solution needs a little more work on the logging part.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -534,6 +534,9 @@ void merge_config(struct vpn_config *dst, struct vpn_config *src)
 		free(dst->cookie);
 		dst->cookie = src->cookie;
 	}
+	if(src->saml_port != 0){
+		dst->saml_port = src->saml_port;
+	}
 	if (src->pinentry) {
 		free(dst->pinentry);
 		dst->pinentry = src->pinentry;
--- a/src/config.c
+++ b/src/config.c
@@ -28,6 +28,7 @@
 
 #include "config.h"
 #include "log.h"
+#include "ipv4.h"
 
 #include <openssl/x509.h>  /* work around OpenSSL bug: missing definition of STACK_OF */
 #include <openssl/tls1.h>
@@ -37,6 +38,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
 
 const struct vpn_config invalid_cfg = {
 	.gateway_host = {'\0'},
@@ -54,7 +56,9 @@ const struct vpn_config invalid_cfg = {
 	.iface_name = {'\0'},
 	.sni = {'\0'},
 	.set_routes = -1,
+	.routes = NULL,
 	.set_dns = -1,
+	.dns = NULL,
 	.pppd_use_peerdns = -1,
 #if HAVE_RESOLVCONF
 	.use_resolvconf = -1,
@@ -88,6 +92,9 @@ const struct vpn_config invalid_cfg = {
 	.hostcheck = NULL,
 	.check_virtual_desktop = NULL,
 };
+
+
+
 
 /*
  * Adds a sha256 digest to the list of trusted certificates.
@@ -417,6 +424,23 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			cfg->user_cert = strdup(val);
 			if (strncmp(cfg->user_cert, "pkcs11:", 7) == 0)
 				cfg->use_engine = 1;
+		} else if (strcmp(key, "dns") == 0) {
+			Node *newNode = (Node *)malloc(sizeof(Node));
+			newNode->value = val;
+			newNode->next = cfg->dns;
+			cfg->dns = newNode;
+			continue;
+		} else if (strcmp(key, "route") == 0) {			
+    		Node *newNode = (Node *)malloc(sizeof(Node));
+			newNode->value = cidr_to_ip_mask(val);
+			newNode->next = cfg->routes;
+			cfg->routes = newNode;
+			log_warn("add route: %s\n", val);
+			continue;
+		} else if (strcmp(key, "saml-login") == 0) {
+			free(cfg->saml_port);
+			cfg->saml_port = strdup(val);
+			continue;
 		} else if (strcmp(key, "user-key") == 0) {
 			free(cfg->user_key);
 			cfg->user_key = strdup(val);
--- a/src/config.c
+++ b/src/config.c
@@ -59,6 +59,7 @@ const struct vpn_config invalid_cfg = {
 	.routes = NULL,
 	.set_dns = -1,
 	.dns = NULL,
+	.domain_suffix = {'\0'},
 	.pppd_use_peerdns = -1,
 #if HAVE_RESOLVCONF
 	.use_resolvconf = -1,
@@ -425,6 +426,10 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			if (strncmp(cfg->user_cert, "pkcs11:", 7) == 0)
 				cfg->use_engine = 1;
 		} else if (strcmp(key, "dns") == 0) {
+			free(cfg->domain_suffix);
+			cfg->domain_suffix = val;
+			continue;
+		} else if (strcmp(key, "dns") == 0) {
 			Node *newNode = (Node *)malloc(sizeof(Node));
 			newNode->value = val;
 			newNode->next = cfg->dns;
@@ -435,11 +440,10 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			newNode->value = cidr_to_ip_mask(val);
 			newNode->next = cfg->routes;
 			cfg->routes = newNode;
-			log_warn("add route: %s\n", val);
 			continue;
 		} else if (strcmp(key, "saml-login") == 0) {
 			free(cfg->saml_port);
-			cfg->saml_port = strdup(val);
+			cfg->saml_port = atol(val);
 			continue;
 		} else if (strcmp(key, "user-key") == 0) {
 			free(cfg->user_key);
--- a/src/http.c
+++ b/src/http.c
@@ -870,6 +870,9 @@ static int parse_xml_config(struct tunnel *tunnel, const char *buffer)
 	if (!gateway)
 		log_warn("No gateway address, using interface for routing\n");
 
+
+	
+
 	// The dns search string
 	val = buffer;
 	while ((val = xml_find('<', "dns", val, 2))) {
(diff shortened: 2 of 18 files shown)


src/config.c
```diff
@@ -417,6 +425,26 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			cfg->user_cert = strdup(val);
 			if (strncmp(cfg->user_cert, "pkcs11:", 7) == 0)
 				cfg->use_engine = 1;
+		} else if (strcmp(key, "dns") == 0) {
+			free(cfg->domain_suffix);
+			cfg->domain_suffix = val;
+			continue;
+		} else if (strcmp(key, "dns") == 0) {
```
cpfeiffer, line 432: AFAICS this block is never reached due to the same if-clause above.

src/config.c
```diff
@@ -417,6 +425,26 @@ int load_config(struct vpn_config *cfg, const char *filename)
...
+			continue;
+		} else if (strcmp(key, "route") == 0) {			
+    		Node *newNode = (Node *)malloc(sizeof(Node));
+			newNode->value = cidr_to_ip_mask(val);
+			newNode->next = cfg->routes;
+			cfg->routes = newNode;
+			continue;
+		} else if (strcmp(key, "saml-login") == 0) {
+			free(cfg->saml_port);
+			cfg->saml_port = atol(val);
```
cpfeiffer, line 446: Are we freeing an int here? And why atol instead of atoi when saml_port is actually an int?

IslandNolan:
Any updates on this? Just had to go through the rather cumbersome SAML flow with openfortivpn-webview and realized that this branch obsoletes that method entirely. 

Hoping this can be merged soon.

src/http.c
```diff
@@ -621,6 +621,44 @@ static int try_otp_auth(struct tunnel *tunnel, const char *buffer,
...
+ */
+int saml_login(struct tunnel *tunnel)
+{
+	log_debug("SAML login\n");
+
+	int ret;
+	ssl_connect(tunnel);
+
+	char uri[1024];
+	snprintf(uri, sizeof(uri), "/remote/saml/auth_id?id=%s", tunnel->config->saml_session_id);
```
Rainer-Keller, line 638: In case the session ID gets very long, the resulting string will not be null terminated anymore.
This lead to other parts dealing with this string behave wrong.
There should either be a check if the string fits, or at least force it to be null terminated.

src/http.c
```diff
@@ -621,6 +621,44 @@ static int try_otp_auth(struct tunnel *tunnel, const char *buffer,
...
+	char *response;
+	uint32_t response_size = 0;
+	ret = http_request(tunnel, "GET", uri, "", &response, &response_size);
+	if(ret != 1 || response_size <= 15) return ret;
```
Rainer-Keller, line 642: If  `ret` was 1, but the response too short, returning `ret` will indicate success when it wasn't.

src/http.c
```diff
@@ -621,6 +621,44 @@ static int try_otp_auth(struct tunnel *tunnel, const char *buffer,
...
+		log_error("SAML login failed: %s\n", response);
+		return ret;
+	}
+	auth_get_cookie(tunnel, response, response_size);
+	if (ret == ERR_HTTP_NO_COOKIE){
+		log_error("SAML login failed: no cookie\n");
+		return ret;
+	}	
+
+	// free(response);
```
Rainer-Keller, line 653: Is this not needed?

src/http_server.c
```diff
@@ -0,0 +1,155 @@
...
+    // dup2(new_socket, STDOUT_FILENO);
+    // dup2(new_socket, STDERR_FILENO);
+
+
+
+    char request[1024];
+    ssize_t read_result = read(new_socket, request, sizeof(request));
+
+    // Check for '=id' in the response
+    if (read_result < 0 || strlen(request) < 10 || strncmp(request, "GET /?id=", 9) != 0) {
```
Rainer-Keller, line 44: If the received request from the server is larger than the buffer, the result will not be null-terminated, causing `strlen` to behave wrong.

src/http_server.c
```diff
@@ -0,0 +1,155 @@
...
+    char *id_end = memchr(&request[id_start], ' ', 1000);
+
+    if (id_end == NULL) {
+        log_error("Bad request format\n");
+        return -1;
+    }
+
+    int id_length = id_end - &request[id_start];
+
+    if(id_length == 0 || id_length > 1000) {
```
Rainer-Keller, line 60: The 1000 has to match the id buffer size of 1024.
In order to prevent future mismatches I would suggest to use a define for the buffer size.

src/http_server.c
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    strncpy(id, &request[id_start], id_length);
+
+    for (int i = 0; i < id_length; i++) {
+        if (isalnum(id[i]) || id[i] == '-') continue;
+        log_error("Invalid id format\n");
+        return -1;
+    }
+    // close(new_socket);
```
Rainer-Keller, line 72: Why not closing the socket? I can't find any other code location where this is done.

src/http_server.c
```diff
@@ -0,0 +1,155 @@
...
+    struct vpn_config *config = (struct vpn_config *)void_config;
+    
+    int server_fd, new_socket;
+    struct sockaddr_in address;
+    int opt = 1;
+    int addrlen = sizeof(address);
+    long saml_port = config->saml_port;
+
+    // Creating socket file descriptor
+    if ((server_fd = socket(AF_INET, SOCK_STREAM, 0)) == 0) {
```
Rainer-Keller, line 90: The socket does not seem to be shut down anywhere.
Also it looks like the socket is not closed in the error returns after bind and  listen.

src/http_server.c
```diff
@@ -0,0 +1,155 @@
...
+        return NULL;
+    }
+    log_info("Listening for saml login on port: %d\n", saml_port);
+    int running = 1;
+
+    pthread_t vpn_thread = 0;
+
+
+    while(running) {
+        if ((new_socket = accept(server_fd, (struct sockaddr *)&address, (socklen_t*)&addrlen)) < 0) {
```
Rainer-Keller, line 122: From my understanding, this loop will be run forever?
This would mean the port will be accepting connections for the entire runtime.
I am not too famliliar with the SAML protocol. If there is any malicious browser tab, they
could just connect to this server at any time and trigger further VPN tunnels.

src/http.c
jimis: `snprintf()` will truncate and null-terminate if it's too long. 
An undetected truncated string might be an issue, so better check the return value anyway.

src/config.c
```diff
@@ -417,6 +425,26 @@ int load_config(struct vpn_config *cfg, const char *filename)
 			cfg->user_cert = strdup(val);
 			if (strncmp(cfg->user_cert, "pkcs11:", 7) == 0)
 				cfg->use_engine = 1;
+		} else if (strcmp(key, "dns") == 0) {
+			free(cfg->domain_suffix);
+			cfg->domain_suffix = val;
+			continue;
+		} else if (strcmp(key, "dns") == 0) {
+			Node *newNode = (Node *)malloc(sizeof(Node));
```
Rainer-Keller, line 433: I can't find the location where this is freed.

src/config.c
```diff
@@ -417,6 +425,26 @@ int load_config(struct vpn_config *cfg, const char *filename)
...
+			newNode->value = val;
+			newNode->next = cfg->dns;
+			cfg->dns = newNode;
+			continue;
+		} else if (strcmp(key, "route") == 0) {			
+    		Node *newNode = (Node *)malloc(sizeof(Node));
```
Rainer-Keller, line 439: I can't find the location where this is freed.

src/config.c
Rainer-Keller: I guess the string being checked for in the above condition should have been like "domain-suffix", but not "dns".

src/http.c
```diff
@@ -844,9 +885,29 @@ static int parse_xml_config(struct tunnel *tunnel, const char *buffer)
 		}
 	}
 
+	Node *route = tunnel->config->routes;
+	while(route){
```
Rainer-Keller, line 889: In the code creating the nodes, there is

    newNode->next = cfg->routes;
    cfg->routes = newNode;

Using `while(route)` assumes there will be a null pointer somewhere. But for me it seems the last item loops back to the beginning.

src/http.c
```diff
@@ -844,9 +885,29 @@ static int parse_xml_config(struct tunnel *tunnel, const char *buffer)
...
+		Node *next =  route->next;
+		IP_Mask *_route = (IP_Mask *)route->value;
+		log_info("add route: %s %s\n", _route->ip, _route->mask );
+		ipv4_add_split_vpn_route(tunnel, _route->ip, _route->mask, gateway);
+		route = next;
+	}
+
+	Node *dns = tunnel->config->dns;
+	while(dns){
```
Rainer-Keller, line 898: Same issue with the linked list as above.

src/http.c
```diff
@@ -844,9 +885,29 @@ static int parse_xml_config(struct tunnel *tunnel, const char *buffer)
...
+		Node *next =  dns->next;
+		log_warn("add dns: %s\n", dns->value);
```
Rainer-Keller, line 900: For the routes this was `log_info` is there are reason to make a warning here?

src/main.c
```diff
@@ -738,6 +757,19 @@ int main(int argc, char *argv[])
...
+	if(cfg.saml_port != 0) {
+		pthread_t server_thread;
+
+		if(pthread_create(&server_thread, NULL, start_http_server, &cfg) != 0){
+			log_error("Failed to create saml login server thread\n");
+			// ret = EXIT_FAILURE;
+			goto exit;
+		}
+		// log_debug("Running http server on port %d\n", cfg.saml_port);
+		while(get_sig_received() == 0);
```
Rainer-Keller, line 769: Won't this burn a 100% CPU while waiting?

src/main.c
```diff
@@ -604,6 +610,19 @@ int main(int argc, char *argv[])
 				free(cookie);
 				break;
 			}
+			if (strcmp(long_options[option_index].name,
+			           "saml-login") == 0) {
+				long port = 8020;
+				if(optarg != NULL){
+					port =	strtol(optarg, NULL, 0);
+				}
+				if (port < 0 || port > UINT_MAX) {
```
Rainer-Keller, line 619: Why checking against UINT_MAX? Ports can't be higher than 65535.

src/log.c
```diff
@@ -40,6 +42,9 @@ struct log_param_s {
 	int syslog_prio;
 };
 
+static int initial_stdout;
```
Rainer-Keller, line 45: Can anyone explain what this (and the other related changes below) is supposed to fix?

Rainer-Keller:
I've created a counter proposal based on this change that has all changes unrelated to SAML removed and the open issues resolved.

mrbaseman:
thanks for the contributions. PR #1241 which references this one, has been merged now.

PR #1220, opened 2024-05-21
fix config file name in service configuration

dsl400:
The config path is `etc/openfortivpn/config`
The service does not start because it tries to load `etc/openfortivpn/config.conf`

### DIFF
--- a/lib/systemd/system/openfortivpn@.service.in
+++ b/lib/systemd/system/openfortivpn@.service.in
@@ -9,7 +9,7 @@ Documentation=https://github.com/adrienverge/openfortivpn/wiki
 [Service]
 Type=notify
 PrivateTmp=true
-ExecStart=@BINDIR@/openfortivpn -c @SYSCONFDIR@/openfortivpn/%I.conf
+ExecStart=@BINDIR@/openfortivpn -c @SYSCONFDIR@/openfortivpn/%I
 Restart=on-failure
 OOMScoreAdjust=-100
 


DimitriPapadopoulos:
Please provide background information in the commit message. Why should the confg file be `%I` instead of `%I.conf`?

DimitriPapadopoulos:
@dsl400 Why do you want to remove the `.conf` file extension?

dsl400:
@DimitriPapadopoulos  The config is stored at etc/openfortivpn/config
I need to run openfortivpn as service and it does not start because it tries to load etc/openfortivpn/config.config

The fastest option was to remove .conf

DimitriPapadopoulos:
The file name is generated from `/etc/openfortivpn/%I.conf`, but I don't see why the value of `%I` would be `config` in your case.

dsl400:
Sorry. It tries to load `config.conf`

DimitriPapadopoulos:
And?

dsl400:
> And? Ask yourself why first?

yep .... It is what it is .... a strange world out there.

You could just spare us the pain and uncover the secret or dismiss the PR and move on with your life leaving  us to live in the eternal pain caused by not knowing this

DimitriPapadopoulos:
I cannot recall myself. I only know this pull request cannot be merged without any form of explanation: it breaks existing behaviour for the sake of an obscure use case withoàut any explanation. I hoped you could look into this question since you opened an issue.

dsl400:
`root@desktop:/home/dsl400# systemctl start openfortivpn@config.service
Job for openfortivpn@config.service failed because the control process exited with error code.
See "systemctl status openfortivpn@config.service" and "journalctl -xeu openfortivpn@config.service" for details.
root@desktop:/home/dsl400# systemctl status openfortivpn@config.service
× openfortivpn@config.service - OpenFortiVPN for config
     Loaded: loaded (/lib/systemd/system/openfortivpn@.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Mon 2024-06-03 13:00:08 EEST; 2s ago
       Docs: man:openfortivpn(1)
             https://github.com/adrienverge/openfortivpn#readme
             https://github.com/adrienverge/openfortivpn/wiki
    Process: 6356 ExecStart=/usr/local/bin/openfortivpn -c /usr/local/etc/openfortivpn/config.conf (code=e>
   Main PID: 6356 (code=exited, status=254)
        CPU: 4ms

Jun 03 13:00:08 desktop systemd[1]: openfortivpn@config.service: Main process exited, code=exited, status=>
Jun 03 13:00:08 desktop systemd[1]: openfortivpn@config.service: Failed with result 'exit-code'.
Jun 03 13:00:08 desktop systemd[1]: Failed to start OpenFortiVPN for config.
Jun 03 13:00:08 desktop systemd[1]: openfortivpn@config.service: Scheduled restart job, restart counter is>
Jun 03 13:00:08 desktop systemd[1]: Stopped OpenFortiVPN for config.
Jun 03 13:00:08 desktop systemd[1]: openfortivpn@config.service: Start request repeated too quickly.
Jun 03 13:00:08 desktop systemd[1]: openfortivpn@config.service: Failed with result 'exit-code'.
Jun 03 13:00:08 desktop systemd[1]: Failed to start OpenFortiVPN for config.
`

mrbaseman:
Maybe the point is an inconsistency between the naming convention in the provided [systemd service template file](https://github.com/adrienverge/openfortivpn/blob/371edb5cdbdc53ca1c952cfc180318634e396015/lib/systemd/system/openfortivpn%40.service.in#L12l) where the suffix `.conf` is used and the default config mentioned in the [man page](https://github.com/adrienverge/openfortivpn/blob/371edb5cdbdc53ca1c952cfc180318634e396015/doc/openfortivpn.1.in#L69) or in the [help message](https://github.com/adrienverge/openfortivpn/blob/371edb5cdbdc53ca1c952cfc180318634e396015/src/main.c#L115). 

Usually, the expectation is, that the default `config` file is adapted for different vpn connections, and of course, if a service shall be run, the appropriate configuration file for that particular service should be supplied. This is done by a systemd service template, where the parameter `%I` is replaced by the instance of this service, in order to pick up the correct config file. 

At both places, manual action is required: The sample `config` needs to be copied to one or more specific config files, and filled in with specific connection parameters. The systemd service template needs to be linked to the appropriate service name. Both names should match. Now, the traditional sample config file name for `openfortivpn` has always been `config`, whereas systemd would expect config files to be named with the suffix `.conf` - however, this is not a must. We could either remove the `.comf`suffix in the service template, or we continue to expect users copying the sample `config` file to something like `mycorp.conf`. Or, 3rd option, append `.conf` to the sample config file name, but I guess package maintainers might not be amused about that move.

PR #1224, opened 2024-05-30
configure.ac: fix header checks for macOS

barracuda156:
These headers are available, but they won’t be found on earlier macOS versions due to requirement for inclusion order.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -58,7 +58,6 @@ fcntl.h \
 getopt.h \
 ifaddrs.h \
 netdb.h \
-net/if.h \
 netinet/in.h \
 netinet/tcp.h \
 pthread.h \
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Out of curiosity, which macOS versions fail?

barracuda156:
> Out of curiosity, which macOS versions fail?

AFAIK, < 10.9

DimitriPapadopoulos:
Not exactly recent macOS versions, they have reached EOL in 2015. But then why not support them if minimal changes are required.

configure.ac
```diff
@@ -86,7 +86,7 @@ util.h \
 ])
 
 # Checks for optional header files with prerequisites of other headers.
-AC_CHECK_HEADERS([net/if_var.h],[],[],[#include <net/if.h>])
```
DimitriPapadopoulos, line 89: After #701, `net/if_var.h` is not needed any more . You may remove this paragraph altogether.

barracuda156: Ok, I will rebase today, thanks.

DimitriPapadopoulos: I solved the conflict and merged from the web interface. Thank you very much for this improvement of portability @barracuda156.

barracuda156: Thank you! (Sorry, I got distracted by other stuff.)

Issue #1233, opened 2024-07-19
Cannot connect to Fortigate 7.4.4 SSLVPN "Error reading from SSL connection"


Voriaz:
Hi,

I have an issue connecting to a Fortigate SSLVPN using openfortivpn.
The issue appears right after an upgrade of the Fortigate (7.2.8 -> 7.4.4). It was working perfectly fine before (launched regularly via cron task).

OpenfortiVPN verbose logs say: 
```Error reading from SSL connection (Connection closed). 
Error issuing /remote/login request 
Could not authenticate to gateway. Please check the password, client certificate, etc. 
SSL error (-4) 
Closed connection to gateway."
```

I suspected a TLS issue, so I upgraded openfortivpn to latest version (manual compiled install) on a recent Debian12 VM without success. I also tested on Ubuntu 24.04 with lastest apt package.

When inspecting the Fortigate debug log, I see `do_http_validate:447 Content-Length (11) on uri (/remote/login) not allowed.`

Is this something related to openfortivpn not up to date with lastest fortiVPN specs or is it a bug in latest fortiOS version ?

Please see attached redacted logs from openfortivpn and fortigate.
[openfortivpn.log](https://github.com/user-attachments/files/16310510/openfortivpn.log)
[FWVPN.log](https://github.com/user-attachments/files/16310511/DC2FWVPN.log)

Thanks,

A.

adrienverge:
Sure, we could release openfortivpn 1.23.0 including this patch, and SAML login authentication.

I just opened https://github.com/adrienverge/openfortivpn/pull/1266 for that.

gtache:
By the way, I'm on Fedora 41 but I don't really know how the release process works. Are you the one who should release it there or are there maintainers on the Fedora side who have to do it?

adrienverge:
Hello Guillaume,

It's not only me: any volunteer maintainer can do it :)

For this one, I just did it, see https://bodhi.fedoraproject.org/updates/FEDORA-2025-999492ef10

gtache:
Thanks! Sorry for the late reply, I was waiting for the version to change on [the packages page](https://packages.fedoraproject.org/pkgs/openfortivpn/openfortivpn/) before trying.

[...]

PR #1234, opened 2024-07-22
Patch certificate login error for FortiOS 7.4.4

Voriaz:
Hi,

After I encounter an issue on my Fortigate following FortiOS upgrade from 7.2.8 to 7.4.4 (see the issue I openned: [here](https://github.com/adrienverge/openfortivpn/issues/1233)). 
My investigation leads me to think that now Fortigate > 7.2.4 doesn't support any content when issuing login request on "/remote/login" and so "Content-length" must be 0. This is confirmed by the debug logs on the firewall saying `do_http_validate:447 Content-Length (11) on uri (/remote/login) not allowed` and by observing Forticlient requests content on Windows using Burp proxy that have no content.

So I'm proposing this pull request to fix the issue on newer Fortigate version where GET request on "/remote/login" is now sending nothing instead of "cert=&nup=1".

This has been tested on a Fortigate 500E running FortiOS 7.4.4.

BR,

A.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -651,7 +651,7 @@ int auth_log_in(struct tunnel *tunnel)
 	tunnel->cookie[0] = '\0';
 
 	if (username[0] == '\0' && tunnel->config->password[0] == '\0') {
-		snprintf(data, sizeof(data), "cert=&nup=1");
+		snprintf(data, sizeof(data), "");
 		ret = http_request(tunnel, "GET", "/remote/login",
 		                   data, &res, &response_size);
 	} else {


DimitriPapadopoulos:
The modified code had been introduced in #464 / 320f95f and #631 / abf1a95 to add smartcard support. It is triggered only if you do not pass a password:
```
openfortivpn -vv **********:8443 --user-cert=cert --user-key=key --trusted-cert ****************************
```
I don't know why `cert=&nup=1` ("_no user password_"?) had initially been added. The pull request doesn't get into such details. My only concern is that older versions of FortiOS might require options `nup=1` or `cert=`, while FortiOS 7.4 doesn't like it. But then, if current versions of FortiClient don't send these options, I guess the change is pretty safe.

src/http.c
```diff
@@ -651,7 +651,7 @@ int auth_log_in(struct tunnel *tunnel)
 	tunnel->cookie[0] = '\0';
 
 	if (username[0] == '\0' && tunnel->config->password[0] == '\0') {
-		snprintf(data, sizeof(data), "cert=&nup=1");
+		snprintf(data, sizeof(data), "");
```
DimitriPapadopoulos, line 654: I guess this would have been sufficient instead:
```c
		data[0] = `\0`;
```
I'll change that in a later PR though.

DimitriPapadopoulos:
Thank you @Voriaz.

src/http.c
DimitriPapadopoulos: The later PR is #1235.

PR #1264, opened 2025-02-05
Add support for ifup hook.

DirtyHairy:
This PR adds support for executing a command once the tunnel has been established.


DimitriPapadopoulos:
It might be preferable to use an external script such as [vpnc-script](https://gitlab.com/openconnect/vpnc-scripts) for all routing/DNS settings. When needed, alternative scripts such as [vpn-slice](https://github.com/dlenski/vpn-slice) can easily replace the default script.

DirtyHairy:
Yeah, I've seen #678 . However, my use case is much simpler: I don't want to apply the configuration pushed over the tunnel at all, and I know exactly what routes and nameservers I want to set up, I just need a convenient trigger to do so.

PR #1274, opened 2025-03-10
Add support for split DNS

elsoa-invitech:
Forti can send multiple split dns config in xml:

    <split-dns domains='office.intra,dmz.intra' dnsserver1='10.10.20.5' dnsserver2='10.10.20.6' /><split-dns domains='others.intra' dnsserver1='172.28.1.30' dnsserver2='172.28.1.31' />

This means, use one dns server for domain A+B, and use another dns server for domain C.

This patch parse the /remote/fortisslvpn_xml response to get the split-dns configuration, and pass the info to envs, same logic as ip4 routes.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -899,6 +899,40 @@ static int parse_xml_config(struct tunnel *tunnel, const char *buffer)
 		}
 	}
 
+	// The it could be split-dns instead of global dns
+	val = buffer;
+	while ((val = xml_find('<', "split-dns", val, 2))) {
+		char *domains, *dnsserver1, *dnsserver2;
+
+		domains = xml_get(xml_find(' ', "domains=", val, 1));
+		if (!domains) {
+			log_warn("No domain names for a split-dns\n");
+			continue;
+		}
+
+		dnsserver1 = xml_get(xml_find(' ', "dnsserver1=", val, 1));
+		if (!dnsserver1) {
+			log_warn("No dnsserver1 for a split-dns\n");
+			free(domains);
+			continue;
+		}
+
+		dnsserver2 = xml_get(xml_find(' ', "dnsserver2=", val, 1));
+		if (!dnsserver2) {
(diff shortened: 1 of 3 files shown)


elsoa-invitech:
pppd gets this envs (same as VPN_ROUTE_*), and setup dnsmasq to handle multi splitsdns situation. (systemd-resolved doest support splitdns in same interface :( )

DimitriPapadopoulos:
This doesn't work with plain openfortivpn.

elsoa-invitech:
what is not working for you? have you `<split-dns ...>` settings in your fortissl_xml?

DimitriPapadopoulos:
openfortivpn handles DNS internally by default.

elsoa-invitech:
> openfortivpn handles DNS internally by default.

Sure. if you run openfortivpn in standalone.

With network-manager (gui), ip+routing+dns settings are handled by NM. Thats why ip/dns/routing rules are passed to envvars, and pickup by the linked networkmanager plugin. With this patch splitdns infos are set to env too.

DimitriPapadopoulos:
Because this is the repository for the standalone openfortivpn, adding features unrelated to the standalone case is not really an option. Let me try to find some common ground.

My vision for the long term evolution of openfortivpn is as follows:
1. Drop `pppd` and use embedded PPP code instead.
2.  Switch to an external script such as [`vpnc-script`](https://github.com/cloudflare/vpnc-scripts) to set up routing and DNS parameters when the tunnel goes up and down. The script will handle IPv6, split DNS, systemd. End-users could write their own script to customize routing and DNS.
3. We could even split openfortivpn into a library and a CLI, with NM using only the library.

Because `vpnc-script` expects parameters as environment variables, setting up environment variables doesn't sound bad. However, environment variables should be identical to the ones expected by `vpnc-script` to allow a smooth future transition to `vpnc-script`. How about having a look at that?

PR #1276, opened 2025-03-31
Ability to specify SAML host

ignas2526:
Add the ability to specify the host on which SAML webserver is listening. The use case for this is when the openfortivpn runs on in a docker container or on a router, for example.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -92,6 +92,7 @@ struct vpn_config {
 	int			password_set;
 	char			otp[OTP_SIZE + 1];
 	char			*cookie;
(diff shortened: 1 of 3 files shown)


flokol120:
I'd really love to see this feature merged. The current behavior (hardcoding the SAML callback server to listen on localhost) is a show stopper for using `--saml-login` in Dockerized environments like [adrienverge/openfortivpn](https://github.com/adrienverge/openfortivpn). Without the ability to change the listen address, it's not possible to run the client in a container without resorting to host networking, which isn't ideal in many setups - hope it gets merged soon!

PR #1279, opened 2025-04-23
add --password-on-stdin

### LINKED ISSUE #1063: [EHN] Read secret from stdin
Hi.
I don't know if it's possible, security-wise, but it would be nice to have an option to read the secret via stdin, so I can fetch it from our key store.

i.e.:

```
bw search something | openfortivpn gw:port --secret-stdin
```

mafgh:
This allows the password to be read from stdin, e.g. from a password manager. Closes #1063.


DimitriPapadopoulos:
Not sure if this PR is the place to address these issues, but:
1. What should happen if we pass `--password-on-stdin` multiple times? Right now _openfortivpn_ attempts to read the password multiple times. I believe it should only try once.
2. When omitting the server or passing invalid arguments, I believe _openfortivpn_ should exitbefore asking for the password.

```console
$ openfortivpn --password-on-stdin --password-on-stdin
<password1>
<password2>
WARN:   Could not load configuration file "/volatile/local/opefortivpn/etc/openfortivpn/config" (No such file or directory).
ERROR:  Specify a valid host:port couple.
Usage: openfortivpn [<host>[:<port>]] [-u <user>] [-p <pass>] [--password-on-stdin]
[...]
$ 
```

The `--cookie-on-stdin` option has the same issues, so I am OK with fix both options in a different PR.

mafgh:
How about we just set a flag during argument parsing:

```
if (strcmp(long_options[option_index].name, "password-on-stdin") == 0) {
  password_from_stdin = 1;
  break;
}
```
And then, just before this line:

https://github.com/adrienverge/openfortivpn/blob/912f1af06f598555d7b4451e779cc2e68bd865b7/src/main.c#L673

read from stdin, etc.

DimitriPapadopoulos:
The flag will fix issue 1, but not issue 2. And there's an additional issue:
3. Option `--password` is not compatible with option `--password-on-stdin`.

Again, `--cookie-on-stdin` shares this issue. Perhaps we should defer and address these issues for both `--password-on-stdin` and `--cookie-on-stdin` in a later PR.

mafgh:
I have tried to fix 1. and 3.

PR #1289, opened 2025-06-27
Add OpenSSL 3.0+ PKCS#11 support using OSSL_STORE API

ssoftpro:
- Supports both OpenSSL < 3.0 (ENGINE API) and OpenSSL 3.0+ (Provider API)
- Implements proper PKCS#11 certificate and key loading using OSSL_STORE

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -1158,10 +1158,12 @@ int ssl_connect(struct tunnel *tunnel)
 	/* OpenSSL 3.0+ provider-based PKCS#11 support */
 	if (tunnel->config->use_engine > 0) {
 		// Debug: Print the certificate path/URI being used
-		log_debug_details("Attempting to load certificate from: %s\n", tunnel->config->user_cert);
+		log_debug_details("Attempting to load certificate from: %s\n",
+		                  tunnel->config->user_cert);
 
 		// Use OSSL_STORE to load certificate and private key from PKCS#11
-		OSSL_STORE_CTX *store_ctx = OSSL_STORE_open(tunnel->config->user_cert, NULL, NULL, NULL, NULL);
+		OSSL_STORE_CTX *store_ctx = OSSL_STORE_open(tunnel->config->user_cert,
+		                            NULL, NULL, NULL, NULL);
 
 		if (!store_ctx) {
 			log_error("PKCS11 OSSL_STORE_open failed: %s\n",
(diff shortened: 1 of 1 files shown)


src/tunnel.c
```diff
@@ -1149,7 +1154,88 @@ int ssl_connect(struct tunnel *tunnel)
...
+                break;
+            }
+        }
+
+        OSSL_STORE_close(store_ctx);
+
+        // Check if we successfully loaded both certificate and private key
+        if (!cert) {
+            log_error("PKCS11: Could not load certificate from store\n");
+            goto err_ssl_context;
```
DimitriPapadopoulos, line 1200: Can't it happen that we have `lack` cert but have `pkey`?

```suggestion
            X509_free(pkey);
            goto err_ssl_context;
```

src/tunnel.c
```diff
@@ -1149,7 +1154,88 @@ int ssl_connect(struct tunnel *tunnel)
...
+
+        X509 *cert = NULL;
+        EVP_PKEY *pkey = NULL;
+        OSSL_STORE_INFO *info;
+
+        // Load all objects from the store
+        while ((info = OSSL_STORE_load(store_ctx)) != NULL) {
+            int type = OSSL_STORE_INFO_get_type(info);
+            
+            if (type == OSSL_STORE_INFO_CERT && !cert) {
```
DimitriPapadopoulos, line 1179: Sorry for the noob question, but can it happen that there are multiple keys in the store? If so, why chose the first one?

DimitriPapadopoulos (commented):
Please fix identation.

ssoftpro:
Thank you for useful observations!
When improving certificate and private key selection, I've opted for user-specified selection, rather than relying on automatic algorithms. If multiple certificates or keys are found, the application will now fail with an error, prompting users to provide more specific PKCS#11 URL parameters.
While automatic selection methods exist, like:

- Validity period selection: X509_getm_notAfter() - prefer newer certs
- Key usage validation: X509_check_purpose() - check client auth usage
- Algorithm strength: EVP_PKEY_get_bits() - prefer stronger algorithms

it is expected PKCS#11 token users to know precisely which certificates and keys they need. Automatic selection carries the risk of choosing an incorrect certificate or key.

ezekiel:
Is there more needed for maintainer approval here?

ezekiel:
Actually, one comment which I hope might help.

Because this implements OSSL providers, it should work even when built with `OPENSSL_NO_ENGINE` is defined, right?

If true, some of the macro ordering in `tunnel.c` may need a tweak.

And in fact I think this ends up being the default case when built on Fedora 41+ - so provider support will be a big win for Fedora.

DimitriPapadopoulos:
How would you like to reorder macroS?

DimitriPapadopoulos:
I could merge the @ssoftpro PR as it is if. Then @ezekiel could suggest a new PR that fixes `OPENSSL_NO_ENGINE` builds. By the way, when is `OPENSSL_NO_ENGINE` defined?

ezekiel:
That's great with me.

I can't tell exactly where it's defined, but it seems globally set on Fedora 42.

ezekiel:
Should I dismiss my review so this can be merged? Or is that not holding it back?.

ezekiel:
Yes, thanks for this awesome improvement to enable us to use OSSL Providers, @ssoftpro !!

PR #1048, opened 2022-12-10
Use a TUN device and embedded PPP code, instead of using pppd

DimitriPapadopoulos:
This patch provided by @rain2fog in https://github.com/adrienverge/openfortivpn/issues/801#issuecomment-740306823 works for me. I get a working VPN tunnel without using `pppd`. I wouldn't mind help with reviewing.

From https://github.com/adrienverge/openfortivpn/issues/801#issuecomment-735513189:
> we simulated the PPP negotiation procedure. There are 3 stages in PPP protocal.
> * The 1th stage is LCP. client will send Magic-Number and MRU request to fortivpn server, then will ack the server request. server will send the Magic-Number and ack the client request.
> * The 2th stage is IPCP. both will will send the IPCP config-request to other. client need to send the IPCP config-request Option (3) with the IP which come from the XML. then both config-request need to be ack.
> * The 3th stage is 0x0021. we can read and write IP packet from tun NIC. then sending data by encapsulating or decapsulating the IP packet with header 0x0021(2 Bytes) + (totolsize(2 Bytes) + 0x5050 + pppsize(2 Bytes)).
we also need send the LCP echo-request to fortinet VPN to keep the TCP connection wouldn't close.

I get complier warnings on Ubuntu 22.04:
```
In function ‘conf_option_encode’,
    inlined from ‘conf_request’ at src/io.c:415:2:
src/io.c:337:16: warning: ‘request.tail’ may be used uninitialized [-Wmaybe-uninitialized]
  337 |         optlist->tail->type = type;
      |         ~~~~~~~^~~~~~
src/io.c: In function ‘conf_request’:
src/io.c:410:33: note: ‘request’ declared here
  410 |         struct conf_option_list request;
      |                                 ^~~~~~~
In function ‘conf_option_encode’,
    inlined from ‘ipcp_packet’ at src/io.c:984:4:
src/io.c:337:16: warning: ‘request.tail’ may be used uninitialized [-Wmaybe-uninitialized]
  337 |         optlist->tail->type = type;
      |         ~~~~~~~^~~~~~
src/io.c: In function ‘ipcp_packet’:
src/io.c:981:49: note: ‘request’ declared here
  981 |                         struct conf_option_list request;
      |                                                 ^~~~~~~
In function ‘conf_option_encode’,
    inlined from ‘ipcp_packet’ at src/io.c:1038:4:
src/io.c:337:16: warning: ‘request.tail’ may be used uninitialized [-Wmaybe-uninitialized]
[...]

PR #1297, opened 2025-08-27
Route: always remove wrong pppd route to self.

caeies:
* with the pppd >= 2.5.0 we have to accept the remote ip.
 * but pppd build a wrong route to self in that case.
 * this route is silently removed if set-routes=1 but it should be of course removed is set=routes=0 too.
 * factorize and properly handle non silently for easy troubleshooting.


caeies:
@DimitriPapadopoulos : here's the proposed fix for the route issue. Note: it should by side effect fix the same issue for the current tun case in #1048 .

Hope the code is ok for you.

src/tunnel.c
```diff
@@ -113,7 +113,9 @@ static int on_ppp_if_up(struct tunnel *tunnel)
 {
 	log_info("Interface %s is UP.\n", tunnel->ppp_iface);
 
-	if (tunnel->config->set_routes) {
+	/* In all case call the ipv4_set_tunnel_routes that will properly handle */
+	/* the tunnel->config->set_routes flag anyway */
```
DimitriPapadopoulos, line 117: I would simply call `ipv4_drop_wrong_route()` here to minimize modifications to the existing code and make clear that `ipv4_drop_wrong_route()` is always called and `ipv4_set_tunnel_routes()` is called only `if (tunnel->config->set_routes)`.

src/ipv4.c
```diff
@@ -726,6 +726,40 @@ static int ipv4_del_route(struct rtentry *route)
 	return 0;
 }
 
+/*
+ * This is a workaround for wrongly configured pppd with ip-accept-remote
+ * If the fortiGate is configured to send it's own external address
```
DimitriPapadopoulos, line 731:
```suggestion
 * If the FortiGate is configured to send its own external address
 ```

File: src/ipv4.c [hunk 3]
```diff
@@ -726,6 +726,40 @@ static int ipv4_del_route(struct rtentry *route)
...
+ * So this method test this and drop the wrongly configured route.
+ * And no more silently to help troubleshooting !
+ */
+static int ipv4_drop_wrong_route(struct tunnel *tunnel)
+{
+	struct rtentry *gtw_rt = &tunnel->ipv4.gtw_rt;
+	int ret;
+
+	route_init(gtw_rt);
+	// Set the up a route to the tunnel gateway
```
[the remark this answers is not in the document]
caeies: thanks, fixing it.

src/ipv4.c
```diff
@@ -726,6 +726,40 @@ static int ipv4_del_route(struct rtentry *route)
 	return 0;
 }
 
+/*
+ * This is a workaround for wrongly configured pppd with ip-accept-remote
+ * If the fortiGate is configured to send it's own external address
+ * pppd will create a route to self for it, killing everything ...
+ * So this method test this and drop the wrongly configured route.
+ * And no more silently to help troubleshooting !
```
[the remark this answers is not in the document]
caeies: thanks, fixing it

src/ipv4.c
```diff
@@ -726,6 +726,40 @@ static int ipv4_del_route(struct rtentry *route)
 	return 0;
 }
 
+/*
+ * This is a workaround for wrongly configured pppd with ip-accept-remote
+ * If the fortiGate is configured to send it's own external address
+ * pppd will create a route to self for it, killing everything ...
+ * So this method test this and drop the wrongly configured route.
```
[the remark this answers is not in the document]
caeies: thanks, fixing it.

src/tunnel.c
caeies: modification done (as well as better return code handling ?).

DimitriPapadopoulos: I see. I have no strong opinion about it. Adding IPv6 support will probably require extensive changes any way, for example the function could be renamed  `drop_wrong_route()`.

PR #1298, opened 2025-08-27
SSL: avoid to leak ssl context on logged out in loop.

### LINKED ISSUE #1263: Please fix tunnel leak on run_tunnel out
Hi all,

I found that the tunnel struct was leaking when run_tunnel was going out. While this is not problematic in most cases, IT is when the endpoint is half working, leading to a lot's of loop for run_tunnel, leading to reach the maximum opened files ...

I fix it like this but feel free to fix it in a better way : 

```
diff --git a/src/tunnel.c b/src/tunnel.c
index 667410e..64e4d85 100644
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -1352,6 +1352,7 @@ err_tunnel:
        } else {
                auth_log_out(&tunnel);
                log_info("Logged out.\n");
+               ssl_disconnect(&tunnel);
        }
 
        // explicitly free the buffer allocated for split routes of the ipv4 configuration
```

(was against today master).

Thanks!

caeies:
* Closes #1263.
 * If we go out of run_tunnel we should always ensure to reset the ssl context to avoid any file exhaustion for the process.

### DIFF
--- a/src/tunnel.c
+++ b/src/tunnel.c
@@ -1445,6 +1445,7 @@ err_tunnel:
 	} else {
 		auth_log_out(&tunnel);
 		log_info("Logged out.\n");
+		ssl_disconnect(&tunnel);
 	}
 
 	// explicitly free the buffer allocated for split routes of the ipv4 configuration


DimitriPapadopoulos:
Thank you @caeies.

PR #1299, opened 2025-08-27
Tun

caeies:
* Adding some features to the tun pull request:
  * capacity to rename the tun interface
  * sample file for dispatch.d to allow easy setup additional routes.
  * allow to use the tun / tun-ifname in a openfortivpn.conf file.
  *  Fixes the persistent=X case properly.

Still missing: conditional compilation to HAVE_LINUX_IF_TUN_H to avoid breaking BSD / MacOS builds with the tun interface code ...

NOTE: to get it works, if not using the set-routes=1 option (aka set-routes=0 or no-routes) you should consider the #1297 too !

### DIFF
--- a/etc/NetworkManager/dispatcher.d/30-pppVPN-rules.sh
+++ b/etc/NetworkManager/dispatcher.d/30-pppVPN-rules.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+# Sample script to add in the /etc/NetworkManager/dispatcher.d folder.
+# Allowing you to properly configure additionnal routes if needed
+# There might be better way to do this, but I'm not aware of them in my specific case.
+
+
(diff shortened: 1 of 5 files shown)


caeies:
I run test/lint/run.sh and fixes commit to be ok.

caeies:
> Can you rebase to resolve conflicts?

Sure, done. Be aware that I added the release bug fix proposal which is still under testing on my side.

caeies:
> > Can you rebase to resolve conflicts?
> 
> Sure, done. Be aware that I added the release bug fix proposal which is still under testing on my side.

@DimitriPapadopoulos : ok, the vpn goe down then up, so the fix is working. I will check tomorrow If I see some obvious memory leaks around.

PR #1306, opened 2025-09-03
Tunnel: don't trigger FortiGate password blocking on wrong password.

caeies:
* Create a specific PR for this new "feature".

While doing some testing, I mistakenly enter a wrong password and get blacklisted by the FortiGate. This PR should avoid this kind of issue.

Maybe we should check better the http output to identify the failure reason, but I feel it like a to intrusive change :).


DimitriPapadopoulos:
About the failure reason, see also #1294. It might be worth recovering the HTTP error code. I agree the change is probably intrusive, but it would help both with deciding what to do after an HTTP error and improving error messages.

DimitriPapadopoulos:
I had already tried integrating an external HTTP parser. I cannot recall which one exactly, probably [PicoHTTPParser](https://github.com/h2o/picohttpparser). See https://github.com/adrienverge/openfortivpn/issues/347.

I think it would be OK to rely on an external simple HTTP library, small enough that we could ship it with openfortivpn in case it is not available on the build or target system.

PR #1317, opened 2025-11-11
Extract & log authentication error messages

rmilecki:
```
If authentication fails then server replies with page containing error
message as an HTML comment. Two real life examples:

</tr>
<script>document.getElementById('err_title').innerHTML=fgt_lang['error'];</script>
<!--sslvpnerrmsg=Password has expired.-->
<tr>
<td class="body" height=100>
<table class="body"><tr><td id='err_val' title='sslvpn_login_password_expired' align="center">
<script>

</tr>
<script>document.getElementById('err_title').innerHTML=fgt_lang['error'];</script>
<!--sslvpnerrmsg=Permission denied.-->
<tr>
<td class="body" height=100>
<table class="body"><tr><td id='err_val' title='sslvpn_login_permission_denied' align="center">

Check HTTP response looking for the <!--sslvpnerrmsg=.*--> pattern and
display extracted message if found. This should help resolving
authentication issues for end users.

Example:
NetworkManager[16950]: INFO:   Connected to gateway.
NetworkManager[16950]: ERROR:  Received error message: "Permission denied."
NetworkManager[16950]: ERROR:  Could not authenticate to gateway. Please check the password, client certificate, etc.
NetworkManager[16950]: INFO:   Closed connection to gateway.
```

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -664,6 +664,7 @@ int auth_log_in(struct tunnel *tunnel)
 	char token[128], tokenresponse[256], tokenparams[320];
 	char action_url[1024] = { '\0' };
 	char *res = NULL;
+	const char *errmsg_start;
 	uint32_t response_size;
 
 	url_encode(username, tunnel->config->username);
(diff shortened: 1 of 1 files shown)


rmilecki:
This pull request deals with a non-standard reply from gateway that is HTML based (instead of plain text based). It seems like a problem itself (we really should get a text response). This needs more work.

Issue #1285, opened 2025-05-27
When a new release?


justalinuxuser1337:
First of all thank you very much for this, it's a life saver!

However, I was just wondering when a new release will be deployed. There are also several PR's ready to be merged which I love to see included in a new release :)

adrienverge:
Hello,

Since the last release (openfortivpn 1.23.1), only 2 commits were merged into master branch: a87b91e "Update package links for distros in README" and 912f1af "Remove deprecated option --plugin". These two are not enough to create a new releases now.

PR #1321, opened 2025-11-19
openfortivpn 1.24.0

pgporada:
Fixes https://github.com/adrienverge/openfortivpn/issues/1285

The changelog was generated with 
```
$ git log --oneline --pretty=format:"%s" a011b5c916075bd240db6a03c02eac95baf09ad8..bb6be5454078cc40de758292b9ca6f2bd5d7a835 | sort -u
```

A maintainer will need to create the actual release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,19 @@ Releases
 This high level changelog is usually updated when a release is tagged.
 On the master branch there may be changes that are not (yet) described here.
 
-### not released yet
+### 1.24.0
 
-* [~] remove deprecated option "plugin"
+* [+] add OpenSSL 3.0+ PKCS#11 support using OSSL_STORE API
+* [+] add OpenSSL Engine support (with OpenSSL < 3.0)
+* [+] update package links for distros in README
+* [~] remove deprecated option `--plugin`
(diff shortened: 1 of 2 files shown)


pgporada:
Alternatively if this should be 1.23.2 instead, that change is easy enough to make.

adrienverge:
Hello Phil, thanks for taking the time!

> Alternatively if this should be 1.23.2 instead, that change is easy enough to make.

Your choice of 1.24.0 seems right to me, since removing `--plugin` is a breaking change.

In my opinion a few changes are needed for consistency with other releases:
- Name the commit like other release commits, in particular remove the `v` from the version number (`v` is only for tags).
- Remove the empty `### not released yet` section.
- Remove the dots at the end?
- Remove cosmetic changes, e.g. these:
  \- `Add C99/POSIX headers`
  \- `Avoid irrelevant fixed-size type uint32_t`
  \- `Address GCC warning`
  \- `actions/checkout`
  \- `github/codeql-action`
  \- `proper indentation`
  \- `Work around Coverity false positive` and `Fix Coverity defect`: merge into `fix a few coverity warnings`?

@DimitriPapadopoulos what do you think? OK to release version 1.24.0?

PR #1322, opened 2025-11-20
Check authentication result (return value)

rmilecki:
```
Authentication replies from gateway include a "ret" value that provides
info about authentication result.

Examples:
ret=0,redir=/remote/login?&err=sslvpn_login_password_expired&lang=en
ret=0,redir=/remote/login?&err=sslvpn_login_permission_denied&lang=en
ret=1,redir=/remote/hostcheck_install?auth_type=1&user=0123456789ABCDEF&&grpname=0123456789ABCDEF012345&portal=0123456789ABCDEF0123456789&rip=1.1.1.1&realm=
ret=6,actionurl=/remote/logincheck,magic=1-12345678,reqid=0,grpid=1,pid=249,is_chal_rsp=1,pass_renew=1,allow_cancel=1,chal_msg=Your password will expire in 3 days. Would you like to change it?

Extract & use that value to determine authentication status.
```

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -637,6 +637,7 @@ int auth_log_in(struct tunnel *tunnel)
 	char username[3 * USERNAME_SIZE + 1];
 	char password[3 * PASSWORD_SIZE + 1];
 	char realm[3 * REALM_SIZE + 1];
+	char auth_ret_text[AUTH_RET_SIZE + 1];
 	char reqid[32] = { '\0' };
 	char polid[32] = { '\0' };
 	char group[128] = { '\0' };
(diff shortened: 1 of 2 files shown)


rmilecki:
Ping @adrienverge @DimitriPapadopoulos

Can we have this authentication check accepted & merged? Could you take a look at it, please?

src/http.c
```diff
@@ -637,6 +637,7 @@ int auth_log_in(struct tunnel *tunnel)
 	char username[3 * USERNAME_SIZE + 1];
 	char password[3 * PASSWORD_SIZE + 1];
 	char realm[3 * REALM_SIZE + 1];
+	char auth_ret_text[8];
```
DimitriPapadopoulos: Why `8` by the way? Perhaps a constant variable would help understand this value.

DimitriPapadopoulos: I was merely suggesting `long` because you're using [`strtol`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/strtol.html).

DimitriPapadopoulos (approved):

rmilecki:
Thank you for looking at this and helping!

PR #1333, opened 2026-03-29
macOS: register VPN DNS via SystemConfiguration framework

### LINKED ISSUE #534: DNS Resolving fails on macOS.
As already stated in https://github.com/adrienverge/openfortivpn/issues/501#issuecomment-571600170 currently openfortivpn doesn't properly register itself with the macOS.

The issue boils down to the fact that on macOS `/etc/resolv.conf` is used as a compatibility layer for software that prefers to do DNS resolution directly instead of using the system provided APIs.

This means that while tools like `nslookup` will happily use whatever openfortivpn writes to `/etc/resolv.conf` other tools such as `ping` which gets it's DNS resolution through `gethostbyname` will still be using the data stored in so called `SystemConfiguration` framework. 

Official FortiClient does that by somewhat strange configuration where it registers itself within the SystemConfiguration as piggybacking on `en0`.

dervisgenc:
## Problem

On macOS, `/etc/resolv.conf` is only a compatibility layer. The actual DNS
resolution uses the SystemConfiguration framework (`configd` / `SCDynamicStore`).
VPN nameservers written to `/etc/resolv.conf` are ignored by most applications
(`ping`, `curl`, browsers, etc.).

This is a long-standing issue reported in #534.

## Solution: Split-DNS via Supplemental Resolver

We register a supplemental resolver in `SCDynamicStore` under
`State:/Network/Service/openfortivpn/DNS`.

### When `dns_suffix` is available (e.g. "corp.com") — Split-DNS

`SupplementalMatchDomains` is set to the actual domain suffix:

```
State:/Network/Service/openfortivpn/DNS = {
    ServerAddresses: [x.x.x.x, x.x.x.x]
    SupplementalMatchDomains: ["corp.com"]
}
```

This gives us two things for free:
- **DNS routing**: only `*.corp.com` queries go to the VPN nameservers
- **Search domain**: configd automatically adds "corp.com" to the default
  resolver's search list, so `ping host` expands to `host.corp.com`

Non-matching queries (e.g. `google.com`) continue to use normal DNS
(home router / ISP). This is the **preferred mode**.

### When `dns_suffix` is NOT available — Full-tunnel DNS

`SupplementalMatchDomains` is set to `[""]` (empty string = match all):

```
State:/Network/Service/openfortivpn/DNS = {
    ServerAddresses: [x.x.x.x, x.x.x.x]
    SupplementalMatchDomains: [""]
}
```

All DNS queries go through the VPN. Short hostname expansion does **not**
work in this mode — this is a macOS configd limitation (see below).

## Why not modify the primary service's DNS?

I explored several alternative approaches before settling on this design:

### Approach 1: `Setup:/Network/Service/<UUID>/DNS` injection 

I injected `SearchDomains` into the primary network service's `Setup:` DNS
key. This worked but had serious drawbacks:

- **Crash safety**: `Setup:` keys are persistent (stored on disk). If
  openfortivpn crashes, the injected search domains remain permanently
  in the user's Wi-Fi/Ethernet DNS configuration until manual cleanup.
- **Modifies another service**: We were editing a service we don't own.
- **Complexity**: Required save/restore logic with static state variables.

### Approach 2: Writing to the PPP service's `State:/Network/Service/<UUID>/DNS`

pppd creates a real network service with PPP + IPv4 entries. We tried writing
DNS to that same service UUID. However, because ppp0 is not the primary
interface (split routing), configd classifies it as a **scoped resolver** —
it only handles queries that are already routed through ppp0, not general DNS.

### Approach 3: `State:/Network/Global/DNS` direct write

Writes are accepted but configd's IPMonitor plugin overwrites this key from
its own calculations, and the mDNSResponder notification chain doesn't fire,
so `scutil --dns` shows stale data.

### Why supplemental resolvers can't carry SearchDomains

This is hardcoded in Apple's configd source (`dns-configuration.c`):

1. `add_supplemental()` explicitly strips `SearchDomains` from supplemental resolvers
2. For empty-string match domains, `DomainName` is set to NULL
3. `extract_search_domains()` only reads `SearchDomains` from the default resolver

When `SupplementalMatchDomains` contains a **real domain** (e.g. "corp.com"),
configd adds it to the search list via the `DomainName` field. This is why
split-DNS mode gets search domains "for free" while full-tunnel mode cannot.

**If full-tunnel DNS with search domain support is needed**, the
`Setup:/Network/Service/<UUID>/DNS` injection approach can be re-introduced
as an opt-in feature behind a flag. However, for the common use case (accessing
corporate resources via VPN while keeping normal internet DNS), split-DNS is
the correct and safe solution.

## Changes

- **configure.ac**: Add `AC_CANONICAL_HOST` so `$host_os` is set and the
  `darwin*` case in the SystemConfiguration check matches on macOS
- **src/ipv4.c**: Implement `ipv4_set_dns_scf()` and `ipv4_clear_dns_scf()`
  using SCDynamicStore API with split-DNS / full-tunnel logic
- **src/ipv4.h**: Declare the new functions under `#if HAVE_SYSTEMCONFIGURATION`
- **src/tunnel.c**: Call the new functions from `on_ppp_if_up()` /
  `on_ppp_if_down()`

## Testing

Tested on macOS Tahoe 26.3.1 with FortiGate gateway:

```
$ scutil --dns   # after connecting (split-DNS mode)
resolver #N
  domain   : corp.com
  nameserver[0] : x.x.x.x
  nameserver[1] : x.x.x.x
  flags    : Supplemental, Request A records

$ ping host                        # expands via search domain to host.corp.com ✓
$ ping host.corp.com               # resolves via VPN DNS ✓
$ ping google.com                  # uses normal DNS, not VPN ✓
```

DNS entry is cleaned up on disconnect with a single key removal.

Closes #534

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -7,6 +7,7 @@ AC_CONFIG_SRCDIR([src/main.c])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 
 # Checks for programs.
+AC_CANONICAL_HOST
 AC_PROG_CC
 AC_PROG_MKDIR_P
 AC_PROG_SED
--- a/configure.ac
+++ b/configure.ac
@@ -180,6 +180,22 @@ LIBS="${OPENSSL_LIBS} ${LIBS}"
 CPPFLAGS="${save_openssl_CPPFLAGS}"
 LIBS="${save_openssl_LIBS}"
 
+# Link macOS DNS frameworks and expose a feature macro for SCDynamicStore support.
+AC_MSG_CHECKING([for macOS SystemConfiguration framework])
+case "$host_os" in
+	darwin*)
+		AC_DEFINE([HAVE_SYSTEMCONFIGURATION], [1],
+		          [Define to 1 on macOS for SCDynamicStore DNS support])
+		LIBS="$LIBS -framework SystemConfiguration -framework CoreFoundation"
+		AC_MSG_RESULT([yes])
+		;;
+	*)
+		AC_DEFINE([HAVE_SYSTEMCONFIGURATION], [0],
+		          [Define to 0 when macOS SystemConfiguration is unavailable])
+		AC_MSG_RESULT([no])
+		;;
+esac
+
 # Specialised checks for particular features.
 # When cross compile, don't run the tests.
 AC_ARG_WITH([rt_dst],
--- a/src/ipv4.c
+++ b/src/ipv4.c
@@ -37,6 +37,11 @@
 #include <stdint.h>
 #include <string.h>
 
+#if HAVE_SYSTEMCONFIGURATION
+#include <SystemConfiguration/SCDynamicStore.h>
+#include <CoreFoundation/CoreFoundation.h>
+#endif
+
 #define IPV4_GET_ROUTE_BUFFER_CHUNK_SIZE 65536
 #define SHOW_ROUTE_BUFFER_SIZE 128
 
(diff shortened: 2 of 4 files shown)


NN---:
Works great !
This saved my day.

NN---:
Is this project still developed?
Do we need to make a fork just to fix this inconvenience?

dervisgenc:
@DimitriPapadopoulos whenever you have some free time, I'd appreciate a quick review on this.

DimitriPapadopoulos:
As already written here, the project lacks maintainers. We have been looking for motivated maintainers for a long time, without success. As a result, the pace of updates is indeed quite slow.

Additionally, Fortinet will stop SLL VPN in a few years time, so I'm not sure the heavy uplifting needed for easier maintenance is worth it.

With that said, I remain willing to look into worthwhile PRs such as this one, time permitting.

NN---:
Can we merge this at least ?
This unblocks openfortivpn for many users.

PR #1341, opened 2026-07-06
Show gateway error message

imario42:
The gateway may embed the reason for refusing a connection in an HTML comment, e.g. . This message was only visible with -vv. Extract it and log it as an error so users see why authentication failed, along with the usual generic error message.

(Note: This simple code has been created using Claude Code - my C skills are too rusty already)

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -142,6 +142,41 @@ static const char *find_header(const char *res, const char *header,
 
 
 /*
+ * The gateway may embed the reason for refusing a connection in an HTML
+ * comment such as:
+ *     <!--sslvpnerrmsg=Permission denied.-->
+ *     <!--sslvpnerrmsgkey=sslvpn_login_permission_denied-->
+ * Show it to the user, otherwise it is only visible in the debug details.
+ */
+static void log_gateway_error_message(const char *response, uint32_t size)
+{
+	static const char prefix[] = "<!--sslvpnerrmsg";
+	const char *start, *end;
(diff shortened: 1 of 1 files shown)


rmilecki:
I cooked something similar few months ago but then I realized we shouldn't be getting HTML responses at all.

https://github.com/adrienverge/openfortivpn/pull/1317

Do you know why you don't get plan text response but HTML page instead?

src/http.c
```diff
@@ -141,6 +141,41 @@ static const char *find_header(const char *res, const char *header,
...
 
 
+/*
+ * The gateway may embed the reason for refusing a connection in an HTML
+ * comment such as:
+ *     <!--sslvpnerrmsg=Permission denied.-->
+ *     <!--sslvpnerrmsgkey=sslvpn_login_permission_denied-->
+ * Show it to the user, otherwise it is only visible in the debug details.
+ */
+static void log_gateway_error_message(const char *response, uint32_t size)
```
rmilecki, line 151: Values of `sslvpnerrmsg` and `sslvpnerrmsgkey` are different (text vs. code). I think they have different purposed (one is a friendly user message, the other can be checked against known values). This function should be checking for a specific one, not "random" (whatever appears first in the HTML code).

imario42:
I have no control over the Fortigate appliance used there, but as far as I remember I never saw something else than html responses. The customer is using Okta as their authentication provider. Maybe it is because of this?

